PR #9, opened 2021-10-21
feat: add deploy

### LINKED ISSUE #5: addTransaction()
- [ ] Test it in `__tests__`
- [ ] Generate a valid tx

### LINKED ISSUE #11: compression procedure munges large integers
https://discord.com/channels/793094838509764618/793094838987128844/901123694523482152
https://discord.com/channels/793094838509764618/793094838987128844/901260335296704582

could potentially use https://www.npmjs.com/package/json-bigint to avoid this problem. I've checked that it can solve at least part of the difference in the program compression pipelines (when employed as in https://gist.github.com/RoboTeddy/41bd89ea8a4ac3df53c6d883de5a0501)

janek26:
This PR types the addTransaction method and adds tests to it.

but it'd be nice to test an "invoke_function" transaction also, as just "deploy" transactions are tested right now.

It also adds a helper to deploy contracts more easily.

Closes #5 
Closes #11

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,7 +5,7 @@
     "node": true,
     "jest": true
   },
-  "extends": ["airbnb-base", "prettier", "plugin:prettier/recommended", "plugin:import/typescript"],
+  "extends": ["airbnb-base", "airbnb-typescript/base", "prettier", "plugin:prettier/recommended", ],
   "globals": {
     "Atomics": "readonly",
     "SharedArrayBuffer": "readonly"
@@ -13,17 +13,8 @@
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
     "ecmaVersion": 12,
-    "sourceType": "module"
+    "sourceType": "module",
+    "project": "./tsconfig.eslint.json"
   },
-  "plugins": ["@typescript-eslint"],
-  "rules": {
-    "camelcase": "off",
-    "import/extensions": [
-      "error",
-      "ignorePackages",
-      {
-        "ts": "never"
-      }
-    ]
-  }
+  "plugins": ["@typescript-eslint"]
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,6 +30,7 @@
         "@typescript-eslint/parser": "^5.0.0",
         "eslint": "^7.32.0",
         "eslint-config-airbnb-base": "^14.2.1",
+        "eslint-config-airbnb-typescript": "^14.0.1",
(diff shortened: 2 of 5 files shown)


seanjameshan:
> > Thanks a lot for making great progress on the V1. My general comments:
> > 
> > * For eslint, let's add support for typescript (assuming that's the issue) instead of turning off rules. Unless there is an absolute necessity to turn these off, I would prefer to have these on.
> > * I would prefer to have package parity with the python version of `starknet` including implementation. Unless we find a special use case only relevant for web developers when they start using `starknet.js`
> > * Add documentation in comment functions if you'd like it to be picked up by `typedoc`
> 
> Now that I have had more time to think about it, I think there are two paths we can go down:
> 
> 1. Implementation parity with `starknet`
> 
> * Pros:
>   
>   * Easily implement and scale new features with the python library from starknet
>   * Less decision making on our end
> * Cons:
>   
>   * Harder to use for JS developers
> 
> 1. Implementation as we see fit
> 
> * Pros:
>   
>   * More flexibility
>   * Better developer experience for js devs
> * Cons:
>   
>   * Harder to scale new features with starknet
>   * More decision making on our end

Think we can go for option 2 for now, and if it doesn't scale go with option 1

seanjameshan:
> > > Thanks a lot for making great progress on the V1. My general comments:
> > > 
> > > * For eslint, let's add support for typescript (assuming that's the issue) instead of turning off rules. Unless there is an absolute necessity to turn these off, I would prefer to have these on.
> > > * I would prefer to have package parity with the python version of `starknet` including implementation. Unless we find a special use case only relevant for web developers when they start using `starknet.js`
> > > * Add documentation in comment functions if you'd like it to be picked up by `typedoc`
> > 
> > 
> > Now that I have had more time to think about it, I think there are two paths we can go down:
> > 
> > 1. Implementation parity with `starknet`
> > 
> > 
> > * Pros:
> >   
> >   * Easily implement and scale new features with the python library from starknet
> >   * Less decision making on our end
> > * Cons:
> >   
> >   * Harder to use for JS developers
> > 
> > 
> > 1. Implementation as we see fit
> > 
> > 
> > * Pros:
> >   
> >   * More flexibility
> >   * Better developer experience for js devs
> > * Cons:
> >   
> >   * Harder to scale new features with starknet
> >   * More decision making on our end
> 
> Think we can go for option 2 for now, and if it doesn't scale go with option 1

Oops didn't mean to close this

PR #54, opened 2021-11-12
fix: msgHash length check fix in signature verify function

### LINKED ISSUE #50: verify function doesn't work correctly, plz also add a test case for that.
- Sign a message with sign function and try to verify it, verification function will return false always even if hash and signature passed is correct.

himanshuchawla009:
fixes #50


janek26:
you sir, are a hero 

ta

janek26 (approved):

PR #63, opened 2021-11-21
fix: correctly parse structs in Starknet return types

corbt:
Currently, starknet.js doesn't do a good job with structs, such as the Uint256 struct returned by the ERC20 contract. Instead of returning all the fields of the struct, it just returns the first field.

Concretely, if a user had a balance of 1 in an ERC20 contract, this is what would have been returned from `balance_of` previously:

```json
{ "res": "0x0" }
```

With this change we properly parse the Uint256 struct, so the method will instead return the following:

```json
{ "res": { "low": "0x1", "high": "0x0" } }
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -16,13 +16,25 @@ export type Type = 'DEPLOY' | 'INVOKE_FUNCTION';
 export type EntryPointType = 'EXTERNAL';
 export type CompressedProgram = string;
 
-export type Abi = {
-  inputs: { name: string; type: 'felt' | 'felt*' }[];
+export type AbiEntry = { name: string; type: 'felt' | 'felt*' | string };
+
+export type FunctionAbi = {
+  inputs: AbiEntry[];
   name: string;
-  outputs: { name: string; type: 'felt' | 'felt*' }[];
+  outputs: AbiEntry[];
   stateMutability?: 'view';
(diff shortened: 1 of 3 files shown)


seanjameshan:
@janek26 could you have a review when you're available? Out of my scope of knowledge

src/types.ts
```diff
@@ -16,13 +16,25 @@ export type Type = 'DEPLOY' | 'INVOKE_FUNCTION';
...
+  inputs: AbiEntry[];
   name: string;
-  outputs: { name: string; type: 'felt' | 'felt*' }[];
+  outputs: AbiEntry[];
   stateMutability?: 'view';
   type: 'function';
 };
+
+export type StructAbi = {
+  members: { name: string; offset: number; type: 'felt' | 'felt*' | string }[];
```
janek26, line 30:
```suggestion
  members: (AbiEntry & { offset: number })[];
```

or maybe offset should even be an optimal property in AbiEntry?

janek26 (changes requested):
Thanks for submitting this!

src/types.ts
corbt: Sure, this feels possibly slightly more correct.

janek26 (approved):

PR #67, opened 2021-11-23
feat: extend signer to sign arbitrary message hashes

vixidev99:
This is a pull request to allow wallets to sign messages without broadcasting them. 

Ethereum had a lot of issues with this, so it will likely require some debate and the establishment of a standard on how messages are signed. See EIP 712 and EIP 191.

https://eips.ethereum.org/EIPS/eip-191
https://eips.ethereum.org/EIPS/eip-712

I know there's a lot of security issues around it, so I don't expect it to get approved without a lot of discussion and additional code, but I wanted to get the ball rolling by putting up some code to start.

### DIFF
--- a/src/signer/default.ts
+++ b/src/signer/default.ts
@@ -1,7 +1,7 @@
 import assert from 'minimalistic-assert';
 
 import { Provider } from '../provider';
-import { AddTransactionResponse, KeyPair, Transaction } from '../types';
+import { AddTransactionResponse, KeyPair, Signature, Transaction } from '../types';
 import { sign } from '../utils/ellipticCurve';
 import { addHexPrefix } from '../utils/encode';
 import { hashMessage } from '../utils/hash';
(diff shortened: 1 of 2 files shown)


vixidev99:
For Zigzag, we sign orders off-chain and send them on-chain when they are matched. 

Without the ability to sign messages through an external Signer, we can't support external wallets like Argent browser extension.

vixidev99:
The security issue is that if you all arbitrary messages to be signed, a malicious app can present a user with a valid Starknet transaction to sign.

Since the garbled message won't look like a valid transaction, a user might sign it and sign their funds away.

seanjameshan:
Thanks for starting this PR, let's continue the discussion here:

https://github.com/argentlabs/argent-x/discussions/14

seanjameshan:
Converting to draft as we get more clarity from the OZ and Argent team on standards. @vixidev99 feel free to open this PR for review as we make progress on the specifications. Thanks!

janek26:
moved to #87 

@vixidev99 feedback appreciated !

PR #78, opened 2021-12-01
feat(utils): support shortstring and uint256

janek26:
This PR implements utils to handle Cairo short strings and Cairo uint256 structs

### DIFF
--- a/src/utils/uint256.ts
+++ b/src/utils/uint256.ts
@@ -1,0 +1,32 @@
+import { addHexPrefix } from './encode';
+import { BigNumberish, toBN } from './number';
+
+// Represents an integer in the range [0, 2^256).
+export interface Uint256 {
+  // The low 128 bits of the value.
+  low: BigNumberish;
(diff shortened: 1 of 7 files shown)


src/utils/uint256.ts
```diff
@@ -0,0 +1,32 @@
+import { addHexPrefix } from './encode';
+import { BigNumberish, toBN } from './number';
+
+// Represents an integer in the range [0, 2^256).
+export interface Uint256 {
+  // The low 128 bits of the value.
+  low: BigNumberish;
+  // The high 128 bits of the value.
+  high: BigNumberish;
+}
```
delaaxe, line 10: Would it make more sense to have a have a class implementing this that would contain the factories, constants above and potential operations? Will people need to multiply Uint256’s a lot?

janek26: I don't think you want to handle Uint256 at all, but just do all operations using BN and convert it to the struct at the last second when committing to starknet

PR #81, opened 2021-12-01
Specifying custom nonces

vixidev99:
When sending multiple transactions at once, the chain state does not update fast enough to reveal an updated nonce, so transactions get signed with the same nonce and rejected. 

This feature allows passing a custom nonce to a transaction so that users can manually input nonces to push transactions through.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,51 @@
+# [2.2.0](https://github.com/seanjameshan/starknet.js/compare/v2.1.0...v2.2.0) (2021-11-30)
+
+### Bug Fixes
+
+- bump version ([02c8d07](https://github.com/seanjameshan/starknet.js/commit/02c8d0772e42c81e35a3a841169eb25cde68716e))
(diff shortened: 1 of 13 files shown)


seanjameshan:
@vixidev99 Thanks a lot for the PR. 

Could you merge this to `develop` first?  I think @delaaxe may have merged the transaction renaming which could have a conflict.

@janek26 could you take a look at this as well? It looks good to me but would like your eyes on it as you've implemented the signer

delaaxe:
@seanjameshan I didn't merge the PR in question but I can still help if there are any conflict issues!

janek26:
Looks good and also makes a lot of sense to integrate this to Argent X.
Keep a local copy of the nonce and use the extended method to queue multiple transactions into one block.

@vixidev99 also notice wallets will most likely support batch execute/multicall soon.

Thanks for the contribution!

janek26:
will get merged in #88 as it needed a rebase + merge target `develop`

PR #85, opened 2021-12-08
feat: support eip712 for starknet

janek26:
implemented as described in https://github.com/argentlabs/argent-x/discussions/14

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -18,4 +18,4 @@ export * as stark from './utils/stark';
 export * as ec from './utils/ellipticCurve';
 export * as uint256 from './utils/uint256';
 export * as shortString from './utils/shortString';
-export * as eip712 from './utils/eip712';
+export * as typedData from './utils/typedData';
--- a/src/index.ts
+++ b/src/index.ts
@@ -18,3 +18,4 @@ export * as stark from './utils/stark';
 export * as ec from './utils/ellipticCurve';
 export * as uint256 from './utils/uint256';
 export * as shortString from './utils/shortString';
+export * as eip712 from './utils/eip712';
--- a/src/signer/interface.ts
+++ b/src/signer/interface.ts
@@ -1,6 +1,6 @@
 import { Provider } from '../provider';
 import { AddTransactionResponse, Signature, Transaction } from '../types';
-import { TypedData } from '../utils/eip712/types';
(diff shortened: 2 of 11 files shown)


src/utils/eip712/types.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Note that the `uint` and `int` aliases like in Solidity, and fixed point numbers are not supported by the EIP-712
+ * standard.
+ */
+export type EIP712Type = Infer<typeof EIP_712_TYPE>;
+
+export const EIP_712_DOMAIN_TYPE = object({
+  name: optional(string()),
+  version: optional(string()),
+  chainId: optional(union([string(), number()])),
+  verifyingContract: optional(pattern(string(), /^0x[0-9a-z]{40}$/i)),
```
juniset, line 64: In the proposed specification the `verifyingContract` is moved out of the `domain separator` since it is an Account Contract that will verify the signature and there is an `account` field in the encoding (also means the domain separator can be pre-computed in dapp contracts if needed). There is also no `salt`.

I would suggest to follow the proposed specification for now and remove these 2 entries from the domain.

I would suggest to keep

src/signer/interface.ts
```diff
@@ -14,4 +15,22 @@ export abstract class SignerInterface extends Provider {
...
   ): Promise<AddTransactionResponse>;
+
+  /**
+   * Sign an JSON object with the starknet private key and return the signature
+   *
+   * @param json - JSON object to be signed
+   * @returns the signature of the JSON object
+   * @throws {Error} if the JSON object is not a valid JSON
+   */
+  public abstract sign(typedData: TypedData): Promise<Signature>;
```
juniset, line 26: We must discriminate between signing for a transaction and signing an off-chain message and that should be reflected in the `Signer` interface. So I would suggest to rename that method to `signMessage`.

src/index.ts
```diff
@@ -18,3 +18,4 @@ export * as stark from './utils/stark';
 export * as ec from './utils/ellipticCurve';
 export * as uint256 from './utils/uint256';
 export * as shortString from './utils/shortString';
+export * as eip712 from './utils/eip712';
```
juniset, line 21: We should not make references to EIP 712 in the code as we are not following the standard exactly. The current specification is only inspired by the EIP 712 standard.

PR #87, opened 2021-12-09
Feature/typed structured data hashing and signing

### LINKED ISSUE #89: Add support of new status to "waitForTx"
The StarkNet in the latest update has split `PENDING` status into `PENDING` and `ACCEPTED_ON_L2`. So we need to add `ACCEPTED_ON_L2` in the "waitForTx" method

janek26:
as described in https://github.com/argentlabs/argent-x/discussions/14

you can find the Cairo implementation here:
https://github.com/argentlabs/argent-contracts-starknet/pull/19

Also contains updates for Cairo 0.6.2 as described in #89 as test wouldn't pass without
closes #89 
closes #90

### DIFF
--- a/src/signer/interface.ts
+++ b/src/signer/interface.ts
@@ -17,7 +17,8 @@ export abstract class SignerInterface extends Provider {
   ): Promise<AddTransactionResponse>;
 
   /**
-   * Sign an JSON object with the starknet private key and return the signature
+   * Sign an JSON object for off-chain usage with the starknet private key and return the signature
+   * This adds a message prefix so it cant be interchanged with transactions
    *
    * @param json - JSON object to be signed
    * @returns the signature of the JSON object
@@ -27,6 +28,7 @@ export abstract class SignerInterface extends Provider {
 
   /**
    * Hash a JSON object with pederson hash and return the hash
+   * This adds a message prefix so it cant be interchanged with transactions
    *
    * @param json - JSON object to be hashed
    * @returns the hash of the JSON object
--- a/src/signer/interface.ts
+++ b/src/signer/interface.ts
@@ -1,6 +1,6 @@
 import { Provider } from '../provider';
 import { AddTransactionResponse, Signature, Transaction } from '../types';
-import { TypedData } from '../utils/eip712/types';
+import { TypedData } from '../utils/typedData/types';
 
 export abstract class SignerInterface extends Provider {
   public abstract address: string;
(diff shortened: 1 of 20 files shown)


src/signer/interface.ts
```diff
@@ -14,4 +15,22 @@ export abstract class SignerInterface extends Provider {
   public abstract override addTransaction(
     transaction: Transaction
   ): Promise<AddTransactionResponse>;
+
+  /**
+   * Sign an JSON object with the starknet private key and return the signature
```
juniset, line 20: To avoid future confusion I think the description of the method should make it clear that it is signing an off-chain message and that it will prefix the message with "StarkNet Message". Otherwise people will assume that you can also use that to sign transactions.

src/signer/interface.ts
```diff
@@ -14,4 +15,22 @@ export abstract class SignerInterface extends Provider {
...
+   *
+   * @param json - JSON object to be signed
+   * @returns the signature of the JSON object
+   * @throws {Error} if the JSON object is not a valid JSON
+   */
+  public abstract signMessage(typedData: TypedData): Promise<Signature>;
```
juniset, line 26: More food for thought, but with account abstraction it is possible to have an account controlled by multiple keys and that all keys must sign an off-chain message if it is to be valid. In that case the method would have to return an array of signatures. Maybe that is something we should anticipate now.

Note that the Account Contract interface in Argent or OpenZeppelin (which can validate signatures with `is_valid_signature` ) already supports signatures as an array.

src/utils/typedData/types.ts
```diff
@@ -0,0 +1,82 @@
...
+  object,
+  optional,
+  record,
+  refine,
+  string,
+  type as t,
+  union,
+} from 'superstruct';
+
+export const STATIC_TYPES = ['felt', 'felt*'];
```
juniset, line 15: I understand what you are doing but I find it weird to have arrays (i.e. `felt*`) listed as static types while they are dynamic variables like structs. Maybe it's just the naming.

src/types.ts
```diff
@@ -3,14 +3,20 @@ import type { ec as EC } from 'elliptic';
...
   Starknet: string;
   GpsStatementVerifier: string;
 };
 
-export type Status = 'NOT_RECEIVED' | 'RECEIVED' | 'PENDING' | 'REJECTED' | 'ACCEPTED_ONCHAIN';
+export type Status =
+  | 'NOT_RECEIVED'
+  | 'RECEIVED'
+  | 'PENDING'
+  | 'ACCEPTED_ON_L1'
```
juniset, line 17: Does not matter but `ACCEPTED_ON_L1` should come after `ACCEPTED_ON_L2` for consistency.

juniset (approved):

PR #90, opened 2021-12-13
fix: add ACCEPTED_ON_L2 status to the waitForTx

### LINKED ISSUE #89: Add support of new status to "waitForTx"

in19farkt:
Close #89


janek26:
will be fixed today by merging #87 

sorry for mixing, but it's needed to pass the tests anyways, and it's also deleting some out of date states

PR #95, opened 2021-12-21
Fix/get block block hash number

seanjameshan:


seanjameshan:
Actually, we may want to reverse blockNumber and blockHash to preserve backwards compaitability

src/provider/default.ts
```diff
@@ -104,16 +104,28 @@ export class Provider implements ProviderInterface {
...
-   * Gets the block information from a block ID.
+   * Gets the block information from a block hash or block number.
    *
-   * [Reference](https://github.com/starkware-libs/cairo-lang/blob/f464ec4797361b6be8989e36e02ec690e74ef285/src/starkware/starknet/services/api/feeder_gateway/feeder_gateway_client.py#L27-L31)
+   * [Reference](https://github.com/starkware-libs/cairo-lang/blob/fc97bdd8322a7df043c87c371634b26c15ed6cee/src/starkware/starknet/services/api/feeder_gateway/feeder_gateway_client.py#L41-L53)
    *
-   * @param blockId
+   * @param blockHash
+   * @param blockNumber
    * @returns the block object { block_id, previous_block_id, state_root, status, timestamp, transaction_receipts, transactions }
```
bbrandtom, line 113: `block_id` -> `block_hash`
same for previous one

src/provider/default.ts
```diff
@@ -104,16 +104,28 @@ export class Provider implements ProviderInterface {
...
    */
-  public async getBlock(blockId?: number): Promise<GetBlockResponse> {
+  public async getBlock(blockHash?: BigNumberish, blockNumber?: number): Promise<GetBlockResponse> {
```
bbrandtom, line 115: I'm not an expert on JS, but blockNumber can also be a string - 'pending'.
Does `blockNumber?: number` support it?
You can read more here - https://community.starknet.io/t/cairo-v0-6-2-api-change-pending-block/195

bbrandtom (commented):
Another comment - this should also be applied to all other API calls.

seanjameshan:
Moved to #98

PR #115, opened 2022-01-21
Feature/add-verifyMessage-to-Signer

### LINKED ISSUE #108: Implement verifyMessage() on signer
Signer should get an additional method `verifyMessage(typedData: TypedData, signature: Signature): Promise<boolean>`, which allows the signer to check if the given signature signs the given message using the signers pub key.
There should also be a `verifyMessageHash(typedDataHash: BigNumberish, signature: Signature): Promise<boolean>`, which allows the signer to check if the given signature is valid for the account by calling `is_valid_signature` on the account contract.

janek26:
closes #108 

- [x] Add tests

### DIFF
--- a/__mocks__/typedDataExample.json
+++ b/__mocks__/typedDataExample.json
@@ -1,0 +1,35 @@
+{
+  "types": {
+    "StarkNetDomain": [
+      { "name": "name", "type": "felt" },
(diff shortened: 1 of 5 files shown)


MilGard91 (approved):
Looks good to me👍

PR #122, opened 2022-01-27
feat(provider): statuses moved to be in types file with proper enum

vaske:
Moving statuses to be inside proper types file.


dhruvkelawala:
Seems like a good PR

janek26:
I tend to agree with this: https://fettblog.eu/tidy-typescript-avoid-enums/ and remove enums in favor of union types, especially in libraries which get used by 3rd parties

janek26:
this is already implemented using union types as you can see here: 
https://github.com/seanjameshan/starknet.js/blob/696cf5ae565bd16365045cc1f20e9fa55184d054/src/types.ts#L13
closed in favor of union types

PR #123, opened 2022-02-01
fix: introduce block identifier type

janek26:
Usage like `getBlock(undefined, 3003)` is very uncommon is JS.
Therefore I introduced an easy pattern to allow blockhash and blockid

### DIFF
--- a/src/provider/default.ts
+++ b/src/provider/default.ts
@@ -3,7 +3,6 @@ import urljoin from 'url-join';
 
 import {
   AddTransactionResponse,
-  BlockNumber,
   CallContractResponse,
   CallContractTransaction,
   CompiledContract,
@@ -19,7 +18,7 @@ import { parse, stringify } from '../utils/json';
 import { BigNumberish, toBN, toHex } from '../utils/number';
 import { compressProgram, formatSignature, randomAddress } from '../utils/stark';
 import { ProviderInterface } from './interface';
-import { getFormattedBlockIdentifier } from './utils';
+import { BlockIdentifier, getFormattedBlockIdentifier } from './utils';
 
 type NetworkName = 'mainnet-alpha' | 'goerli-alpha';
 
@@ -93,10 +92,9 @@ export class Provider implements ProviderInterface {
    */
   public async callContract(
     invokeTransaction: CallContractTransaction,
-    blockHash?: BigNumberish,
-    blockNumber: BlockNumber = null
+    blockIdentifier: BlockIdentifier = null
   ): Promise<CallContractResponse> {
-    const formattedBlockIdentifier = getFormattedBlockIdentifier(blockHash, blockNumber);
+    const formattedBlockIdentifier = getFormattedBlockIdentifier(blockIdentifier);
 
     const { data } = await axios.post<CallContractResponse>(
       urljoin(this.feederGatewayUrl, 'call_contract', formattedBlockIdentifier),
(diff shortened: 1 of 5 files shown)


src/provider/utils.ts
```diff
@@ -18,6 +18,42 @@ export function formatHash() {}
...
+  | { type: 'BLOCK_NUMBER'; data: BlockNumber }
+  | { type: 'BLOCK_HASH'; data: BigNumberish };
+
+/**
+ * Identifies the block to be queried.
+ *
+ * @param blockIdentifier - block identifier
+ * @returns block identifier object
+ */
+export function getBlockIdentifier(blockIdentifier: BlockIdentifier): BlockIdentifierObject {
```
0xs34n, line 35: Nothing to change but a comment: I'm not very familiar with all the edge cases. Just hope that we catch all edge cases which seems like there could be a lot or more in the future.

src/provider/default.ts
```diff
@@ -172,10 +166,9 @@ export class Provider implements ProviderInterface {
   public async getStorageAt(
     contractAddress: string,
     key: number,
-    blockHash?: BigNumberish,
-    blockNumber: BlockNumber = null
+    blockIdentifier: BlockIdentifier = null
   ): Promise<object> {
-    const formattedBlockIdentifier = getFormattedBlockIdentifier(blockHash, blockNumber);
+    const formattedBlockIdentifier = getFormattedBlockIdentifier(blockIdentifier);
```
0xs34n, line 171: Would it be easier to pass in `BlockIdentifierObject` instead of `blockIdentifier` here?

src/provider/utils.ts
0xs34n: I think if we pass in a `BlockIdentifierObject` instead of `BlockIdentifier`, we may not need to handle all edge cases. Thoughts?

src/provider/utils.ts
```diff
@@ -27,12 +63,13 @@ export function txIdentifier() {}
...
-  return `?blockNumber=${blockNumber}`;
+export function getFormattedBlockIdentifier(blockIdentifier: BlockIdentifier = null): string {
+  const blockIdentifierObject = getBlockIdentifier(blockIdentifier);
+  if (blockIdentifierObject.type === 'BLOCK_NUMBER' && blockIdentifierObject.data === null) {
+    return '';
+  }
+  if (blockIdentifierObject.type === 'BLOCK_NUMBER') {
+    return `?blockNumber=${blockIdentifierObject.data}`;
+  }
+  return `?blockHash=${toHex(toBN(blockIdentifierObject.data))}`;
```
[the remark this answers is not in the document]
0xs34n: If we pass in `BlockIdentifierObject` then may not be relevant, same comments as above.

src/provider/default.ts
```diff
@@ -93,10 +92,9 @@ export class Provider implements ProviderInterface {
    */
   public async callContract(
     invokeTransaction: CallContractTransaction,
-    blockHash?: BigNumberish,
-    blockNumber: BlockNumber = null
+    blockIdentifier: BlockIdentifier = null
   ): Promise<CallContractResponse> {
-    const formattedBlockIdentifier = getFormattedBlockIdentifier(blockHash, blockNumber);
+    const formattedBlockIdentifier = getFormattedBlockIdentifier(blockIdentifier);
```
0xs34n, line 97: Duplicate

src/provider/default.ts
```diff
@@ -142,10 +137,9 @@ export class Provider implements ProviderInterface {
    */
   public async getCode(
     contractAddress: string,
-    blockHash?: BigNumberish,
-    blockNumber: BlockNumber = null
+    blockIdentifier: BlockIdentifier = null
   ): Promise<GetCodeResponse> {
-    const formattedBlockIdentifier = getFormattedBlockIdentifier(blockHash, blockNumber);
+    const formattedBlockIdentifier = getFormattedBlockIdentifier(blockIdentifier);
```
0xs34n, line 142: Duplicate

0xs34n (commented):
I think having one argument object type makes sense. 

Just wonder if passing in an object would be better than handling any wild type.

Pros for object:
- Less error-prone IMO as the developer needs to state clearly the type and data

Cons for object:
- More work for developer

Pros of wild type (current implementation)
- Users don't need to care about type and input whatever they want and let us handle it.

Cons of wild type (current implementation)
- We need to handle every edge case of the wild type, may not be scalable long term

janek26:
@0xs34n and I discussed this and think the interface should only require an identifier, no matter what that identifier may be (block_hash or block_number). That's up to the implementation to handle/validate.


For the included default implementation we'll stick with the proposed logic here, handling hashes as a citizen first class, while block_number is just a fallback

0xs34n (approved):
Looks good. If we end up having more identifiers in the future, may be good to revisit complexity of `getBlockIdentifier` we hope that `blockNumber` will be deprecated in the future in favor of `blockHash`;

PR #125, opened 2022-02-08
feat/contract type transformation

MilGard91:
- [x] Add response type transformations 
- [x] Add request type transformations
- [x] Add test for type transformation

PR #127, opened 2022-02-08
feat(Contract): add an option to pass block identifier to call API

dan-ziv:
Add an option to pass block identifier to contract calls.

For example: sometimes we would like to query the calls for the most updated state (i.e, on the pending block).
`contract.call('balanceOf', {account}, "pending")`

### DIFF
--- a/src/contract.ts
+++ b/src/contract.ts
@@ -2,6 +2,7 @@ import BN from 'bn.js';
 import assert from 'minimalistic-assert';
 
 import { Provider, defaultProvider } from './provider';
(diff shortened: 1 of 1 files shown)


janek26:
Hey @dan-ziv ! Thanks for your contribution!!

I'll merge it for now as you're probably waiting for this change to be available :-) But be aware we're already working on a new Contract interface! eg #125

dan-ziv:
Thanks @janek26

janek26 (approved):

PR #128, opened 2022-02-09
feat(account): introduce execute method

janek26:
This PR aims to support batch transactions natively. 

To archive this it's necessary to introduce a new method and not further inherit the interface of `invokeFunction` which is designed to always match the sequencers api, which in any case accepts one call only.

### DIFF
--- a/src/account/default.ts
+++ b/src/account/default.ts
@@ -1,12 +1,10 @@
-import assert from 'minimalistic-assert';
-
 import { compileCalldata } from '../contract';
 import { Provider } from '../provider';
 import { Signer, SignerInterface } from '../signer';
 import {
   Abi,
   AddTransactionResponse,
-  Invocation,
+  ExecuteInvocation,
   InvocationsDetails,
   KeyPair,
   Signature,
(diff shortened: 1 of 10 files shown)


dhruvkelawala:
Looks good to me

src/account/default.ts
```diff
@@ -36,41 +41,52 @@ export class Account extends Provider implements AccountInterface {
    * @param transaction - transaction to be invoked
    * @returns a confirmation of invoking a function on the starknet contract
    */
-  public override async invokeFunction(
-    { contractAddress, calldata = [], nonce, entrypoint, ...invocation }: Invocation,
-    _abi?: Abi
+  public async execute(
```
0xs34n, line 44: Might want to change the comment on this function above `Invoke a function` -> `call execute function in account contract` also the reference is probably not relevant for this function

src/account/interface.ts
```diff
@@ -39,9 +40,10 @@ export abstract class AccountInterface extends ProviderInterface {
    *
    * @returns response from addTransaction
    */
-  public abstract override invokeFunction(
-    invocation: Invocation,
-    abi?: Abi
+  public abstract execute(
```
0xs34n, line 43: same comment as the implementation, change comment from Invoke -> call execute on account contract

0xs34n (changes requested):
Looks good! Small changes on the comments on the functions.

Also, I noticed some formatting changes that I think are being auto-changed from our editors. Not sure what's causing this as we have the prettier config file

janek26:
@0xs34n thanks! done!

PR #138, opened 2022-03-08
Human readable contract interface

MilGard91:
- adding invoke and call directly methods to the contract
- adding estimate fee functions to the contract
- adding populate transaction to the contract


dhruvkelawala:
I know the PR is wip, but instead of having `connectedTo` , can we have `address`?  `contract.address` is more intuitive than `contract.connectedTo`

src/contract.ts
```diff
@@ -1,22 +1,47 @@
...
-  [key: string]: BigNumberish | ParsedStruct;
-};
 export type Args = {
   [inputName: string]: BigNumberish | BigNumberish[] | ParsedStruct | ParsedStruct[];
 };
 
+export type AsyncContractFunction<T = any> = (args: Args, options?: any) => Promise<T>;
+export type ContractFunction = (args: Args) => any;
+
+export interface PopulatedTransaction {}
```
janek26, line 31:
```suggestion
```

src/contract.ts
```diff
@@ -25,24 +50,64 @@ function parseFelt(candidate: string): BN {
...
+}
+
+function buildEstimate(contract: Contract, functionAbi: FunctionAbi): ContractFunction {
+  return function (args: Args): any {
+    return contract.estimate(functionAbi.name, args);
+  };
+}
+
 export class Contract {
   connectedTo: string | null = null;
```
janek26, line 85: as suggested by @dhruvkelawala 
```suggestion
  address: string | null = null;
```

maybe we should also add readonly and public to all properties?

src/contract.ts
```diff
@@ -25,24 +50,64 @@ function parseFelt(candidate: string): BN {
...
 
   abi: Abi;
 
   structs: { [name: string]: StructAbi };
 
-  provider: Provider;
+  providerOrAccount: Provider | Account;
+
+  readonly [key: string]: AsyncContractFunction | any;
```
janek26, line 93: add protected to the next 4 properties?

src/contract.ts
```diff
@@ -244,7 +380,7 @@ export class Contract {
    * @param inputs  - list of inputs(fields) that are in the abi
    * @return {Calldata} - parsed arguments in format that contract is expecting
    */
-  private compileCalldata(args: Args, inputs: AbiEntry[]): Calldata {
+  compileCalldata(args: Args, inputs: AbiEntry[]): Calldata {
```
janek26, line 383: can we keep everything either private, protected or public? I think this method (and following) should be protected

janek26, line 383: Should compileCalldata receive the full ABI to be able to compile Structs aswell? or is that handled somewhere else?

src/contract.ts
```diff
@@ -25,24 +50,64 @@ function parseFelt(candidate: string): BN {
...
 }
 
+function buildCall(contract: Contract, functionAbi: FunctionAbi): AsyncContractFunction {
+  return async function (args: Args, blockIdentifier: BlockIdentifier): Promise<any> {
+    return contract.call(functionAbi.name, args, blockIdentifier);
+  };
+}
+
+function buildInvoke(contract: Contract, functionAbi: FunctionAbi): AsyncContractFunction {
+  return async function (args: Args): Promise<any> {
```
janek26, line 60: `Args` is still the v2 way of accepting invocation/call arguments. I think we should support multiple formats, like ethers.js does and as described in https://github.com/0xs34n/starknet.js/discussions/130. wdyt?
Same for return values, which are objects only right now

janek26 (commented):
Great work! Few comments

janek26:
Could we also split the Contract in the folder pattern used by Provider/Signer/Account? I think having an explicit interface file would be great

src/contract.ts
dhruvkelawala, line 85:
```suggestion
  connectedTo: string | null = null;
  address: string;
```

src/contract.ts
dhruvkelawala: I agree that this method should be private or protected because we already have a `compileCalldata` function exported from `utils/stark`

0xs34n:
Adding some js docs comments would also be very much appreciated!

dhruvkelawala:
Is the PR ready or something is being worked on?

src/contract.ts
janek26: we agreed to improve on this in the future and keep one single format for now.

src/contract.ts
```diff
@@ -73,7 +202,7 @@ export class Contract {
    * @param method  - name of the method
    * @param args - arguments that are passed to the method
    */
-  private validateMethodAndArgs(type: 'INVOKE' | 'CALL', method: string, args: Args = {}) {
+  protected validateMethodAndArgs(type: 'INVOKE' | 'CALL', method: string, args: Args = {}) {
```
janek26, line 205: can we please add a new validation that the provided object doesn't contain more keys than needed?
eg:
```
// ABI expects:
{
  account: address
}
// provided: 
{
  user: "0x5"
}
```
should throw as "user" is not expected in ABI. This should throw. Current behavior is that it just returns with `["0x0"]` as account wasn't provided, so it assumes 0 (which is fine, but confusing without the error I'm talking about).
Would solve #140

janek26 (commented):
I still think we should add a contract folder and split contract.ts into interface.ts and default.ts like we did with account/signer/provider

src/contract.ts
MilGard91: This will change with the new method signature of the Contract since the arguments are going to be positional.

janek26:
well done! I proposed some more changes in #143, please have a look @MilGard91 !

janek26 (approved):

dhruvkelawala:
Nice changes! Thank you for this amazing work. @MilGard91 @janek26 .

PR #139, opened 2022-03-09
feat: add LedgerSigner class + blind signing

jcaporossi-ledger:
Add a Ledger signer + blind signing support only

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,16 +61,13 @@
     "typescript": "^4.4.4"
   },
   "dependencies": {
-    "@ethersproject/bytes": "^5.5.0",
     "@ledgerhq/hw-app-eth": "^6.26.0",
     "@ledgerhq/hw-transport": "^6.24.1",
     "@ledgerhq/hw-transport-webhid": "^6.24.1",
     "axios": "^0.23.0",
     "bn.js": "^5.2.0",
-    "buffer": "^6.0.3",
     "elliptic": "^6.5.4",
     "ethereum-cryptography": "^0.2.0",
-    "ethers": "^5.5.4",
     "hash.js": "^1.1.7",
     "json-bigint": "^1.0.0",
     "minimalistic-assert": "^1.0.1",
--- a/package.json
+++ b/package.json
@@ -61,10 +61,16 @@
     "typescript": "^4.4.4"
   },
   "dependencies": {
+    "@ethersproject/bytes": "^5.5.0",
+    "@ledgerhq/hw-app-eth": "^6.26.0",
+    "@ledgerhq/hw-transport": "^6.24.1",
+    "@ledgerhq/hw-transport-webhid": "^6.24.1",
     "axios": "^0.23.0",
     "bn.js": "^5.2.0",
+    "buffer": "^6.0.3",
     "elliptic": "^6.5.4",
     "ethereum-cryptography": "^0.2.0",
+    "ethers": "^5.5.4",
     "hash.js": "^1.1.7",
     "json-bigint": "^1.0.0",
     "minimalistic-assert": "^1.0.1",
--- a/src/signer/ledger.ts
+++ b/src/signer/ledger.ts
@@ -1,0 +1,63 @@
+import { hexZeroPad } from '@ethersproject/bytes';
+import Eth from '@ledgerhq/hw-app-eth';
+import Transport from '@ledgerhq/hw-transport';
+import TransportWebHID from '@ledgerhq/hw-transport-webhid';
+import { ethers } from 'ethers';
+
+import { Abi, Invocation, InvocationsSignerDetails, Signature } from '../types';
+import { TypedData } from '../utils/typedData';
+import { SignerInterface } from './interface';
+
+export class LedgerSigner implements SignerInterface {
+  public derivationPath = "/2645'/579218131'/1148870696'/0'/0'/0";
+
+  static transport: Transport;
+
+  static async getEthApp(): Promise<Eth> {
+    console.log('isSupported', await TransportWebHID.isSupported());
+    if (!this.transport) {
+      console.log('create TransportWebHID');
+      this.transport = await TransportWebHID.create();
+    }
+    return new Eth(this.transport);
+  }
+
+  public async getPubKey(): Promise<string> {
+    const eth = await LedgerSigner.getEthApp();
+    const response = await eth.starkGetPublicKey(this.derivationPath);
+    const starkPub = `0x${response.slice(1, 1 + 32).toString('hex')}`;
+    return starkPub;
+  }
+
+  public async signTransaction(
+    transactions: Invocation[],
+    transactionsDetail: InvocationsSignerDetails,
+    abis?: Abi[]
+  ): Promise<Signature> {
+    if (abis && abis.length !== transactions.length) {
+      throw new Error('ABI must be provided for each transaction or no transaction');
+    }
+    console.log(transactions);
+    console.log(transactionsDetail);
+    return ['', ''];
+  }
+
+  public async signMessage(typedData: TypedData, accountAddress: string): Promise<Signature> {
+    const eth = await LedgerSigner.getEthApp();
+    const messageHash = ethers.utils.keccak256(ethers.utils.toUtf8Bytes(JSON.stringify(typedData)));
+
+    console.log(`Message to be signed by Nano S/X ${messageHash} for ${accountAddress}`);
+
+    const signature = (await eth.starkUnsafeSign(
+      this.derivationPath,
+      hexZeroPad(messageHash, 32)
+    )) as { r: string; s: string };
+
+    return [`0x${signature.r}`, `0x${signature.s}`];
+  }
+
+  static async askPermissionIfNeeded(): Promise<void> {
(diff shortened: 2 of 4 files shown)


src/signer/ledger.ts
```diff
@@ -0,0 +1,63 @@
...
+      throw new Error('ABI must be provided for each transaction or no transaction');
+    }
+    console.log(transactions);
+    console.log(transactionsDetail);
+    return ['', ''];
+  }
+
+  public async signMessage(typedData: TypedData, accountAddress: string): Promise<Signature> {
+    const eth = await LedgerSigner.getEthApp();
+    const messageHash = ethers.utils.keccak256(ethers.utils.toUtf8Bytes(JSON.stringify(typedData)));
```
janek26, line 47: that's not how EIP712 message hash calculation works. Please use `getMessageHash` as done here: https://github.com/seanjameshan/starknet.js/blob/a709eab7826be89077c9502954e7dc80068be982/src/signer/default.ts#L39

src/signer/ledger.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    console.log(`Message to be signed by Nano S/X ${messageHash} for ${accountAddress}`);
```
janek26, line 49: please don't ship console logs to prod

src/signer/ledger.ts
```diff
@@ -0,0 +1,63 @@
...
+    transactions: Invocation[],
+    transactionsDetail: InvocationsSignerDetails,
+    abis?: Abi[]
+  ): Promise<Signature> {
+    if (abis && abis.length !== transactions.length) {
+      throw new Error('ABI must be provided for each transaction or no transaction');
+    }
+    console.log(transactions);
+    console.log(transactionsDetail);
+    return ['', ''];
```
janek26, line 42: can we get a proper transaction signing implementation? should be as easy as calculating the has and sign that as done below. eg: https://github.com/seanjameshan/starknet.js/blob/a709eab7826be89077c9502954e7dc80068be982/src/signer/default.ts#L18

src/signer/ledger.ts
```diff
@@ -0,0 +1,63 @@
...
+import { TypedData } from '../utils/typedData';
+import { SignerInterface } from './interface';
+
+export class LedgerSigner implements SignerInterface {
+  public derivationPath = "/2645'/579218131'/1148870696'/0'/0'/0";
+
+  static transport: Transport;
+
+  static async getEthApp(): Promise<Eth> {
+    console.log('isSupported', await TransportWebHID.isSupported());
```
janek26, line 17: please remove console log before merge. Does `TransportWebHID.isSupported() ` need to be checked and throw in case it's not supported?

src/signer/ledger.ts
```diff
@@ -0,0 +1,63 @@
...
+import { ethers } from 'ethers';
+
+import { Abi, Invocation, InvocationsSignerDetails, Signature } from '../types';
+import { TypedData } from '../utils/typedData';
+import { SignerInterface } from './interface';
+
+export class LedgerSigner implements SignerInterface {
+  public derivationPath = "/2645'/579218131'/1148870696'/0'/0'/0";
+
+  static transport: Transport;
```
janek26, line 14:
```suggestion
  static transport: Transport | undefined;
```

and should this be private?

package.json
```diff
@@ -61,10 +61,16 @@
...
+    "@ethersproject/bytes": "^5.5.0",
+    "@ledgerhq/hw-app-eth": "^6.26.0",
+    "@ledgerhq/hw-transport": "^6.24.1",
+    "@ledgerhq/hw-transport-webhid": "^6.24.1",
     "axios": "^0.23.0",
     "bn.js": "^5.2.0",
+    "buffer": "^6.0.3",
     "elliptic": "^6.5.4",
     "ethereum-cryptography": "^0.2.0",
+    "ethers": "^5.5.4",
```
janek26, line 73: should not be needed. Especially if the hashing is done right

package.json
```diff
@@ -61,10 +61,16 @@
     "typescript": "^4.4.4"
   },
   "dependencies": {
+    "@ethersproject/bytes": "^5.5.0",
+    "@ledgerhq/hw-app-eth": "^6.26.0",
+    "@ledgerhq/hw-transport": "^6.24.1",
+    "@ledgerhq/hw-transport-webhid": "^6.24.1",
     "axios": "^0.23.0",
     "bn.js": "^5.2.0",
+    "buffer": "^6.0.3",
```
janek26, line 70: why?

janek26 (changes requested):
Thanks for the contribution!

src/signer/ledger.ts
```diff
@@ -0,0 +1,63 @@
...
+import Eth from '@ledgerhq/hw-app-eth';
+import Transport from '@ledgerhq/hw-transport';
+import TransportWebHID from '@ledgerhq/hw-transport-webhid';
+import { ethers } from 'ethers';
+
+import { Abi, Invocation, InvocationsSignerDetails, Signature } from '../types';
+import { TypedData } from '../utils/typedData';
+import { SignerInterface } from './interface';
+
+export class LedgerSigner implements SignerInterface {
```
janek26, line 11:
```suggestion
export class LedgerBlindSigner implements SignerInterface {
```

janek26:
is it possible to add tests? or isn't that possible due to all the side effects (hardware interaction) ?

src/signer/ledger.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    const signature = (await eth.starkUnsafeSign(
+      this.derivationPath,
+      hexZeroPad(messageHash, 32)
+    )) as { r: string; s: string };
+
+    return [`0x${signature.r}`, `0x${signature.s}`];
+  }
+
+  static async askPermissionIfNeeded(): Promise<void> {
```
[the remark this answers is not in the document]
jcaporossi-ledger: This function is used to trig a first pairing with Ledger device. Maybe I can move it into a constructor, I will check...

src/signer/ledger.ts
jcaporossi-ledger: Yes correct, I will fix this

src/signer/ledger.ts
jcaporossi-ledger: ok, I will delete all console.log calls

src/signer/ledger.ts
jcaporossi-ledger: Ok

src/signer/ledger.ts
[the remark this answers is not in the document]
janek26: if it just triggers the pairing if needed you could also call it in the beginning of every method instead of in the constructor

src/signer/ledger.ts
jcaporossi-ledger: Ok done

src/signer/ledger.ts
jcaporossi-ledger: Done

package.json
jcaporossi-ledger: Dependencies required when performing tests with argent-x, should be suppressed....

src/signer/ledger.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  public async signTransaction(
+    transactions: Invocation[],
+    transactionsDetail: InvocationsSignerDetails,
+    abis?: Abi[]
+  ): Promise<Signature> {
+    if (abis && abis.length !== transactions.length) {
+      throw new Error('ABI must be provided for each transaction or no transaction');
+    }
```
janek26, line 51: As it's blind signing I think this check isn't needed

src/signer/ledger.ts
```diff
@@ -0,0 +1,81 @@
...
+      transactionsDetail.walletAddress,
+      transactions,
+      transactionsDetail.nonce.toString(),
+      transactionsDetail.maxFee.toString()
+    );
+
+    const signature = (await eth.starkUnsafeSign(this.derivationPath, hexZeroPad(msgHash, 32))) as {
+      r: string;
+      s: string;
+    };
```
janek26, line 64: as this is done 2 times (here and in signMessage) maybe we can extract it into a `protected` method?
like
```typescript
protected sign(msgHash: string): Promise<Signature> {
  const eth = await this.getEthApp();

  const { r, s }: {
     r: string;
     s: string;
   } = await eth.starkUnsafeSign(this.derivationPath, hexZeroPad(msgHash, 32));

  return [addHexPrefix(r), addHexPrefix(s)]
}
```

janek26 (changes requested):

src/signer/ledger.ts
jcaporossi-ledger: yes correct, fixed

src/signer/ledger.ts
jcaporossi-ledger: yes you're right, making the modification now ;)

janek26 (approved):
looks good :) merge after tests pass

jcaporossi-ledger:
Hello,

 Do you know when the PR will be merged into develop branch ?

PR #148, opened 2022-03-15
Support 0.8.0

MilGard91:
- Contract expanded with estimate fee
- Provider expanded with transaction trace

### DIFF
--- a/src/provider/default.ts
+++ b/src/provider/default.ts
@@ -9,7 +9,6 @@ import {
   CompiledContract,
   DeployContractPayload,
   Endpoints,
-  EstimateFeeResponse,
   GetBlockResponse,
   GetCodeResponse,
   GetContractAddressesResponse,
@@ -332,15 +331,6 @@ export class Provider implements ProviderInterface {
   public invokeFunction(invocation: Invocation, _abi?: Abi): Promise<AddTransactionResponse> {
     return this.fetchEndpoint('add_transaction', undefined, {
       type: 'INVOKE_FUNCTION',
-      contract_address: invocation.contractAddress,
-      entry_point_selector: getSelectorFromName(invocation.entrypoint),
-      calldata: bigNumberishArrayToDecimalStringArray(invocation.calldata ?? []),
-      signature: bigNumberishArrayToDecimalStringArray(invocation.signature ?? []),
-    });
-  }
-
-  public estimateFee(invocation: Invocation): Promise<EstimateFeeResponse> {
-    return this.fetchEndpoint('estimate_fee', undefined, {
       contract_address: invocation.contractAddress,
       entry_point_selector: getSelectorFromName(invocation.entrypoint),
       calldata: bigNumberishArrayToDecimalStringArray(invocation.calldata ?? []),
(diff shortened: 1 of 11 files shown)


src/provider/default.ts
```diff
@@ -327,12 +340,10 @@ export class Provider implements ProviderInterface {
...
     return this.fetchEndpoint('estimate_fee', undefined, {
-      // TODO: change the TYPE of the call
-      type: 'INVOKE_FUNCTION',
       contract_address: invocation.contractAddress,
-      entry_point_selector: getSelectorFromName(invocation.entrypoint),
-      calldata: bigNumberishArrayToDecimalStringArray(invocation.calldata ?? []),
-      signature: bigNumberishArrayToDecimalStringArray(invocation.signature ?? []),
+      entry_point_selector: invocation.entrypoint,
+      calldata: invocation.calldata ?? [],
+      signature: invocation.signature ?? [],
```
janek26, line 346: why does this get removed?

janek26 (changes requested):
looks good, just one q! Thanks!!

src/provider/default.ts
MilGard91: It was throwing 400 when testing. But it isn't now, so I brought it back

src/provider/default.ts
```diff
@@ -327,10 +340,8 @@ export class Provider implements ProviderInterface {
 
   public estimateFee(invocation: Invocation): Promise<any> {
     return this.fetchEndpoint('estimate_fee', undefined, {
-      // TODO: change the TYPE of the call
-      type: 'INVOKE_FUNCTION',
       contract_address: invocation.contractAddress,
-      entry_point_selector: getSelectorFromName(invocation.entrypoint),
+      entry_point_selector: invocation.entrypoint,
```
janek26, line 344: don't you need to call it with the hex string instead of the plain method name?

src/provider/default.ts
```diff
@@ -327,8 +340,6 @@ export class Provider implements ProviderInterface {
 
   public estimateFee(invocation: Invocation): Promise<any> {
```
0xs34n, line 341: Could we type this `Promise` ?

src/provider/interface.ts
```diff
@@ -139,6 +139,19 @@ export abstract class ProviderInterface {
...
+   *
+   * @param invocation the invocation object containing:
+   * - contractAddress - the address of the contract
+   * - entrypoint - the entrypoint of the contract
+   * - calldata - (defaults to []) the calldata
+   * - signature - (defaults to []) the signature
+   *
+   * @returns response from addTransaction
+   */
+  public abstract estimateFee(invocation: Invocation): Promise<any>;
```
0xs34n, line 153: Could we type this Promise ?

0xs34n (commented):
Looks good! Question about typing the promise but everything else looks good on my end. Thanks @MilGard91 !

janek26:
signing to estimate fee is currently needed and will change with 0.9.0

version in signed estimate will change to a higher number in 0.8.1

PR #152, opened 2022-03-23
feat(Account): account constructor should take a KeyPair or a Signer …

ghost:
…as parameter

### DIFF
--- a/src/account/default.ts
+++ b/src/account/default.ts
@@ -30,10 +30,10 @@ export class Account extends Provider implements AccountInterface {
 
   private signer: SignerInterface;
 
-  constructor(provider: Provider, address: string, keyPairOrSigner: KeyPair | Signer) {
+  constructor(provider: Provider, address: string, keyPairOrSigner: KeyPair | SignerInterface) {
     super(provider);
     this.signer =
-      keyPairOrSigner instanceof Signer ? keyPairOrSigner : new Signer(keyPairOrSigner as KeyPair);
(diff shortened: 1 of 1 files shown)


src/account/default.ts
```diff
@@ -30,9 +30,10 @@ export class Account extends Provider implements AccountInterface {
 
   private signer: SignerInterface;
 
-  constructor(provider: Provider, address: string, keyPair: KeyPair) {
+  constructor(provider: Provider, address: string, keyPairOrSigner: KeyPair | Signer) {
```
janek26, line 33: should be SignerInterface

src/account/default.ts
```diff
@@ -30,9 +30,10 @@ export class Account extends Provider implements AccountInterface {
...
     super(provider);
-    this.signer = new Signer(keyPair);
+    this.signer =
+      keyPairOrSigner instanceof Signer ? keyPairOrSigner : new Signer(keyPairOrSigner as KeyPair);
```
janek26, line 36: I have bad experiences with `instanceof` not detecting an instance.. Could we change this to something more low level?

janek26 (changes requested):

src/account/default.ts
jcaporossi-ledger: yes fixed

src/account/default.ts
jcaporossi-ledger: okay, used "in" typeguard operator in the next commit....

janek26 (approved):

MilGard91 (approved):

PR #155, opened 2022-03-24
Release V3.6.1

MilGard91:
- Added support for the Cairo 0.8.0
- New documentation format
- New option to pass a Signer instead of KeyPar to the constructor of the account


janek26:
Hey, I think we should postpone the stable channel release until cairo v0.8.1 or 0.9.0 comes out.
The library contains some dangerous code right now, where estimate fee transactions could be misused as normal transactions
This should be solves in one of the mentioned versions

Users who need these fee features today can use @next

MilGard91:
OK, then let's wait for the 0.8.1, so everyone can check it before the fees are mandatory.

janek26:
dangerous code was fixed, we now use a different version for fee estimation.
Release v3.8.0 ready for public

janek26 (approved):

PR #157, opened 2022-03-31
feat: pending as success

janek26:
After speaking to starkware it turned out "PENDING" should already be used as a done state.
To reflect that in calls we need to query the pending block by default.

### DIFF
--- a/src/account/default.ts
+++ b/src/account/default.ts
@@ -1,6 +1,7 @@
 import assert from 'minimalistic-assert';
 
 import { Provider } from '../provider';
+import { BlockIdentifier } from '../provider/utils';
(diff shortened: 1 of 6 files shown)


dhruvkelawala:
> After speaking to starkware it turned out "PENDING" should already be used as a done state. To reflect that in calls we need to query the pending block by default.

Noice. For end-users, this will "speed up" the transaction.

maciejka:
I had a long discussion with feedthefed about sn transaction life cycle. Here a diagram that summarizes it: https://community.starknet.io/t/transaction-lifecycle-diagram/999

MilGard91 (approved):

PR #159, opened 2022-04-05
feat(account): adding new transaction version for fee estimation

MilGard91:
- Added transaction version to the payload
- Added transaction version to the msgHash


janek26:
closed in favor of #161

PR #161, opened 2022-04-07
Fix invoke override

MilGard91:
- Adding options extraction in build invoke function

### DIFF
--- a/src/contract/default.ts
+++ b/src/contract/default.ts
@@ -46,7 +46,18 @@ function buildCall(contract: Contract, functionAbi: FunctionAbi): AsyncContractF
  */
 function buildInvoke(contract: Contract, functionAbi: FunctionAbi): AsyncContractFunction {
   return async function (...args: Array<any>): Promise<any> {
-    return contract.invoke(functionAbi.name, args);
+    const { inputs } = functionAbi;
+    const inputsLength = inputs.reduce((acc, input) => {
+      if (!/_len$/.test(input.name)) {
+        return acc + 1;
+      }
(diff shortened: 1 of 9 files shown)


src/contract/default.ts
```diff
@@ -46,7 +46,18 @@ function buildCall(contract: Contract, functionAbi: FunctionAbi): AsyncContractF
...
+      if (!/_len$/.test(input.name)) {
+        return acc + 1;
+      }
+      return acc;
+    }, 0);
+    const options = {};
+    if (inputsLength + 1 === args.length && typeof args[args.length - 1] === 'object') {
+      Object.assign(options, args.pop());
+    }
+    return contract.invoke(functionAbi.name, args, options);
```
janek26, line 60: why was this added?

src/utils/stark.ts
```diff
@@ -48,3 +49,8 @@ export function compileCalldata(args: RawArgs): Calldata {
     return toBN(value).toString();
   });
 }
+
+export function estimatedFeeToMaxFee(estimatedFee: BigNumberish, overhead: number = 0.15): BN {
+  const fee = isBN(estimatedFee) ? estimatedFee : toBN(estimatedFee);
```
janek26, line 54: no check needed, if you pass a BN to `toBN` it'll just return that BN

src/contract/default.ts
MilGard91: it checks if the last arg when invoking is the override object.

Example: contract.methodName(...args, {maxFee: '0'})

PR #176, opened 2022-04-26
fix(encode): maximum call stack size exceeded for large compiled cair…

FawadHa1der:
maximum call stack size exceeded error when the cairo file size is larger than 8 mb. I have put in the fix.

### DIFF
--- a/src/utils/encode.ts
+++ b/src/utils/encode.ts
@@ -4,7 +4,7 @@ export const IS_BROWSER = typeof window !== 'undefined';
 const STRING_ZERO = '0';
 
 export function arrayBufferToString(array: ArrayBuffer): string {
-  return new Uint8Array(array).reduce(function (data, byte) {
+  return new Uint8Array(array).reduce((data, byte) => {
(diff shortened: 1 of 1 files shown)


src/utils/encode.ts
```diff
@@ -4,7 +4,9 @@ export const IS_BROWSER = typeof window !== 'undefined';
 const STRING_ZERO = '0';
 
 export function arrayBufferToString(array: ArrayBuffer): string {
-  return String.fromCharCode.apply(null, array as unknown as number[]);
+  return new Uint8Array(array).reduce(function (data, byte) {
```
janek26, line 7: can we have an arrow function here?

janek26 (commented):
Thanks for your contribution and for discovering the bug! One minor comment

src/utils/encode.ts
FawadHa1der: done

janek26:
ah, please base off of develop branch and also contribute back to develop please!

PR #185, opened 2022-05-04
Feat/zondax starware app support

yogh333:


janek26:
Hey @yogh333 
whilst a starknet.js call we decided that ledger should maintain their own package implementing a SignerInterface.
This will help with mitigating dependencies not everybody needs. As an example see the briq guys avoiding the ledger dependencies: https://github.com/briqNFT/briq-builder/blob/main/src/starknet_wrapper.ts
Thanks :)

PR #193, opened 2022-05-26
Adding RPC provider to the library

MilGard91:
- Adding RPC provider interface
- Adding PRC provider class
- Expanding API types with RPC Methods and Responses
- Adding tests for RPC Provider

_* Invoking functions will be added to the rpcAccount_


janek26:
I don't think we should introduce a new interface. An RPC provider should just use the same interface as any other Provider.

0xs34n (commented):
Looks good to me. Just some small optional syntactically comments.

__tests__/rpcProvider.test.ts
```diff
@@ -0,0 +1,138 @@
+import { RPCProvider } from '../src';
+import { compiledArgentAccountRPC } from './fixtures';
+
+describe('RPCProvider', () => {
+  let provider: RPCProvider;
+
+  beforeAll(async () => {
+    provider = new RPCProvider({ nodeUrl: process.env.rpcUrl });
```
janek26, line 8:
```suggestion
    provider = new RPCProvider({ nodeUrl: process.env.TEST_RPC_URL });
```

src/provider/rpcProvider.ts
```diff
@@ -0,0 +1,333 @@
...
+      const successStates = ['ACCEPTED_ON_L1', 'ACCEPTED_ON_L2', 'PENDING'];
+      const errorStates = ['REJECTED', 'NOT_RECEIVED'];
+
+      if (successStates.includes(res.status)) {
+        onchain = true;
+      } else if (errorStates.includes(res.status)) {
+        // const message = res.tx_failure_reason
+        //   ? `${res.tx_status}: ${res.tx_failure_reason.code}\n${res.tx_failure_reason.error_message}`
+        //   : res.tx_status;
+        const error = new Error('Comething went wrong');
```
0xs34n, line 314:
```suggestion
        const error = new Error('Something went wrong');
```

__tests__/rpcProvider.test.ts
```diff
@@ -0,0 +1,138 @@
+import { RPCProvider } from '../src';
+import { compiledArgentAccountRPC } from './fixtures';
```
janek26, line 2:
```suggestion
import { compiledOpenZeppelinAccount } from './fixtures';
```

__tests__/rpcProvider.test.ts
```diff
@@ -0,0 +1,138 @@
...
+        provider.callContract({
+          contractAddress: '0x9ff64f4ab0e1fe88df4465ade98d1ea99d5732761c39279b8e1374fa943e9b',
+          entrypoint: 'balance_of',
+          calldata: ['0x9ff64f4ab0e1fe88df4465ade98d1ea99d5732761c39279b8e1374fa943e9b'],
+        })
+      ).resolves.not.toThrow();
+    });
+    test('deployContract', async () => {
+      const response = await provider.deployContract({
+        contract: compiledArgentAccountRPC,
```
janek26, line 131:
```suggestion
        contract: compiledOpenZeppelinAccount,
```

janek26:
this PR should not change the Provider interface.
If so, we need another major version

When that's the case, we should do that, but still stick with one generic interface for Provider, which can be used by both a normal Provider and an RpcProvider

janek26:
closed in favor of https://github.com/0xs34n/starknet.js/pull/240

PR #195, opened 2022-05-29
fix: allow starknet.js in service workers

avimak:
axios (which uses the `XMLHttpRequest` interface) isn't supported in service workers.
instead, we can use `cross-fetch` which supports single js code for node, browser and react-native

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "@types/url-join": "^4.0.1",
     "@typescript-eslint/eslint-plugin": "^5.0.0",
     "@typescript-eslint/parser": "^5.0.0",
-    "axios-logger": "^2.6.0",
     "eslint": "^7.32.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -58,14 +57,16 @@
     "prettier": "^2.4.1",
     "prettier-plugin-import-sort": "^0.0.7",
     "typedoc": "^0.22.6",
-    "typescript": "^4.4.4"
+    "typescript": "^4.4.4",
+    "whatwg-fetch": "^3.6.2"
```
janek26, line 61: is this needed?

janek26:
great PR, thanks!

package.json
avimak: yes, for `fetch-intercept` to work during the `test` phase

janek26 (approved):

PR #226, opened 2022-06-17
fix(verify): return false when 500 returned from the gateway

1swaraj:
Currently the verifyMessage and the verifyMessageHash functions don't work as intended. 
If there is a 500, the `verifyMessageHash` would still return true
```
public async verifyMessage(typedData: TypedData, signature: Signature): Promise<boolean> {
    const hash = await this.hashMessage(typedData);
    return this.verifyMessageHash(hash, signature);
  }

public async verifyMessageHash(hash: BigNumberish, signature: Signature): Promise<boolean> {
    try {
      await this.callContract({
        contractAddress: this.address,
        entrypoint: 'is_valid_signature',
        calldata: compileCalldata({
          hash: toBN(hash).toString(),
          signature: signature.map((x) => toBN(x).toString()),
        }),
      });
      return true;
    } catch {
      return false;
    }
  } 
```
If you change the signature the test would still pass
```
test('sign and verify offchain message', async () => {
    const signature = await account.signMessage(typedDataExample);
    signature[0] += "123" // change signature to make it invalid
    expect(await account.verifyMessage(typedDataExample, signature)).toBe(true);
  });
```

This PR is a quick fix to this issue


janek26:
may be useful to throw on a deeper level. can we get something in fetchEndpoint?

src/provider/utils.ts
```diff
@@ -49,7 +49,7 @@ export function getBlockIdentifier(blockIdentifier: BlockIdentifier): BlockIdent
   if (blockIdentifier === 'pending') {
     return { type: 'BLOCK_NUMBER', data: 'pending' };
   }
-  if (typeof blockIdentifier === 'number') {
+  if (typeof blockIdentifier === 'number' || typeof blockIdentifier === 'bigint') {
```
1swaraj, line 52: Because there was no status code check from the fetchEndpoint method, the test case 
test(`getBlock(blockHash=undefined, blockNumber=${exampleBlockNumber})`, () => {
      return expect(provider.getBlock(exampleBlockNumber)).resolves.not.toThrow();
    });
passed 
but if you look at the logs, it resolves the blockNumber in this case to blockHash
i.e. http://127.0.0.1:5050/feeder_gateway/get_block?blockHash=0x55

src/provider/default.ts
```diff
@@ -159,7 +159,12 @@ export class Provider implements ProviderInterface {
       body: stringify(request),
       headers,
     })
-      .then((res) => res.text())
+      .then((res) => {
+        if (res.status !== 200) {
```
janek26, line 163: sure it's always 200 and never 201 POST?

```suggestion
        if (res.status >= 400) {
```

src/provider/default.ts
```diff
@@ -170,6 +175,9 @@ export class Provider implements ProviderInterface {
           });
         }
         return parse(res) as Endpoints[T]['RESPONSE'];
+      })
+      .catch((err) => {
+        throw Error(`Could not ${method} from endpoint: ${url} ${err.message}`);
```
janek26, line 180:
```suggestion
        throw Error(`Could not ${method} from endpoint \`${url}\`: ${err.message}`);
```

src/provider/utils.ts
janek26: nice catch

janek26 (approved):

PR #234, opened 2022-06-21
Fix GatewayError error path 

Arcticae:
In testing i found that the error might not be recognised properly with `instanceof` as per this issue:
https://github.com/Microsoft/TypeScript-wiki/blob/main/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work


janek26:
should be fixed by https://github.com/0xs34n/starknet.js/pull/237 ! Let me know if it is not

PR #236, opened 2022-06-23
feat(npm): add npmignore file

LucasLvy:
Add `.npmignore` file to prevent from publishing `.github` `node_modules` `tsconfig.json`

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,7 @@
+.DS_Store
+tsconfig.json
+lib._esm
+tsconfig.tsbuildinfo
+node_modules
+.github/
+.husky/


src/provider/default.ts
```diff
@@ -359,7 +359,8 @@ export class Provider implements ProviderInterface {
    */
   public deployContract(
     payload: DeployContractPayload,
-    _abi?: Abi
+    _abi?: Abi,
+    token?: { token: string }
```
janek26, line 363: this should be part of interface.ts aswell

janek26 (changes requested):
maybe split this into 2 pr's? nom ignore looks good

janek26 (approved):

PR #248, opened 2022-07-11
feat: support SN 0.9.1 estimate_fee response

yoga-braavos:
estimate_fee response was changed. namely 'amount'->'overall_fee'

Added support in a backward-compatible manner per @janek26 suggestion


janek26:
can you maybe make this backwards compatible? we could remove the support for the old max_fee field later on

logic could look something like:
1. check for new field
2. if not there assume that it's the old api

bot cases should result into the same function response

janek26 (approved):

Issue #260, opened 2022-07-18
Hashing an array of structs is not supported


calvwang9:
# Context
I am building a StarkNet dapp, which involves the user signing a payload off-chain using a browser wallet such as ArgentX, then verifying the signature on-chain in a Cairo contract. To do so, I needed to reconstruct the [hashing process](https://github.com/0xs34n/starknet.js/blob/6ee478119f4d33c5bcab72d525fff9fe59500702/src/utils/typedData/index.ts#L166) from starknet.js in Cairo, however I ran into an issue when attempting to hash a struct that has an array of structs as a member.

# Issue
Example struct (`typedData.types`):
```
types: {
    Fee: [
      { name: "a", type: "felt" },
      { name: "b", type: "felt" },
    ],
    MyStruct: [
      { name: "member1", type: "felt" },
      { name: "fees_len", type: "felt" },
      { name: "fees", type: "Fee*" },
    ],
}
```
In the `encodeValue` [function](https://github.com/0xs34n/starknet.js/blob/6ee478119f4d33c5bcab72d525fff9fe59500702/src/utils/typedData/index.ts#L99), if the data type is not a defined type in `typedData.types` or is not `felt*` then it defaults to trying to use the hex value of the data (cast to a string). If data is a list of structs, this cannot be properly converted into a `BN` in the `getHex` func and results in `<BN: 0>`, which in turn results in the hex value `0x0`. This means that any arbitrary list of structs will return the same hash.
```
// type = 'Fee*'
// data = [{a: 1. b: 2}, {a: 3, b: 4}, ... ]

const encodeValue = (typedData: TypedData, type: string, data: unknown): [string, string] => {
  if (typedData.types[type]) {
    // eslint-disable-next-line @typescript-eslint/no-use-before-define
    return [type, getStructHash(typedData, type, data as Record<string, unknown>)];
  }

  if (type === 'felt*') {
    return ['felt*', computeHashOnElements(data as string[])];
  }

  // returns here because type (Fee*) not in typedData.types && type != 'felt*'
  return [type, getHex(data as string)];
};

// toBN(data as string) = <BN: 0>
// toHex(toBN(data as string) = '0x0
```

This was also confirmed by adding a console log in [function](https://github.com/0xs34n/starknet.js/blob/6ee478119f4d33c5bcab72d525fff9fe59500702/src/utils/typedData/index.ts#L120) `encodeData`:
```
...
const [t, encodedValue] = encodeValue(typedData, field.type, value);
[...]

PR #263, opened 2022-07-19
Struct array hashing

calvwang9:
Fixes https://github.com/0xs34n/starknet.js/issues/260 as well as generation of the type hash with struct array dependencies.

Hashes an array of structs like so:
```
// struct s
a = [s1, s2, ..., sn]

// hash each struct individually to get a list of hashes(felts)
[H(s1), H(s2), ..., H(sn)]

// hash list of felts as per usual (computeHashOnElements in starknet.js)
H(a) = H(...H(H(H(s1), H(s2)), H(s3)), ..., H(sn))
```

### DIFF
--- a/__mocks__/typedDataStructArrayExample.json
+++ b/__mocks__/typedDataStructArrayExample.json
@@ -1,0 +1,44 @@
+{
+  "types": {
+    "StarkNetDomain": [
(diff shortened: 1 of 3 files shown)


janek26:
thank you fir your first contribution! 🙂 Very solid and it got tests, appreciated!

PR #267, opened 2022-07-26
fix: RPC provider getEvents

rzmahmood:
The getEvents query on the Pathfinder RPC implementation has optional `address` and `keys` which is more valuable for many use cases. This does unfortunately differ from the RPC spec https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json BUT considering Pathfinder is the largest node implementation, we should support that. I will also be reaching out to Starkware to see if we can update the openrpc spec.

Also fromBlock and toBlock should be numbers and getEventResponse.islastpage should be a boolean 
```json
{
	"jsonrpc":"2.0",
	"method":"starknet_getEvents",
    "params": [
        {"fromBlock": 277645, "toBlock": 277700, "page_size": 1000, "page_number": 1}
    ],
	"id":1
}
```

### DIFF
--- a/src/types/api/rpc.ts
+++ b/src/types/api/rpc.ts
@@ -99,10 +99,10 @@ export namespace RPC {
     | boolean;
 
   export type EventFilter = {
-    fromBlock: string;
-    toBlock: string;
-    address: string;
-    keys: string[];
+    fromBlock: number;
+    toBlock: number;
+    address?: string;
+    keys?: string[];
     page_size: number;
     page_number: number;
   };
--- a/src/types/api/rpc.ts
+++ b/src/types/api/rpc.ts
@@ -110,7 +110,7 @@ export namespace RPC {
   export type GetEventsResponse = {
     events: StarknetEmittedEvent[];
     page_number: number;
-    is_last_page: number;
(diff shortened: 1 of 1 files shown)


janek26:
Thanks mate! Please make sure to follow this up with StarkWare to get the spec right

rzmahmood:
Have followed up with Starkware. They said they can do this and will discuss with nodes/sdks. 

They will update the master branch on the repo soon.

tabaktoni:
made draft until RPC spec update is ready and PR is ready to be reviewed.

tabaktoni:
Pathfinder RPC 0.2.0 eta is 2 weeks from now, so I think we can skip type changes and go straight to 
new continuation_token?
https://github.com/starkware-libs/starknet-specs/blob/438df17c26609caa44eb2e8a34d368e25759fb08/api/starknet_api_openrpc.json#L665

rzmahmood:
Makes sense @tabaktoni . Will close this PR

lior-stark:
Please note that according to the spec, all the properties in the filter are optional.
This is simply the default defined behavior for a JSON schema.

PR #271, opened 2022-07-28
fix: use isomorphic-unfetch

janek26:


0xs34n (approved):
I would prefer more readability on the `fetchPonyfill.ts` but since @ivpavici voted to merge I shall defer to the majority

PR #278, opened 2022-08-03
feat: get-code

janek26:
adds the currently supported `getCode` endpoint for rpc and sequencer
In the future this will be replaced by contract class calls, but that's not supported right now while get_code is supported by sequencer and rpc


0xs34n (approved):
Looks good! Thanks!

PR #290, opened 2022-08-14
fix EventFilter type

coburn24:
change fromBlock and toBlock from string to number

### DIFF
--- a/src/types/api/rpc.ts
+++ b/src/types/api/rpc.ts
@@ -99,8 +99,8 @@ export namespace RPC {
     | boolean;
 
   export type EventFilter = {
-    fromBlock: string;
-    toBlock: string;
+    fromBlock: number;
+    toBlock: number;
     address: string;
     keys: string[];
     page_size: number;


ivpavici (changes requested):
This EventFilter is also present in documentation (provider.md) so please take a look and change there also!

tabaktoni:
What Issue does this resolve?
Can you provide more specifications and descriptions (error, intention, resolution)?

from_block and to_block are block_id in js it is the block identifier, and it can be block number (number), block hash (felt-string), block tag (string).

https://github.com/starkware-libs/starknet-specs/blob/30e5bafcda60c31b5fb4021b4f5ddcfc18d2ff7d/api/starknet_api_openrpc.json#L767

It need to support all types of block indetifiers.

ivpavici:
This was an issue on older version of Pathfinder node, should be fixed with new release (0.3.0 , on Infura) soon:
https://github.com/eqlabs/pathfinder/blob/1e741dcb3bd4d23703c53909706f27082da0bddc/crates/pathfinder/src/rpc/types.rs#L147-L161

PR #300, opened 2022-08-19
fix(RpcProvider): update of RpcProvider getBlock method to work with v0.3.0 of pathfinder

envusr:
with the recent [update of pathfinder](https://github.com/eqlabs/pathfinder/releases/tag/v0.3.0-alpha), some methods have been deleted, making starknet.js not working when using them
e.g. `starknet_getBlockByHash` and `starknet_getBlockByNumber`, that are used in getBlock method of RpcProvider
here is an update using `starknet_getBlockWithTxHashes` instead

### DIFF
--- a/src/provider/rpc.ts
+++ b/src/provider/rpc.ts
@@ -1,30 +1,30 @@
 import { StarknetChainId } from '../constants';
 import {
-  BlockTag,
-  Call,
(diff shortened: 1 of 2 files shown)


envusr:
I'm going to reopen the pr

PR #301, opened 2022-08-19
fix(RpcProvider): update of RpcProvider methods to work with v0.3.0 of pathfinder

envusr:
with the recent [update of pathfinder](https://github.com/eqlabs/pathfinder/releases/tag/v0.3.0-alpha), some methods have been deleted, making starknet.js not working when trying to use them
e.g. `starknet_getBlockByHash` and `starknet_getBlockByNumber`, that are used in getBlock method of starknet.js RpcProvider Class
here is an update using new pathfinder `starknet_getBlockWithTxHashes` method instead of these ones

### DIFF
--- a/src/provider/rpc.ts
+++ b/src/provider/rpc.ts
@@ -22,15 +22,14 @@ import { stringify } from '../utils/json';
 import {
   BigNumberish,
   bigNumberishArrayToDecimalStringArray,
-  isHex,
   toBN,
-  toHex,
-} from '../utils/number';
+  toHex
+  } from '../utils/number';
 import { parseCalldata, parseContract, wait } from '../utils/provider';
 import { RPCResponseParser } from '../utils/responseParser/rpc';
 import { randomAddress } from '../utils/stark';
 import { ProviderInterface } from './interface';
-import { BlockIdentifier } from './utils';
+import { BlockIdentifier, BlockIdentifierClass } from './utils';
 
 export type RpcProviderOptions = { nodeUrl: string };
 
@@ -90,41 +89,21 @@ export class RpcProvider implements ProviderInterface {
   }
 
   public async getBlock(blockIdentifier: BlockIdentifier = 'pending'): Promise<GetBlockResponse> {
+    const blockIdentifierGetter = new BlockIdentifierClass(blockIdentifier);
     const method = 'starknet_getBlockWithTxHashes';
-    if (typeof blockIdentifier === 'string' && isHex(blockIdentifier)) {
-      return this.fetchEndpoint(method, [{ block_hash: blockIdentifier }]).then(
-        this.responseParser.parseGetBlockResponse
-      );
-    }
-    else if (typeof blockIdentifier === 'number') {
-      return this.fetchEndpoint(method, [{ block_number: blockIdentifier }]).then(
-        this.responseParser.parseGetBlockResponse
-      );
-    }
-    else {
-      return this.fetchEndpoint(method, [blockIdentifier]).then(
-        this.responseParser.parseGetBlockResponse
-      );
-    }
+    return this.fetchEndpoint(method, [blockIdentifierGetter.getIdentifier()]).then(
+      this.responseParser.parseGetBlockResponse
+    );
   }
 
-  public async getBlockWithTxs(blockIdentifier: BlockIdentifier = 'pending'): Promise<GetBlockResponse> {
+  public async getBlockWithTxs(
+    blockIdentifier: BlockIdentifier = 'pending'
+  ): Promise<GetBlockResponse> {
+    const blockIdentifierGetter = new BlockIdentifierClass(blockIdentifier);
     const method = 'starknet_getBlockWithTxs';
-    if (typeof blockIdentifier === 'string' && isHex(blockIdentifier)) {
-      return this.fetchEndpoint(method, [{ block_hash: blockIdentifier }]).then(
-        this.responseParser.parseGetBlockResponse
-      );
-    }
-    else if (typeof blockIdentifier === 'number') {
-      return this.fetchEndpoint(method, [{ block_number: blockIdentifier }]).then(
-        this.responseParser.parseGetBlockResponse
-      );
-    }
-    else {
-      return this.fetchEndpoint(method, [blockIdentifier]).then(
-        this.responseParser.parseGetBlockResponse
-      );
-    }
+    return this.fetchEndpoint(method, [blockIdentifierGetter.getIdentifier()]).then(
+      this.responseParser.parseGetBlockResponse
+    );
   }
 
   public async getNonce(contractAddress: string): Promise<any> {
@@ -248,9 +227,8 @@ export class RpcProvider implements ProviderInterface {
     contractAddress: string,
     _blockIdentifier?: BlockIdentifier
   ): Promise<RPC.GetCodeResponse> {
+    // console.log('WARNING: deprecated method, please wait for an update of starknet.js');
 
-    console.log('WARNING: deprecated method, please wait for an update of starknet.js');
-    
     const result = await this.fetchEndpoint('starknet_getCode', [contractAddress]);
 
     return this.responseParser.parseGetCodeResponse(result);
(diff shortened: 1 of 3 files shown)


src/provider/rpc.ts
```diff
@@ -90,14 +90,17 @@ export class RpcProvider implements ProviderInterface {
...
+    if (typeof blockIdentifier === 'string' && isHex(blockIdentifier)) {
+      return this.fetchEndpoint(method, [{ block_hash: blockIdentifier }]).then(
+        this.responseParser.parseGetBlockResponse
+      );
+    }
+    else {
+      return this.fetchEndpoint(method, [{ block_number: blockIdentifier }]).then(
+        this.responseParser.parseGetBlockResponse
+      );
+    }
```
tabaktoni, line 103: by RPC 0.1.0 
method 'starknet_getBlockWithTxHashes'
has parameter 'block_id' that is hash, number or block id ?

https://github.com/starkware-libs/starknet-specs/blob/v0.1.0/api/starknet_api_openrpc.json

ivpavici: blockIdentifier can now be string, number or tag
https://github.com/starkware-libs/starknet-specs/blob/30e5bafcda60c31b5fb4021b4f5ddcfc18d2ff7d/api/starknet_api_openrpc.json#L767

envusr:
### so, here is what we've done:

- @ExoMonk implemented an **update of** the following **RpcProvider methods** so that they fit the **new version of pathfinder** (v0.3.0):  `getTransactionCount`, `getBlock` (we could not use 'pending' and 'latest' as parameters in my version)
- @ExoMonk changed the type of variables `fromBlock `and `toBlock` of the **EventFilter** object, which is used in **getEvents**
- @ExoMonk implemented this **new RpcProvider method**: `getBlockWithTxs`, which works exactly **like getBlock**, except that it gives **more information than just the hash** for each **transaction**
- @envusr implemented this **new RpcProvider method**: `getNonce`, pretty clear  - _still in progress_
- @envusr added a **disclaimer** in `getCode` RpcProvider method, as we **cannot update it** at the moment (no similar query available in pathfinder)


### please note
the methods we've added still need to be added to the Provider class (as we can create an RpcProvider object  using Provider keyword)
take them as suggested features that you could now add to starknet.js easily, implemented because they can be useful and because pathfinder v0.3.0 allows us to create them

also note that we did not delete `getCode` for similar reasons

src/provider/rpc.ts
```diff
@@ -90,14 +90,45 @@ export class RpcProvider implements ProviderInterface {
   }
 
   public async getBlock(blockIdentifier: BlockIdentifier = 'pending'): Promise<GetBlockResponse> {
-    const method =
-      typeof blockIdentifier === 'string' && isHex(blockIdentifier)
-        ? 'starknet_getBlockByHash'
-        : 'starknet_getBlockByNumber';
+    const method = 'starknet_getBlockWithTxHashes';
```
ivpavici, line 93: I think this is more complex than it should be. I think we can remove these if-elses (since it's mostly copy-paste anyway) and have just one:  
``` js
return this.fetchEndpoint('starknet_getBlockWithTxHashes', ...).then(
        this.responseParser.parseGetBlockResponse
 );
```
... and let the client send block_hash, block_number... Btw shouldn't it be just `block_id` instead of  block_hash, block_number? Specification states only  `block_id` as the name

src/provider/rpc.ts
```diff
@@ -90,14 +90,45 @@ export class RpcProvider implements ProviderInterface {
...
+      return this.fetchEndpoint(method, [blockIdentifier]).then(
+        this.responseParser.parseGetBlockResponse
+      );
+    }
+  }
 
-    return this.fetchEndpoint(method, [blockIdentifier]).then(
-      this.responseParser.parseGetBlockResponse
-    );
+  public async getBlockWithTxs(blockIdentifier: BlockIdentifier = 'pending'): Promise<GetBlockResponse> {
```
ivpavici, line 111: Same as above

src/provider/rpc.ts
```diff
@@ -270,10 +304,16 @@ export class RpcProvider implements ProviderInterface {
   public async getTransactionCount(
     blockIdentifier: BlockIdentifier
   ): Promise<RPC.GetTransactionCountResponse> {
-    if (typeof blockIdentifier === 'number') {
-      return this.fetchEndpoint('starknet_getBlockTransactionCountByNumber', [blockIdentifier]);
+    const method = 'starknet_getBlockTransactionCount';
```
ivpavici, line 307: Same as in previous comments

envusr:
we actually cannot do that since the fetcher needs:
- a string if the blockIdentifier is 'latest' or 'pending'
- an object if the blockIdentifier is a number or a hash with only this key:
   1. block_hash if the blockIdentifier is the block hash
   2. block_number if the blockIdentifier is the number of the block
 
as parameter

this way, fetchEndpoint would send a query with a number/string parameter when blockIdentifier is the block number/hash

ivpavici:
> we actually cannot do that since the fetcher needs:
> 
> * a string if the blockIdentifier is 'latest' or 'pending'
> * an object if the blockIdentifier is a number or a hash with only this key:
>   
>   1. block_hash if the blockIdentifier is the block hash
>   2. block_number if the blockIdentifier is the number of the block
> 
> as parameter
> 
> this way, fetchEndpoint would send a query with a number/string parameter when blockIdentifier is the block number/hash

Disussed this in DM-s, so this comment is outdated now

envusr:
so i finally decided to fix it by creating a new class with a getIdentifier method because the modification of the BlockIdentifier type would have required to modify also each use of this type, especially those for the sequencer

src/provider/rpc.ts
```diff
@@ -90,16 +89,27 @@ export class RpcProvider implements ProviderInterface {
   }
 
   public async getBlock(blockIdentifier: BlockIdentifier = 'pending'): Promise<GetBlockResponse> {
-    const method =
-      typeof blockIdentifier === 'string' && isHex(blockIdentifier)
-        ? 'starknet_getBlockByHash'
-        : 'starknet_getBlockByNumber';
+    const blockIdentifierGetter = new BlockIdentifierClass(blockIdentifier);
+    const method = 'starknet_getBlockWithTxHashes';
```
ivpavici, line 93: No need for this method variable anymore

src/provider/rpc.ts
```diff
@@ -90,16 +89,27 @@ export class RpcProvider implements ProviderInterface {
...
+      this.responseParser.parseGetBlockResponse
+    );
+  }
 
-    return this.fetchEndpoint(method, [blockIdentifier]).then(
+  public async getBlockWithTxs(
+    blockIdentifier: BlockIdentifier = 'pending'
+  ): Promise<GetBlockResponse> {
+    const blockIdentifierGetter = new BlockIdentifierClass(blockIdentifier);
+    const method = 'starknet_getBlockWithTxs';
```
ivpavici, line 103: No need for this method variable anymore

src/provider/rpc.ts
```diff
@@ -270,10 +282,9 @@ export class RpcProvider implements ProviderInterface {
   public async getTransactionCount(
     blockIdentifier: BlockIdentifier
   ): Promise<RPC.GetTransactionCountResponse> {
-    if (typeof blockIdentifier === 'number') {
-      return this.fetchEndpoint('starknet_getBlockTransactionCountByNumber', [blockIdentifier]);
-    }
-    return this.fetchEndpoint('starknet_getBlockTransactionCountByHash', [blockIdentifier]);
+    const blockIdentifierGetter = new BlockIdentifierClass(blockIdentifier);
+    const method = 'starknet_getBlockTransactionCount';
```
ivpavici, line 286: No need for this method variable anymore

src/provider/utils.ts
```diff
@@ -31,6 +31,27 @@ export function txIdentifier(txHash?: BigNumberish, txId?: BigNumberish): string
 // hex string and BN are detected as block hashes
 // decimal string and number are detected as block numbers
 // null appends nothing to the request url
```
ivpavici, line 33: these comments belong to the `BlockIdentifier` below... please rearrange this
@tabaktoni can add more comments about this approach. I generally would avoid suing new() inside of methods and use dependency injection... if it is even possible here

src/provider/utils.ts
```diff
@@ -31,11 +31,33 @@ export function txIdentifier(txHash?: BigNumberish, txId?: BigNumberish): string
...
+
 export type BlockIdentifier = BlockNumber | BigNumberish;
 type BlockIdentifierObject =
   | { type: 'BLOCK_NUMBER'; data: BlockNumber }
   | { type: 'BLOCK_HASH'; data: BigNumberish };
 
+
+export class BlockIdentifierClass {
```
ivpavici, line 41: Come to think of it, couldn't we have just used the existing getBlockIdentifier function from utils? Maybe would need to be edited a bit, BLOCK_HASH to lower case etc

src/provider/rpc.ts
```diff
@@ -217,6 +225,8 @@ export class RpcProvider implements ProviderInterface {
     contractAddress: string,
     _blockIdentifier?: BlockIdentifier
   ): Promise<RPC.GetCodeResponse> {
+    console.log('WARNING: deprecated method, please wait for an update of starknet.js');
```
ivpavici, line 228: Should remove this console log, maybe just leave a todo

ivpavici:
Missing unit tests, but maybe we can do a separate PR for those

tabaktoni (approved):
Thank you for updates. This looks good to me. We will tackle block Identifier in more details as it would be nice to unify it with utils for both providers in separate pr.

PR #302, opened 2022-08-23
Create pull_request_template.md

tabaktoni:


ivpavici (commented):
LGTM

.github/pull_request_template.md
```diff
@@ -0,0 +1,28 @@
+## Motivation and Resolution
+...
+
+### RPC version
```
janek26, line 4: should this be mandatory ? The change could also come from something different than an roc version change

ivpavici: True, could add next to it something like "if applicable"

ivpavici: [cd7a8ac](https://github.com/0xs34n/starknet.js/pull/302/commits/cd7a8ac2e049595d0bc78a5c4cd7c3cd0582c8df)

janek26 (approved):

PR #307, opened 2022-08-25
Breaking test Issue Example RPC common interface

tabaktoni:
## Motivation and Resolution
Test not passig for getBlock(blockIdentifier=latest) because RPC diverged from Sequncer result.
RPC can return now different results, in one case it returns param:
```timestamp // The time in which the block was created```
Sequencer return always the same result with param
```accepted_time // When the block was accepted on L1```

GetBlock RPC results in two cases:
1.BLOCK_WITH_TX_HASHES
- status
- BLOCK_HEADER
  - block_hash
  - parent_hash
  - block_number
  - new_root
  - timestamp
  - sequencer_address
- BLOCK_BODY_WITH_TX_HASHES
  - transactions

2.PENDING_BLOCK_WITH_TX_HASHES
- timestamp
- sequencer_address
- parent_hash
- BLOCK_BODY_WITH_TX_HASHES
  - transactions

DIFF:
  accepted_time?: number; 		// SEQUENCER
  timestamp?: number;			// RPC IN BOTH CASES
  block_hash: string;			// RPC IN CASE 1, SEQUENCER
  block_number: number;			// RPC IN CASE 1, SEQUENCER
  gas_price: string;			// SEQUENCER
  new_root: string;				// RPC IN CASE 1, SEQUENCER
  old_root?: string;			// SEQUENCER
  parent_hash: string;			// RPC IN BOTH CASES, SEQUENCER
  sequencer: string;			// SEQUENCER
  status: Status;				// RPC IN CASE 1, SEQUENCER
  transactions: Array<string>; 	// RPC IN BOTH CASES, SEQUENCER
  starknet_version?: string;	// SEQUENCER

How to resolve:
- Split Tests and code?
- Common interface. This is just an example but multiple endpoints have the same issue now?

I can't find a solution with a maintainable single interface that has sense with multiple parameters not being optional?

0.1.0

## Usage related changes
The result from getBlock does not have the same interface from RPC as from Sequencer

RPC and Sequencer results diverged
This is not a final solution as GetBlockResponse can't have one single resulting object that can cover all RPC cases and Sequencer case without complicated merging with multiple optional? parameter.

## Checklist:
- [ ] Documented the changes
- [ ] Updated the docs (www)
- [x] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,6 @@ npm-debug.log
 .env
 .DS_Store
 .eslintcache
+.vscode
 coverage
 yarn.lock
--- a/__tests__/defaultProvider.test.ts
+++ b/__tests__/defaultProvider.test.ts
@@ -46,6 +46,16 @@ describe('defaultProvider', () => {
     test(`getBlock(blockHash=${exampleBlockHash}, blockNumber=undefined)`, () => {
       return expect(testProvider.getBlock(exampleBlockHash)).resolves.not.toThrow();
     });
+    test('getBlock(blockIdentifier=latest)', async () => {
+      expect(exampleBlock).not.toBeNull();
+
+      const { block_number, timestamp } = exampleBlock;
+
+      expect(typeof block_number).toEqual('number');
+
+      return expect(typeof timestamp).toEqual('number');
+    });
+
     test('getBlock() -> { blockNumber }', async () => {
       const block = await testProvider.getBlock();
       return expect(block).toHaveProperty('block_number');
--- a/__tests__/defaultProvider.test.ts
+++ b/__tests__/defaultProvider.test.ts
@@ -46,17 +46,6 @@ describe('defaultProvider', () => {
     test(`getBlock(blockHash=${exampleBlockHash}, blockNumber=undefined)`, () => {
       return expect(testProvider.getBlock(exampleBlockHash)).resolves.not.toThrow();
     });
-
-    test('getBlock(blockIdentifier=latest)', async () => {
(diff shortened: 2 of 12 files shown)


0xs34n:
I just had a great conversation with @tabaktoni about how to move forward. We discussed solutions and this was the optimal solution we came up with:

- Split `defaultProvider.test.ts` to `rpcProvider.test.ts` and `sequencerProvider.test.ts` 
- As for the shared interface `src/provider/interface.ts` we will change the interface to accept Sequencer response OR RPC response. 

For example: 
`public abstract getBlock(blockIdentifier: BlockIdentifier): Promise<GetBlockResponse>;` 

will now be

`public abstract getBlock(blockIdentifier: BlockIdentifier): Promise<GetSequencerBlockResponse | GetRPCBlockResponse>;` 

(exact naming/implementation will be different).

Any comments please let us know!

janek26:
I think splitting the common interface like this should be a total exception! Even if it's needed in this case, I think we shouldn't make it a common pattern.

Lets talk to Starkware to ask them if they could add the missing field to the validator
Or add the missing field to the RPC. wdyt?

ivpavici:
> I think splitting the common interface like this should be a total exception! Even if it's needed in this case, I think we shouldn't make it a common pattern.
> 
> Lets talk to Starkware to ask them if they could add the missing field to the validator Or add the missing field to the RPC. wdyt?

We had a call and agreed on a way forward! Toni will add the agreed changes for all to see

PR #309, opened 2022-08-26
fix: supply calldata as hexadecimal string array

### LINKED ISSUE #285: `deployContract` of the `RpcProvider` throws an error
From Consensys:

When using the "deployContract" function of the "RpcProvider" class, we always get the following error:

```
Error: -32602: More than 64 digits found: 76 at line 1 column 147
  at RpcProvider.eval (eval at <anonymous> (eval at <anonymous> (lockdown.umd.min.js:1:18455)), <anonymous>:70169:35)
```

I think the root cause is that inside the function, `constructorCalldata` was converted to decimal string and passed to `starknet_addDeployTransaction`, but they should instead be in hex string...

dkillen:
## Motivation and Resolution

The StarkNet OpenRPC API spec defines a FELT as a "...field element. Represented as up to 63 hex digits and leading 4 bits zeroed." The RPC write api defines three methods which take an array of FELT as calldata. The RpcProvider has three methods which incorrectly converts an array of BigNumbersih values into a decimal string array rather than an hexadecimal string array to pass as call data. Those methods are:

* getEstimateFee()
* deployContract()
* invokeFunction()

This is the cause of the error described in issue #285

* Adds new utility method bigNumberishArrayToHexadecimalStringArray() in utils/number.ts to return an hexadecimal string array from an array of BigNumberish values
* Replaces the use of bigNumberishArrayToDecimalStringArray() in the three methods mentioned with bigNumberishArrayToHexadecimalStringArray()
* Adds two tests for the RpcProvider test suite rpcProvider.tests.ts

Acknowledging the notice that invokeFunction() is deprecated, I have not added a test for it to the test suite.

Please feel free to make any changes to this PR, including the addition of a test for invokeFucntion(), if you deem it necessary.

Fixes #285

## Usage related changes

Changes should be transparent to users.

- Added 2 new tests to the RpcProvider test suite.

## Checklist:
- [X] Linked the issues which this PR resolves
- [ ] Documented the changes
- [ ] Updated the docs (www)
- [X] Updated the tests
- [X] All tests are passing

### DIFF
--- a/__tests__/rpcProvider.test.ts
+++ b/__tests__/rpcProvider.test.ts
@@ -1,11 +1,33 @@
-import { RpcProvider } from '../src';
-import { describeIfRpc, getTestProvider } from './fixtures';
+import { isBN } from 'bn.js';
+
+import { Account, RpcProvider, ec } from '../src';
+import {
+  compiledErc20,
+  compiledOpenZeppelinAccount,
+  describeIfRpc,
+  getTestAccount,
+  getTestProvider,
+} from './fixtures';
 
 describeIfRpc('RPCProvider', () => {
-  let provider: RpcProvider;
+  const provider = getTestProvider() as RpcProvider;
+  const account = getTestAccount(provider);
+  let erc20Address: string;
+  let accountPublicKey: string;
 
   beforeAll(async () => {
-    provider = getTestProvider() as RpcProvider;
+    expect(account).toBeInstanceOf(Account);
+
+    const erc20Response = await provider.deployContract({
+      contract: compiledErc20,
+    });
+
+    erc20Address = erc20Response.contract_address;
+
+    await provider.waitForTransaction(erc20Response.transaction_hash);
+
+    const accountKeyPair = ec.genKeyPair();
+    accountPublicKey = ec.getStarkKey(accountKeyPair);
   });
 
   describe('RPC methods', () => {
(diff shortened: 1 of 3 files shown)


__tests__/rpcProvider.test.ts
```diff
@@ -1,17 +1,61 @@
...
+    test('getEstimateFee', async () => {
+      const { overall_fee } = await account.estimateFee({
+        contractAddress: erc20Address,
+        entrypoint: 'mint',
+        calldata: [account.address, '1000'],
+      });
+      expect(isBN(overall_fee)).toBe(true);
+    });
+
+    xtest('invokeFunction', async () => {});
```
tabaktoni, line 59: test.todo('invokeFunction')

dkillen: Fixed

ivpavici:
@dkillen we'll wait with this just for a bit, today/tomorrow once we restructure rpc tests and then merge

__tests__/rpcProvider.test.ts
```diff
@@ -1,17 +1,61 @@
...
+        contract: compiledOpenZeppelinAccount,
+        constructorCalldata: [accountPublicKey],
+        addressSalt: accountPublicKey,
+      });
+      await provider.waitForTransaction(transaction_hash);
+      expect(contract_address).toBeTruthy();
+      expect(transaction_hash).toBeTruthy();
+    });
+
+    test('getEstimateFee', async () => {
```
ivpavici, line 50: This test should actually be inside `account.tests.ts` the way it is written (and there already a similar test exists)... but I think the intent was to actually test the rpc `getEstimateFee` function ?

__tests__/rpcProvider.test.ts
```diff
@@ -1,17 +1,61 @@
...
 
   beforeAll(async () => {
-    provider = getTestProvider() as RpcProvider;
+    expect(account).toBeInstanceOf(Account);
+
+    const erc20Response = await provider.deployContract({
+      contract: compiledErc20,
+    });
+
+    erc20Address = erc20Response.contract_address;
```
ivpavici, line 25: This whole erc20 part is redundant (because of https://github.com/0xs34n/starknet.js/pull/309/files#r959545687)
So we can remove them and simplify this whole `beforeAll` part

ivpavici:
@dkillen hi! We merged the changes I mentioned in the comment before... So now, we have some conflicts... if you don't have time I can resolve them and the review comments I left... But I think it would have to be a separate PR, so wanted to see with you first, do you want to resolve it?

ivpavici:
New PR: https://github.com/0xs34n/starknet.js/pull/315 so closing this one
Added you @dkillen as the co-author

Issue #284, opened 2022-08-09
chainId of RpcProvider class not initalized


ivpavici:
From Consensys:
We're recently exploring the use of starknet.js v4.1.0 for calling node API via Infura node, but I found a potential bug that the "chainId" of the "RpcProvider" class is not initalized (remain as undefined) in the class constructor, and the reason is the use of async call inside the constructor:
  
Line 48 from rpc.ts: 

``` javascript
this.getChainId().then((chainId) => {
      this.chainId = chainId;
 });
```
The undefined value of chainId would lead to invalid signature when send a transaction (e.g. via `account.execute()`) given that the txn hash calculation would include the chainId... 

Can you please help take a look with your team to check if this could be a bug or not, thanks a lot!

0xs34n:
I tested this and I can confirm it's undefined 

https://github.com/0xs34n/starknet.js/commit/575f4a737c2782413ec9912ed1281be86c302c9e

jonesho:
@ivpavici I just try to reproduced it in starknet.js v4.3.0 but failed to do so for many attempts, but I think the problem is still here but it's not very easy to reproduce and a bit elusive: e.g. if I have the following codes:
`const provider = new Provider(rpcParam);
  console.log(`**** chainId: ${provider.chainId} ****`);`
I would always see the print log showing chainId as undefined, because the constructor would immediately returned (constructor cannot be async and being await on in Javascript) and then after a fraction of second (depends on how long it takes for receiving the `starknet_chainId` endpoint response) the chainId would be filled with correct value. 

So for our use case with the following call sequence:
`const provider = getProvider(network);
  const account = new Account(provider, senderAddress, senderKeyPair);
  return account.execute(txnInvocation, undefined, {
    maxFee,
  });`
Most of the time, there would be no issue, since chainId is not immediately used (there's an Account contructor call sitting between) and by the time it's being used, it was already initialized.

So I think the issue is still there but only affects use cases where chainId is immediately used after the RpcProvider constructor call, I would recommend to fix it since its elusive and you never know how the developers gonna use the provider's chainId, but its entirely up to your team to decide.

PR #326, opened 2022-09-08
fix: prevent undefined rpc chainId in account

tabaktoni:
## Motivation and Resolution
fix https://github.com/0xs34n/starknet.js/issues/284
1. Issue resolved
RPC constructor retrieves chainId by the async call so calling account estimateFee & execute imediatly after creation could result in failed signerDetails

2. Issue not resolved
This does not resolve issue when you use **provider.chainId** imediatly after construction:
```
    console.log('RpcProvider')
    const provider = new RpcProvider({
        nodeUrl: 'https://starknet-goerli.infura.io/v3/...',
      })
    console.log(`chainId: ${provider.chainId}`); // undefined
```

- One solution is RPC provider constructor return IIF promise and called with
```
let provider = await new RpcProvider({...})
```
- Second would be to use setter and getter, where getter return promise but both are loosely typed with as 'any'. This would require the sequencer Get also return Promise.resolve(this.chainId) for the common interface.
- Third one would be: to remove chaineId as a class attribute and only use getChainId, but the provider for the common interface relies on it to instantiate RPC or sequencer and I was unable to overcome typing limitations.

## Checklist:

- [x] Documented the changes
- [ ] Updated the docs (www)
- [ ] Updated the tests
- [x] All tests are passing

### DIFF
--- a/src/provider/interface.ts
+++ b/src/provider/interface.ts
@@ -22,6 +22,11 @@ import { BlockIdentifier } from './utils';
 export abstract class ProviderInterface {
   public abstract chainId: StarknetChainId;
 
+  /**
+   * Gets the Starknet chain Id
+   *
+   * @returns the chain Id
+   */
   public abstract getChainId(): Promise<StarknetChainId>;
 
   /**
--- a/src/provider/interface.ts
+++ b/src/provider/interface.ts
@@ -22,6 +22,8 @@ import { BlockIdentifier } from './utils';
 export abstract class ProviderInterface {
   public abstract chainId: StarknetChainId;
 
+  public abstract getChainId(): Promise<StarknetChainId>;
+
   /**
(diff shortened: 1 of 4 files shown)


janek26 (approved):

src/provider/interface.ts
```diff
@@ -22,6 +22,8 @@ import { BlockIdentifier } from './utils';
 export abstract class ProviderInterface {
   public abstract chainId: StarknetChainId;
 
+  public abstract getChainId(): Promise<StarknetChainId>;
```
ivpavici, line 25:
```suggestion
  /**
   * Gets the Starknet chain Id
   *
   * @returns the chain Id
   */
  public abstract getChainId(): Promise<StarknetChainId>;
```

ivpavici (approved):

PR #331, opened 2022-09-16
fix right side of a non 2**n merkle tree

gregoryguillou:
## Motivation and Resolution
The length of the proof does not have to be the depth of the tree. I have built a test case to show that is works fine with the corrected algorithm. This PR also corrects the test that was "succeeding" because it was actually a true-negative. I am guessing we will argue on this one so this is what you can do to get conviced:
- rollback the code so that it adds "0x0" as a first argument
- run the last/new test with the "0x0" in the proof: it should not match the merkle root

Another argument is that: [0x1, 0x2, 0x3] would have to have a different root than [0x1, 0x2, 0x3, 0x0] OR you would have to send the length of your data to the contract which you do not have to.

N/A

## Usage related changes
N/A

N/A

## Checklist:

- [ ] Documented the changes
- [ ] Updated the docs (www)
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/utils/merkle.test.ts
+++ b/__tests__/utils/merkle.test.ts
@@ -143,7 +143,7 @@ describe('MerkleTree class', () => {
       expect(proofMerklePath(root, leaf, proof)).toBe(false);
     });
   });
-  describe('data on the right', () => {
+  describe('data on the right with a 2-deep tree', () => {
     let tree: MerkleTree;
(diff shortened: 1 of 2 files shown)


gregoryguillou:
I will restart that PR due to the fact the result is not fixed in size. Thank you @janek26

PR #329, opened 2022-09-12
RPC Improvements

tabaktoni:
## Motivation and Resolution
fix https://github.com/0xs34n/starknet.js/issues/306
fix https://github.com/0xs34n/starknet.js/issues/305
This is still work in progress branch but as it contains missing RPC endpoints it should be good to implement now and subsequential completion done later.

### RPC version (if applicable)
0.1.0

### Added missing RPC methods
- getBlockHashAndNumber
- getClassHashAt
- getPendingTransactions
- getProtocolVersion (with throw error as pathfinder didn't implement it)
- getStateUpdate
- getClass
- traceTransaction (without throw but expect not imeplelemned method resposen)
- traceBlockTransactions (without throw but expect not imeplelemned method resposen)

### Utils Clenup

- BlockIdentifierClass, getIdentifier, getBlockIdentifier and getFormattedBlockIdentifier replaced with single Block Class. It is used both for RPC body identifier params and Sequencer query identifier params.

- Utils test block identifier

```
const block = new Block(blockIdentifier)
rpcparams = block.identifier
sequencerparams = block.queryIdentifier
```

### OpenRPC specification

- [X] Complete OpenRPC 0.1.0 specification
- [X] Complete mapping OpenRpc to RPC
- [X] Refactor current RPC fetch error handling
- [X] Expose RPC fetch as public for custom usage
- [ ] Refactor RPC Methods to OpenRPC method and implementation
- [ ] Refactor RPC fetch and methods to use an object instead of the sorted array.
- [ ] Merge RPC and OpenRPC 
- [X] Add missing test for RPC-specific endpoints
- [ ] Extend Fetch with expected and unexpected error response ex: "{message, code}" vs "1" in some cases
- [ ] Check valid types on all prev. implemented RPC methods (mostly for 'any')
- [ ] Refactor common interface response parser
- [ ] Update documentation

## Checklist:

- [x] Performed a self-review of the code
- [x] Rebased to the last commit of the target branch (or merged it into my branch)
- [x] Linked the issues which this PR resolves
- [ ] Documented the changes
- [ ] Updated the docs (www)
- [x] Updated the tests
- [x] All tests are passing

PR #332, opened 2022-09-16
feat: rpc options, methods, fetch with prop types, imp. types, clenup…

tabaktoni:
## Motivation and Resolution
Continuation on https://github.com/0xs34n/starknet.js/pull/329

0.1.0

## Usage related changes
Enable configure of waitfortrans retries. def 200.
```
new RpcProvider({
  nodeUrl: string;
  retries?: number;
})

```

- [X] Refactor RPC Methods to OpenRPC method and implementation
- [X]  Refactor RPC fetch and methods to use an object instead of the sorted array with types.
- [X] Merge RPC and OpenRPC but kept the separate file for clarity
- [X] Check valid types on all prev. implemented RPC methods (mostly for 'any')
- [X] Refactor common interface response parser (Omitted unnecessary parsers after rpc.0.1.0 impl.)

## Checklist:

- [ ] Documented the changes
- [ ] Updated the docs (www)
- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/provider/rpc.ts
+++ b/src/provider/rpc.ts
@@ -105,24 +105,24 @@ export class RpcProvider implements ProviderInterface {
   public async getBlockWithTxHashes(
     blockIdentifier: BlockIdentifier = 'pending'
   ): Promise<RPC.GetBlockWithTxHashesResponse> {
-    const block = new Block(blockIdentifier);
-    return this.fetchEndpoint('starknet_getBlockWithTxHashes', { block_id: block.identifier });
+    const block_id = new Block(blockIdentifier).identifier;
+    return this.fetchEndpoint('starknet_getBlockWithTxHashes', { block_id });
   }
 
   public async getBlockWithTxs(
     blockIdentifier: BlockIdentifier = 'pending'
   ): Promise<RPC.GetBlockWithTxs> {
-    const block = new Block(blockIdentifier);
-    return this.fetchEndpoint('starknet_getBlockWithTxs', { block_id: block.identifier });
+    const block_id = new Block(blockIdentifier).identifier;
+    return this.fetchEndpoint('starknet_getBlockWithTxs', { block_id });
   }
 
   public async getClassHashAt(
     blockIdentifier: BlockIdentifier,
     contractAddress: RPC.ContractAddress
   ): Promise<RPC.Felt> {
-    const block = new Block(blockIdentifier);
+    const block_id = new Block(blockIdentifier).identifier;
     return this.fetchEndpoint('starknet_getClassHashAt', {
-      block_id: block.identifier,
+      block_id,
       contract_address: contractAddress,
     });
   }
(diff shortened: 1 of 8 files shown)


src/provider/rpc.ts
```diff
@@ -149,48 +158,50 @@ export class RpcProvider implements ProviderInterface {
...
     const block = new Block(blockIdentifier);
-    return this.fetchEndpoint('starknet_getStorageAt', [
-      contractAddress,
-      parsedKey,
-      block.identifier,
-    ]);
+    return this.fetchEndpoint('starknet_getStorageAt', {
+      contract_address: contractAddress,
+      key: parsedKey,
+      block_id: block.identifier,
```
ivpavici, line 164: nitpick: on line https://github.com/0xs34n/starknet.js/pull/332/files#diff-6ae3f9177722bfbaef720fe10fadc424e40d7895481d9a7c09aac78cccf3b17fR134 we have the `block_id` instead of `block.identifier`, so stylistically (consistency) maybe to use only one or the other

src/provider/rpc.ts
```diff
@@ -149,48 +158,50 @@ export class RpcProvider implements ProviderInterface {
...
     return this.getTransactionByHash(txHash).then(this.responseParser.parseGetTransactionResponse);
   }
 
-  public async getTransactionByHash(
-    txHash: BigNumberish
-  ): Promise<RPC.GetTransactionByHashResponse> {
-    return this.fetchEndpoint('starknet_getTransactionByHash', [txHash]);
+  public async getTransactionByHash(txHash: string): Promise<RPC.GetTransactionByHashResponse> {
+    const transaction_hash = txHash;
+    return this.fetchEndpoint('starknet_getTransactionByHash', { transaction_hash });
```
ivpavici, line 175:
```suggestion
    return this.fetchEndpoint('starknet_getTransactionByHash', { transaction_hash: txHash });
```

And remove `const transaction_hash = txHash;` Is this possible?

src/provider/rpc.ts
```diff
@@ -341,7 +353,7 @@ export class RpcProvider implements ProviderInterface {
     blockIdentifier: BlockIdentifier
   ): Promise<RPC.GetTransactionCountResponse> {
     const block = new Block(blockIdentifier);
-    return this.fetchEndpoint('starknet_getBlockTransactionCount', [block.identifier]);
+    return this.fetchEndpoint('starknet_getBlockTransactionCount', { block_id: block.identifier });
```
ivpavici, line 356: Same thing as before for `block.identifier`, to use like this everywhere or use declare `blockId` variable instead of `block`

src/provider/rpc.ts
tabaktoni: https://github.com/0xs34n/starknet.js/pull/332/commits/24c24e8060b76f6a300b21c3148a15183b802fcb

src/provider/rpc.ts
tabaktoni: https://github.com/0xs34n/starknet.js/pull/332/commits/24c24e8060b76f6a300b21c3148a15183b802fcb

src/provider/rpc.ts
tabaktoni: https://github.com/0xs34n/starknet.js/pull/332/commits/24c24e8060b76f6a300b21c3148a15183b802fcb

dhruvkelawala:
https://github.com/0xs34n/starknet.js/blob/09c082e0f3ae6fc3459245861c008537f34e7cbc/__tests__/defaultProvider.test.ts#L224

Let's remove this test condition for now because rpc doesn't return this field while sequencer does. We can also add an if check (`IF_RPC`) for it or remove it completely. `entry_point_selector` is always going to be `__execute__` anyways.

janek26 (approved):

PR #333, opened 2022-09-17
fix issue with fix-size merkle tree

gregoryguillou:
## Motivation and Resolution
This issue was not properly corrected in previous attempt (see https://github.com/0xs34n/starknet.js/pull/331 now closed). The goal of this new PR is to have a merkle proof that is fixed in size and matches the depth of the tree. It would also have to be correct!

1. SessionKey.cairo in the contract project, shows 3 things:
- the proof verification does not treat 0x0 in any specific ways. It would simply be applied
- it does not rebuild the root but just compare it with the proof
- the length of the proof has to be the same for all the calls in the multicall

2. The first step of this PR consists in producing some failing tests, A good one is actually the one with 7 values where we pick 0x5 to start. It shows that:

```
      Array [
        "0x6",
    -   "0x6453d426673d0a0a477bd603579a6aa41c8dc1d862b2138c074be609afeb08e",
    +   "0x7",
        "0x38118a340bbba28e678413cd3b07a9436a5e60fd6a7cbda7db958a6d501e274",
      ]
```

This test it is false because:
- if you choose 0x7, you would have to pad it with 0x0 to have 3 in length
- As a result, if you choose 0x5 or 0x6, **you would also have to hash2 0x7 with 0x0** otherwise, you would not match the previous proof.

The problem with the current algorithm is that the key at the right does not get hashed with 0x0 when it is located on an even index, yet it should. This PR just fix that issue!

3. The fix consists in padding the last key on the leaves with "0x0" if it is on an even index. It actually simplifies the cases as you can see from the code. In addition, all the tests pass including the new ones.

N/A. Only the merkle.ts script from the utils directory has been modified

## Usage related changes
- None

N/A

## Checklist:

- [x] Documented the changes
- [ ] Updated the docs (www)
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/utils/merkle.test.ts
+++ b/__tests__/utils/merkle.test.ts
@@ -53,7 +53,7 @@ describe('MerkleTree class', () => {
           MerkleTree.hash(leaves[0], leaves[1]),
           MerkleTree.hash(leaves[2], leaves[3])
         ),
-        MerkleTree.hash(leaves[4], leaves[5])
+        MerkleTree.hash(MerkleTree.hash(leaves[4], leaves[5]), '0x0')
       );
 
       expect(tree.root).toBe(manualMerkle);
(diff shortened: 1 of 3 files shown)


gregoryguillou:
I did not pay attention to those tests. Let me double-check

gregoryguillou:
@ivpavici can you let the tests back in again? I should have corrected the snapshots on `typedData.test.ts` this time. Yet I would like to make sure!

@janek26 will review this PR next week. On my side, I would like to have it fix to have @argent/x-sessions fix so I will also need on update on that front so there is no rush. 

Last comment, I would prefer we rename `MerkleTree` to something more convenient because it is not the standard Merkle tree algorithm but a merkle tree with a fixed-size proof.

Issue #341, opened 2022-09-22
Merkle tree hash sorts by string rather than int


clexmond:
Currently leaves are sorted in string order prior to hashing. This is at odds with any on-chain verification of a Merkle proof where they will be necessarily sorted as felts / numbers. This results in it becoming increasingly likely a proof will fail the larger it is (i.e. with two unsorted or four sorted leaves there's already a 50% chance the root / proof will be miscalculated).

https://github.com/0xs34n/starknet.js/blob/48ff4eb2cfb868d86a831176d02d17a04b131ee2/src/utils/merkle.ts#L33

PR #343, opened 2022-09-23
fix: fixed hash arg sorting in merkle tree generation

clexmond:
## Motivation and Resolution
Merkle tree generation was relying on a sorted pedersen hash function (as described in #341 ). The sorting function was doing a string compare which can easily fail when comparing to hex strings. For example: `'0xa' > '0x12' = true` when of course `10 > 18 = false`. Given that the vast majority of (all?) on-chain Merkle tree verifiers also sort this causes mismatching roots and proofs with higher and higher likelihood the larger the number of leaves.

This change simply casts the hex strings to BN to compare.

N/A

## Usage related changes

No usage / API changes.

No development related changes.

## Checklist:

- [x] Documented the changes
- [x] Updated the docs (www)
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/src/utils/merkle.ts
+++ b/src/utils/merkle.ts
@@ -1,4 +1,5 @@
 import { pedersen } from './hash';
+import { toBN } from './number';
 
 export class MerkleTree {
   public leaves: string[];
@@ -31,7 +32,14 @@ export class MerkleTree {
   }
 
   static hash(a: string, b: string) {
-    const [aSorted, bSorted] = [a, b].sort();
+    let aSorted = a;
+    let bSorted = b;
+
+    if (toBN(aSorted) > toBN(bSorted)) {
+      aSorted = b;
+      bSorted = a;
+    }
+
     return pedersen([aSorted, bSorted]);
   }
 
--- a/src/utils/merkle.ts
+++ b/src/utils/merkle.ts
@@ -32,14 +32,7 @@ export class MerkleTree {
   }
 
   static hash(a: string, b: string) {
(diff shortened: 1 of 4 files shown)


src/utils/merkle.ts
```diff
@@ -31,7 +32,14 @@ export class MerkleTree {
...
   static hash(a: string, b: string) {
-    const [aSorted, bSorted] = [a, b].sort();
+    let aSorted = a;
+    let bSorted = b;
+
+    if (toBN(aSorted) > toBN(bSorted)) {
+      aSorted = b;
+      bSorted = a;
+    }
+
```
tabaktoni, line 42: Nitpick suggestions
```suggestion
const [aSorted, bSorted] = [toBN(a), toBN(b)].sort((x: any, y: any) => x - y);
```

or in one line

```
  static hash = (a: string, b: string) =>
    pedersen(toBN(a) < toBN(b) ? [toBN(a), toBN(b)] : [toBN(b), toBN(a)]);
```

tabaktoni (approved):
Nice catch! also with dev docs.
Indead functionless sort do it by string order not number

src/utils/merkle.ts
clexmond: I think the first one is good, I adapted it slightly to ensure we avoid any type / casting weirdness. Also realized that even the `>` with BNs had failing edge cases so I fixed and added an additional test case.

ivpavici (approved):

Issue #337, opened 2022-09-22
Declare Transactions v1


tabaktoni:
- declare transaction should now be sent from an account (and is validated using __validate_declare__ in the account contract)
- should remove declare from providers (RPC and sequencer)

> ability to estimate_fee for all tx types that currently have fees
> - l1->l2 message fees is introduces in 0.10.0, I'm considering adding an L1_MESSAGE tx type (this seems to be useful in other places as well, no more weird "l1_origin_message" field in invoke receipt)
> - declare tx fee (for v1.0 declare transaction)

- enable estimate fee on declare transaction v1 from an account contract

PR #353, opened 2022-10-07
Feat/declare tx 0.10

dhruvkelawala:
## Motivation and Resolution

As of starknet 0.10 and with the introduction of transaction v1, all `DECLARE` transaction needs to go through AccountContract which validates this type of transaction using `__validate_declare__` method. 

Upon receiving a declare transaction, the account will first go through validate_declare to make sure that the user approved this declaration. If the validation is successful, then a fee is charged from the account and the new class is added to the state of StarkNet

**v0.2.1**

## Usage related changes

- Change 1.
- Instead of doing `provider.declareContract`, now they are required to do `account.declare` for declaring a class to StarkNet state

## Checklist:

- [x] Documented the changes
- [x] Updated the docs (www)
- [x] Updated the tests
- [x] All tests are passing


dhruvkelawala:
Closes #337

__tests__/sequencerProvider.test.ts
```diff
@@ -19,7 +20,7 @@ describeIfSequencer('SequencerProvider', () => {
     sequencerProvider = getTestProvider() as SequencerProvider;
     customSequencerProvider = new Provider({
       sequencer: {
-        baseUrl: 'https://alpha4.starknet.io',
+        baseUrl: 'http://127.0.0.1:5050/',
```
dhruvkelawala, line 23: This change was made to speed up testing while also showcasing "customSequencerProvider"

src/account/interface.ts
```diff
@@ -29,11 +31,16 @@ export abstract class AccountInterface extends ProviderInterface {
    *
    * @returns response from addTransaction
    */
-  public abstract estimateFee(
+  public abstract estimateInvokeFee(
     calls: Call | Call[],
     estimateFeeDetails?: EstimateFeeDetails
   ): Promise<EstimateFeeResponse>;
 
+  public abstract estimateDeclareFee(
```
ivpavici, line 39: docs here 🙏

src/account/default.ts
```diff
@@ -124,6 +158,49 @@ export class Account extends Provider implements AccountInterface {
...
   }
 
+  public async declare(
+    { classHash, contract }: DeclareContractPayload,
+    transactionsDetail: InvocationsDetails = {}
+  ): Promise<DeclareContractResponse> {
+    const nonce = toBN(transactionsDetail.nonce ?? (await this.getNonce()));
+    let maxFee: BigNumberish = '0';
+
+    if (transactionsDetail.maxFee || transactionsDetail.maxFee === 0) {
```
ivpavici, line 168: maybe here 
```suggestion
    if (transactionsDetail.maxFee && transactionsDetail.maxFee === 0) {
```
?

src/account/default.ts
```diff
@@ -124,6 +158,49 @@ export class Account extends Provider implements AccountInterface {
     );
   }
 
+  public async declare(
+    { classHash, contract }: DeclareContractPayload,
+    transactionsDetail: InvocationsDetails = {}
+  ): Promise<DeclareContractResponse> {
+    const nonce = toBN(transactionsDetail.nonce ?? (await this.getNonce()));
+    let maxFee: BigNumberish = '0';
```
ivpavici, line 166: can this be like:
```suggestion
    let maxFee: ZERO;
```
?

__tests__/sequencerProvider.test.ts
ivpavici, line 23: is this intentional change?

__tests__/account.test.ts
```diff
@@ -15,26 +21,18 @@ describe('deploy and test Wallet', () => {
...
+    const erc20DeployPayload = getERC20DeployPayload(account.address);
+
+    const erc20Response = await provider.deployContract(erc20DeployPayload);
 
     erc20Address = erc20Response.contract_address;
     erc20 = new Contract(compiledErc20.abi, erc20Address, provider);
 
     await provider.waitForTransaction(erc20Response.transaction_hash);
 
-    const mintResponse = await account.execute({
```
ivpavici, line 27: sorry if I missed something, but how do we mint now the 1000 tokens? I only see this later in the test, but unrelated to this part?

src/types/lib.ts
```diff
@@ -1,20 +1,34 @@
...
+}
+
 export type DeployContractPayload = {
   contract: CompiledContract | string;
   constructorCalldata?: RawCalldata;
   addressSalt?: string;
 };
 
 export type DeclareContractPayload = {
+  // The line is commented until we have Class Hash in the compiled contract
```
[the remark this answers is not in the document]
dhruvkelawala: Yes it's a leftover. Thanks for pointing it out

src/provider/sequencer.ts
```diff
@@ -321,21 +321,22 @@ export class SequencerProvider implements ProviderInterface {
...
+      nonce: toHex(toBN(details.nonce)),
+      signature: bigNumberishArrayToDecimalStringArray(signature || []),
+      sender_address: senderAddress,
+      max_fee: toHex(toBN(details.maxFee || 0)),
+      version: toHex(toBN(details.version || 1)),
     }).then(this.responseParser.parseDeclareContractResponse);
   }
 
-  public async getEstimateFee(
+  public async getInvokeEstimateFee(
```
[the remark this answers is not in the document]
dhruvkelawala: Yes, that's a good idea. Let's not introduce a breaking change but keep the `getEstimateFee` which points to `getInvokeEstimateFee` with a deprecated flag and documentation

__tests__/sequencerProvider.test.ts
ivpavici: @dhruvkelawala sorry, forgot about your comment for this line

src/account/default.ts
dhruvkelawala: I copied it from `execute`. However, now that I think of it, it should be something like this.

```suggestion
    if (transactionsDetail.maxFee) {
```

I don't see a point of `=== 0`, in this case.

__tests__/account.test.ts
dhruvkelawala: The new [ERC20 contract](https://github.com/argentlabs/argent-contracts-starknet/blob/develop/contracts/lib/ERC20.cairo) used here calls `_mint` in the constructor and mints 1000 tokens to the `recipient`. Here, `recipient` being the `account`.

src/account/default.ts
dhruvkelawala:
```suggestion
    let maxFee = ZERO;
```
Do you mean this?

__tests__/account.test.ts
ivpavici: ok, maybe it would be good to add a comment about this, to be explicit? Maybe even add a link to this ERC20 contract somewhere, but not sure where, maybe in the docs?

src/account/default.ts
ivpavici: Yeah, sure, if that is ok? Saw being ZERO on another place

__tests__/account.test.ts
dhruvkelawala: Yes

ivpavici (approved):

tabaktoni (approved):
This looks good to me, agree to @ivpavici this non-breaking estimateFee

src/types/lib.ts
```diff
@@ -24,7 +37,7 @@ export type CallDetails = {
 
 export type Invocation = CallDetails & { signature?: Signature };
 
-export type Call = CallDetails & { entrypoint: string };
+export type Call = CallDetails & { entrypoint: BigNumberish };
```
janek26, line 40: entrypoint should only be string and not bignumberish, as the property will get processed using `getSelectorFromName`, which expects human readable selectors

src/utils/hash.ts
```diff
@@ -147,6 +147,38 @@ export function calculateDeployTransactionHash(
...
+): string {
+  let calldata: BigNumberish[] = [];
+  let additionalData: BigNumberish[] = [];
+
+  if (version !== ZERO) {
+    calldata = [classHash];
+    additionalData = [nonce];
+  } else {
+    calldata = [];
+    additionalData = [classHash];
```
janek26, line 167: else case does nothing - not needed

src/utils/hash.ts
```diff
@@ -147,6 +147,38 @@ export function calculateDeployTransactionHash(
...
+  senderAddress: BigNumberish,
+  version: BigNumberish,
+  maxFee: BigNumberish,
+  chainId: StarknetChainId,
+  nonce: BigNumberish
+): string {
+  let calldata: BigNumberish[] = [];
+  let additionalData: BigNumberish[] = [];
+
+  if (version !== ZERO) {
```
janek26, line 162: as this function is new, do we really wanna keep support for old versions?
I would just assume version 1 and remove version 0 logic.

Also keep in mind estimating fees has a different version too!

(Also I'm not sure native operations like `!==` work on non-native BigInts, do they?)

janek26 (approved):
few minor comments, which I think should be addressed before merging :) looks good otherwise

Issue #370, opened 2022-10-24
Support deploy via Universal Deployer Contract


ivpavici:
https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/utils/presets/UniversalDeployer.cairo

https://community.starknet.io/t/universal-deployer-contract-proposal/1864/35

PR #379, opened 2022-10-28
UDC

tabaktoni:
## Motivation and Resolution
Fix https://github.com/0xs34n/starknet.js/issues/370
...

...

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [ ] Documented the changes
- [ ] Updated the docs (www)
- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/account/default.ts
+++ b/src/account/default.ts
@@ -1,4 +1,4 @@
-import { ZERO } from '../constants';
+import { UDC, ZERO } from '../constants';
 import { ProviderInterface, ProviderOptions } from '../provider';
 import { Provider } from '../provider/default';
 import { BlockIdentifier } from '../provider/utils';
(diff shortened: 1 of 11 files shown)


src/account/default.ts
```diff
@@ -228,6 +233,34 @@ export class Account extends Provider implements AccountInterface {
...
+  public async deploy(
+    {
+      classHash,
+      salt,
+      unique = true,
+      callData = [],
+      isDevnet = false,
+    }: UniversalDeployerContractPayload,
+    transactionsDetail: InvocationsDetails = {}
+  ): Promise<any> {
```
ivpavici, line 245: @tabaktoni mentioned about supporting multicalls!

dhruvkelawala: Great idea!

dhruvkelawala: Multicall can't be supported before 0.10.2 because we can't estimate fee of a transaction executed by an undeployed account.

tabaktoni: So I think we can have it but not all use cases will work, like the one you mentioned?

tabaktoni: But one could do multicall deploy true UDC with transactions like deploying Multiple contracts if he already has deployed account. Also i need to figure out how can one chain different transactions? If we can support that (unrelated to this PR)

dhruvkelawala: Yes, 

> So I think we can have it but not all use cases will work, like the one you mentioned?

Yes, if the transactions uses an account that's already deployed, it will work.

PR #367, opened 2022-10-19
Use tsup to build for ESM, CJS and UMD

janek26:
build for ESM, CJS and UMD

## Motivation and Resolution

fixes https://github.com/0xs34n/starknet.js/issues/336

PR #381, opened 2022-10-30
ci: repair alternate format builds

penovicp:
## Motivation and Resolution

Enables usable `iife` and `esm` builds as noted in the PR #367 comments.


package.json
```diff
@@ -3,12 +3,14 @@
...
   "description": "JavaScript library for StarkNet",
   "main": "dist/index.js",
-  "module": "dist/index.esm.js",
+  "module": "dist/index.mjs",
   "browser": "dist/index.global.js",
   "types": "dist/index.d.ts",
   "scripts": {
     "prepare": "npm run build && husky install",
     "build": "tsup",
+    "build:esm": "tsup --clean false --format esm --platform node",
```
janek26, line 12: does this need to be node? wouldnt it work without the plattform parameter and work in browser and node?

src/utils/fetchPonyfill.ts
```diff
@@ -1,4 +1,5 @@
+import isomorphicFetch from 'isomorphic-fetch';
+
 export default (typeof window !== 'undefined' && window.fetch) || // use buildin fetch in browser if available
   (typeof global !== 'undefined' && global.fetch) || // use buildin fetch in node, react-native and service worker if available
-  // eslint-disable-next-line global-require
-  require('isomorphic-fetch'); // ponyfill fetch in node and browsers that don't have it
+  isomorphicFetch; // ponyfill fetch in node and browsers that don't have it
```
janek26, line 5: does this get bundles even if fetch is available?

package.json
penovicp: No, for the `esm` build the `platform` parameter is not necessary, any value will result in the same build. To go into a bit more detail, the `platform` parameter is necessary when bundling the code of the full dependency tree, specifically the `@noble/hashes` sub-dependency through `ethereum-cryptography`. The full bundling is done for the `iife` build, the `esm` and `cjs` builds keep only `import`/`require` references to the direct dependencies, so they don't go deeper than `ethereum-cryptography`.

I left it to have symmetry with `build:iife` and to make its status as an implicit default explicit (not providing any platform value is the same as providing `node`, [link](https://github.com/egoist/tsup/blob/48b3381b15562490870b0cf32f15252038958384/src/esbuild/index.ts#L130)).

src/utils/fetchPonyfill.ts
penovicp: Yes, however it's relatively small, ~150 lines out of 22896 for the `iife` build.

It felt like a smaller cost than trying to refactor the corresponding ponyfill usage to account for using a dynamic import`.

ivpavici: @janek26 what do you think, are we good to merge?

ivpavici (approved):

tabaktoni:
I tested this using svelte app and npm link and it give me issue:
```
✘ [ERROR] [plugin vite:dep-scan] Failed to resolve entry for package "starknet". The package may have incorrect main/module/exports specified in its package.json.
```

penovicp:
> I tested this using svelte app and npm link and it give me issue:
> 
> ```
> ✘ [ERROR] [plugin vite:dep-scan] Failed to resolve entry for package "starknet". The package may have incorrect main/module/exports specified in its package.json.
> ```

I think `vite` tries to use the `esm` entry by default. On this branch `npm run build` creates the `cjs` build, while `build:esm` and `build:iife` create the alternative builds separately as additions. 

Could you try running `npm run build:esm` for `starknet.js`, check that `dist/index.mjs` is present in the symlinked `node_modules` of the svelte app, and rerunning the app?

tabaktoni:
> `build:iife`

Yes Sir This Works.
What would be the best practice here?
Should we create all builds types before deploying?

tabaktoni (approved):
Create a single build with all module types esm, cjs, iife and than on npm release CDN users will get what they need.

PR #393, opened 2022-11-10
RPC testing issue 💥 

tabaktoni:
## Motivation and Resolution
One month ago I noticed this during local testing but ignored it at the moment due to other emerging issues.
It seems that RPC tests are run partially on Sequencer endpoints and all due to failed logic in the parsing of the constructor for the new provider.

...

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [ ] Documented the changes
- [ ] Updated the docs (www)
- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/provider/default.ts
+++ b/src/provider/default.ts
@@ -31,13 +31,20 @@ export class Provider implements ProviderInterface {
   private provider!: ProviderInterface;
 
   constructor(providerOrOptions?: ProviderOptions | ProviderInterface) {
-    if (providerOrOptions && 'chainId' in providerOrOptions) {
-      this.provider = providerOrOptions;
-    } else if (providerOrOptions?.rpc) {
-      this.provider = new RpcProvider(providerOrOptions.rpc);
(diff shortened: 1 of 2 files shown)


penovicp:
I can corroborate the denoted issue and that the supplied changes resolve it.

Here are some minor remarks about possible modifications to the proposed changes:
- It might make sense to make the empty constructor call covered by the `else` explicit with something like `if (providerOrOptions == null) this.provider = new SequencerProvider()` and have the `else` throw an error instead to limit the possibility of potential accidental initializations
- Should the Account constructor accept a default provider? Would something like `if (providerOrOptions?.constructor.name === 'Provider') this.provider = (<Provider>providerOrOptions).provider` be useful?
- Is there a chance someone might want to use a custom provider? It might make sense to swap `implements` with `extends` for the sequencer and PRC providers with relation to the `ProviderInterface` abstract class, then the provider parameter check could use `instanceof ProviderInterface` instead of the constructor name.

PR #397, opened 2022-11-11
fix: use method name similar to gateway endpoint

dhruvkelawala:
## Motivation and Resolution

...

...

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [ ] Documented the changes
- [ ] Updated the docs (www)
- [ ] Updated the tests
- [ ] All tests are passing


__tests__/defaultProvider.test.ts
```diff
@@ -75,43 +75,41 @@ describe('defaultProvider', () => {
...
-      const classResponse = await testProvider.getClassAt(exampleContractAddress);
+      const classResponse = await testProvider.getClassAt(erc20ContractAddress);
 
       expect(classResponse).toHaveProperty('program');
       expect(classResponse).toHaveProperty('entry_points_by_type');
     });
 
     // TODO see if feasible to split
-    describe('getClassHashAt & GetClass', () => {
+    describe(' GetClass', () => {
```
ivpavici, line 90: Maybe:
```suggestion
    describe('GetClassByHash', () => {
```
?

src/provider/rpc.ts
```diff
@@ -194,7 +194,7 @@ export class RpcProvider implements ProviderInterface {
     return this.fetchEndpoint('starknet_getTransactionReceipt', { transaction_hash: txHash });
   }
 
-  public async getClass(classHash: RPC.Felt): Promise<RPC.ContractClass> {
+  public async getClassByHash(classHash: RPC.Felt): Promise<RPC.ContractClass> {
```
tabaktoni, line 197: I would recommend creating an alias function the same as we did with getBlock(), and leaving the original name as is.

tabaktoni (approved):

src/provider/rpc.ts
dhruvkelawala: Just for RPC or for the Provider Interface?

tabaktoni: I thought only RPC as the endpoint is get_class.
I am not sure how many users we have that will be impacted but in general, I think it is good to do if we can soft transition. where we create alias depercate and than remove in couple of versions in the future.
like:
v1:
methodA()

v2:
@deprecated - use methodB() 
methodA()
...
methodB()

vN:
methodB()

PR #400, opened 2022-11-17
feat: add starknet.id getters

### LINKED ISSUE #382: Add a starknet id hook to starknet js
Hey I'm ben from [starknet id. ](https://github.com/starknet-id). We're building the identity service and naming service of starknet.

Some of our partners (realms and cartridge in particular) asked us to make some hooks directly in starknet js to access domain names from a particular address and vice-versa. 

Here is an example of the hook we want to make : 

```
const addressToCheck = '0x2fd23d9182193775423497fc0c472e156c57c69e4089a1967fb288a2d84e914';

const domain = address.getDomainFromAddress(addressToCheck, "goerli");

console.log(domain); // Fricoben.stark
```

In the other sense, we have :

```
const domainToCheck = 'Fricoben.stark';

const address = address.getDomainFromAddress(domainToCheck, "goerli");

console.log(address); // 0x2fd23d9182193775423497fc0c472e156c57c69e4089a1967fb288a2d84e914
```

[Here is the documentation about dev integration of starknet id.](https://docs.starknet.id/for-devs/trustless-integration)

I discussed this with one of you in starknet.cc and Starknet Building (hello Toni) and we can use a few solutions to make it happen. 

1) I can make the PR Myself
2) I can use Starknet Id onlydust credit (if you have someone to propose don't hesitate)

Tell me if it's ok for you !

Thank you for your help,
Don't hesitate to ask questions here or on telegram (@fricoben).

0xBenaparte:
close #382 

I needed to use some of our code as utils (because of [our encoding algorithm](https://docs.starknet.id/for-devs/encoding-algorithm)), so I created my own utils file (these functions will only be used for starknet id so it makes sense to create our own file).

I tried to be as clean as possible and everything works smoothly. Tell me what you think.


src/account/default.ts
```diff
@@ -54,6 +57,53 @@ export class Account extends Provider implements AccountInterface {
...
+        calldata: compileCalldata({
+          address: this.address,
+        }),
+      });
+      const decimalDomain = hexDomain.result
+        .map((element) => new BN(hexToDecimalString(element)))
+        .slice(1);
+
+      const stringDomain = useDecoded(decimalDomain);
+      // const stringDomain = useDecodedSeveral(decimalDomain);
```
ivpavici, line 77: leftover?

src/account/default.ts
```diff
@@ -54,6 +57,53 @@ export class Account extends Provider implements AccountInterface {
...
+
+      return stringDomain;
+    } catch {
+      return Error('Couldnt get stark name');
```
ivpavici, line 81:
```suggestion
      return Error('Couldn't get stark name');
```

src/account/default.ts
```diff
@@ -54,6 +57,53 @@ export class Account extends Provider implements AccountInterface {
...
+        contractAddress: contract,
+        entrypoint: 'domain_to_address',
+        calldata: compileCalldata({
+          domain: [useEncoded(name.replace('.stark', '')).toString(10)],
+        }),
+      });
+
+      return addressData.result[0];
+    } catch {
+      return Error('Couldnt get address from stark name');
```
ivpavici, line 103:
```suggestion
      return Error('Couldn't get address from stark name');
```

src/account/interface.ts
```diff
@@ -219,4 +219,21 @@ export abstract class AccountInterface extends ProviderInterface {
...
+
+  /**
+   * Get starknet.id stark name with the address
+   * @param  {string} StarknetIdContract
+   * @returns name
+   */
+  public abstract getStarkName(StarknetIdContract?: string): Promise<string | Error>;
+
+  /**
+   * Get starknet.id address with the stark name
```
ivpavici, line 231:
```suggestion
   * Get starknet.id address with the stark name
   *
```

src/account/interface.ts
```diff
@@ -219,4 +219,21 @@ export abstract class AccountInterface extends ProviderInterface {
     estimateFeeAction: EstimateFeeAction,
     details: EstimateFeeDetails
   ): Promise<BigNumberish>;
+
+  /**
+   * Get starknet.id stark name with the address
```
ivpavici, line 224: nitpick: new line here

```suggestion
   * Get starknet.id stark name with the address
   *
```

ivpavici:
@0xBenaparte left a couple of cosmetic comments, plus please add the changes in documentation! :)

src/account/default.ts
0xBenaparte: Oh my bad, too many leftovers ^^

src/account/default.ts
0xBenaparte: yep will use `could not` instead

ivpavici (approved):

dhruvkelawala:
I don't think so `starknet.id` should be a part of starknet.js. I think it's a better idea to create a separate lib for all the encoding and decoding.

ivpavici:
@janek26 @dhruvkelawala seems that Ethers.js has support for ENS
https://docs.ethers.io/v5/api/providers/provider/#Provider--ens-methods

janek26:
I'm with Dhruv on this one. I dont think it should be part of starknet.js for now.
We dont know how/if ENS will enter Starknet etc.

In any case I would avoid to change the Provider/Account interface

ivpavici:
What do you think about some compromise, like Ethers has a separate `EnsResolver`  ? https://docs.ethers.io/v5/api/providers/provider/#EnsResolver

Btw I'm sorry this discussion is happening now when the work by @0xBenaparte is already done

0xBenaparte:
I think it would be easier for all starknet developers if it is (Loaf from Realms, @fracek, and Astraly directly asked us to do it).

If ENS enter starknet (starknet.id gonna make a bridge in the future for it) there is still no problem cause you can access them through another hook (```getEnsName``` !== ```getStarkName```). 

To avoid confusion, I could not place it in the account class but more as a util that can be accessed through starknet.js (cause at the end of the day considering the [popularity](https://twitter.com/Starknet_id) of starknet.id right now, it'll be util for a lot of devs).

Tell me what you think.

0xBenaparte:
Hey @janek26,

As discussed with @ivpavici I removed my changes to the account interface. Tell me it's ok for you.

src/utils/starknetId.ts
```diff
@@ -0,0 +1,81 @@
...
+      encoded = encoded.add(multiplier.mul(new BN(newid)));
+      multiplier = multiplier.mul(bigAlphabetSize);
+    }
+  }
+
+  return encoded;
+}
+
+export function getStarknetIdContract(chainId: StarknetChainId): string {
+  const starknetIdMainnetContract = '';
```
janek26, line 74: Should this stay an empty string? The user would get an error telling them that the name could not be resolved etc.
I think we should probably throw right here and have a meaningfull error message like `Starknet.id is not yet deployed on this network` or something

src/utils/starknetId.ts
```diff
@@ -0,0 +1,81 @@
...
+  const starknetIdTestnetContract =
+    '0x05cf267a0af6101667013fc6bd3f6c11116a14cda9b8c4b1198520d59f900b17';
+
+  return StarknetChainId.MAINNET === chainId
+    ? starknetIdMainnetContract
+    : starknetIdTestnetContract;
```
janek26, line 80: can we explicitly match the chainIds?
is it the same address on testnet and testnet2?
Please throw an error if chainId was not recognised

PS keep in mind testnet2 has changed (or will change soon) to a different chainID

src/utils/starknetId.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+      } else decoded += basicAlphabet[code];
+    }
+    decoded += '.';
+  });
+
+  return decoded.concat('stark');
+}
+
+export function useEncoded(decoded: string): BN {
```
janek26, line 39: can we get a test for this?

src/utils/starknetId.ts
```diff
@@ -0,0 +1,81 @@
...
+import { StarknetChainId } from '../constants';
+
+const basicAlphabet = 'abcdefghijklmnopqrstuvwxyz0123456789-';
+const basicSizePlusOne = new BN(basicAlphabet.length + 1);
+const bigAlphabet = '这来';
+const basicAlphabetSize = new BN(basicAlphabet.length);
+const bigAlphabetSize = new BN(bigAlphabet.length);
+const bigAlphabetSizePlusOne = new BN(bigAlphabet.length + 1);
+
+export function useDecoded(encoded: BN[]): string {
```
janek26, line 13: can we get a test for this too?

src/utils/starknetId.ts
```diff
@@ -0,0 +1,81 @@
+/* eslint-disable no-param-reassign */
+import BN from 'bn.js';
+
+import { StarknetChainId } from '../constants';
+
+const basicAlphabet = 'abcdefghijklmnopqrstuvwxyz0123456789-';
+const basicSizePlusOne = new BN(basicAlphabet.length + 1);
+const bigAlphabet = '这来';
```
janek26, line 8: bigAlphabet?

src/utils/starknetId.ts
janek26: so we can catch bugs early and not wait for the integration tests (with network requests) to fail

janek26 (commented):
Yep, happy to merge! Thanks

src/utils/starknetId.ts
0xBenaparte: Actually we deployed on mainnet today so I'll change that quickly

src/utils/starknetId.ts
0xBenaparte: Ok no problem

0xBenaparte:
It is going to be used to allow using non Latin characters but the alphabet extension has not yet be decided.

More info about encoding here ==> https://docs.starknet.id/for-devs/encoding-algorithm

0xBenaparte:
I made all the changes, tell me what you think, @janek26.

Thanks a lot !

src/utils/starknetId.ts
```diff
@@ -0,0 +1,101 @@
...
+      return starknetIdTestnetContract;
+
+    default:
+      throw new Error('Starknet.id is not yet deployed on this network');
+  }
+}
+
+const characters = basicAlphabet.concat(bigAlphabet);
+
+export function generateString(length: number): string {
```
janek26, line 93: as this is just used by the test, can we move it into the test file?
no need to bundle and ship this to every user

janek26 (approved):
looks good! one minor thing which should be done before merge

src/utils/starknetId.ts
0xBenaparte: Yep you're right

PR #402, opened 2022-11-21
Fix Rpc 0.2.1 v1 Transaction and Edpoint

tabaktoni:
## Motivation and Resolution
Infura does not use rpc/v0.2 endpoint, so we restored to open entry for node RPC
The issue raised from Telegram/Slack

0.2.1

## Usage related changes
- nodeUrl should contain the full path to the node providing RPC 0.2.1

## Checklist:

- [ ] Documented the changes
- [ ] Updated the docs (www)
- [ ] Updated the tests
- [ ] All tests are passing


src/provider/rpc.ts
```diff
@@ -242,7 +242,7 @@ export class RpcProvider implements ProviderInterface {
     return this.fetchEndpoint('starknet_estimateFee', {
       request: {
         type: 'INVOKE',
-        contract_address: invocation.contractAddress,
+        sender_address: invocation.contractAddress,
```
tabaktoni, line 245: Note that the entry parameter is contractAddress but used as sender_address.
This is left to be like this because of Sequencer.

penovicp (approved):

ivpavici (approved):

dhruvkelawala (approved):

Issue #405, opened 2022-11-21
replace functions `@ethersproject/bytes`


ivpavici:
Replace most of our utils functions, like: 

- @ethersproject/bytes 

...by using the audited `noble` and `scure` packages by https://github.com/paulmillr (base64, utils like utf8tohex etc) **that are potentially not covered through** https://github.com/0xs34n/starknet.js/issues/407

irisdv:
I'm interested to work on this issue

PR #399, opened 2022-11-17
feat(v5): Native BigInt and faster hashing

dhruvkelawala:
## Motivation and Resolution

The current Pedersen Hash computation is very slow in starknet.js. This PR aims to use the faster hashing package called micro-starknet to improving hashing speeds. Also moved to Native BigInt

## Checklist:

- [x] Performed a self-review of the code
- [x] Rebased to the last commit of the target branch (or merged it into my branch)
- [ ] Linked the issues which this PR resolves
- [x] Documented the changes
- [x] Updated the docs (www)
- [x] Updated the tests
- [x] All tests are passing

PR #413, opened 2022-11-23
fix: tsup config

janek26:
This PR fixes the build process.
The current package published at npm (next) does not work due to bad compiling
This PR should fix that. Tested with cjs and esm imports

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,16 +2,24 @@
   "name": "starknet",
   "version": "4.13.1",
   "description": "JavaScript library for StarkNet",
+  "files": [
+    "dist",
+    "LICENSE"
+  ],
   "main": "dist/index.js",
-  "module": "dist/index.mjs",
-  "browser": "dist/index.global.js",
+  "module": "dist/esm/index.js",
+  "browser": "dist/iife/index.js",
   "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "require": "./dist/index.js",
+      "import": "./dist/esm/index.js",
+      "browser": "./dist/iife/index.js"
+    }
+  },
   "scripts": {
     "prepare": "npm run build && husky install",
-    "build": "tsup && npm run build:esm && npm run build:iife && npm run build:dts",
-    "build:esm": "tsup --clean false --format esm --platform node",
-    "build:iife": "tsup --clean false --format iife --platform browser",
-    "build:dts": "tsup --clean false --dts-only",
+    "build": "tsup",
     "pretest": "npm run lint",
     "test": "jest",
     "posttest": "npm run format",
--- a/package.json
+++ b/package.json
@@ -2,10 +2,6 @@
   "name": "starknet",
   "version": "4.13.1",
   "description": "JavaScript library for StarkNet",
-  "files": [
-    "dist",
-    "LICENSE"
-  ],
   "main": "dist/index.js",
   "module": "dist/esm/index.js",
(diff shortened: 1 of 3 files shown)


dhruvkelawala:
Works! Thanks @janek26

package.json
```diff
@@ -3,15 +3,19 @@
...
+      "browser": "./dist/iife/index.js"
+    }
+  },
   "scripts": {
     "prepare": "npm run build && husky install",
-    "build": "tsup && npm run build:esm && npm run build:iife && npm run build:dts",
-    "build:esm": "tsup --clean false --format esm --platform node",
-    "build:iife": "tsup --clean false --format iife --platform browser",
-    "build:dts": "tsup --clean false --dts-only",
+    "build": "tsup",
```
penovicp, line 18: This and the `tsup.config.ts` change of running all the build outputs together will result in an unusable `iife` build result. The `ethereum-cryptography` dependency has a `@noble/hashes` sub-dependency which necessitates that the different build outputs are run with different values of the `tsup` `platform` parameter, `node` (the default value) for `cjs` and `esm`, `browser` for `iife`. Building `iife` with the default value will result in the code attempting to import the native `node.js` `crypto` module in the browser which will fail (search for `node_modules/@noble/hashes/crypto` in the `iife` build result to find the relevant section of code).

Once issue #405 is resolved (probably when PR #399 is finalized) the build script can be revisited and unified as a single `tsup` invocation.

tsconfig.json
```diff
@@ -48,7 +48,7 @@
 
     /* Emit */
     "declaration": true /* Generate .d.ts files from TypeScript and JavaScript files in your project. */,
-    // "declarationMap": true,                           /* Create sourcemaps for d.ts files. */
+    "declarationMap": true /* Create sourcemaps for d.ts files. */,
```
penovicp, line 51: This setting will be ignored by `tsup`, the `.d.ts` bundling is done by `rollup-plugin-dts` which does not support declaration mapping ([link]( https://github.com/egoist/tsup/issues/311)).

tsconfig.json
```diff
@@ -11,7 +11,7 @@
     // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
 
     /* Language and Environment */
-    "target": "es5" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
+    "target": "ES2020" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
```
penovicp, line 14: This change is fine but won't actually change the build result.

penovicp (changes requested):
> This PR fixes the build process. The current package published at npm (next) does not work due to bad compiling This PR should fix that. Tested with cjs and esm imports

The `cjs` and `esm` build results of this branch don't differ from those found in the published `4.13.1` module. Here are the comparisons:
- [cjs](https://www.diffchecker.com/lrBt8FWt)
- [esm](https://www.diffchecker.com/wkc1Wvmt)

I'm guessing that the issue that motivated this PR arose from some import/bundling tool that did not utilize `main/module/browser` from `package.json` and that adding `exports` resolved it. I suggest first testing if just the `exports` addition with the `index.mjs` and `index.global.js` files would be sufficient, and if it is not then adding the `legacyOutput` parameter.

janek26:
thank you so much for the review, this has been super helpful!
I replaced this PR by #418 which fixes our issues!

PR #426, opened 2022-11-29
Bugfix/split-getnonce-interface

janek26:
## Motivation and Resolution

Hey all,
by using the same `getNonce`
 method name in both the Provider and Account interface it's no longer feasable to extend the Provider for an Account.
As every function accepting a ProviderInterface will accept an account, the method signatures between these two need to be the same.
The AccountInterface should only __extend__ the ProviderInterface

I choose to rename the Provider method rather than the Account method as I think it's the less used interface for the getNonce case, and will avoid pain in the ecosystem adopting this interface change

## Usage related changes

- ProviderInterface will change from `getNonce` to `getNonceForAddress`

## Checklist:

- [x] Documented the changes
- [x] Updated the docs (www)
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,7 +4,6 @@ about: Create a report to help us improve
 title: ''
 labels: 'Type: bug'
 assignees: ''
-
 ---
 
 **Describe the bug**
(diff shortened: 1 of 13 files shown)


ivpavici (approved):

ivpavici:
We need to check the rpc failing tests

janek26:
On it, will push one commit in a few mins.

PR #435, opened 2022-12-02
feat: set testnet2 chain id, cleanup

tabaktoni:
## Motivation and Resolution
Testnet 2 used Testnet chainId,
It will be fixed on 5.11.2022 on testnet-2

- This will change the testnet-2 provider to stop using the testnet chain id and use the testnet-2 chain id
- sequencer provider default options cleanup


src/provider/sequencer.ts
```diff
@@ -64,6 +55,19 @@ export type SequencerProviderOptions =
...
+  return (
+    obj === undefined ||
+    Object.keys(obj).length === 0 ||
+    (Object.keys(obj).length === 1 &&
+      Object.entries(obj).every(([k, v]) => k === 'blockIdentifier' && v === null))
+  );
+}
+
+const defaultOptions: SequencerProviderOptions = {
+  network: 'goerli-alpha-2',
```
ivpavici, line 68: what do you think about moving this string, along with 'mainnet-alpha'  and  'goerli-alpha' to constants file? For example, 'goerli-alpha-2' is magic string in 3 places

ivpavici (approved):

janek26 (approved):
I'll merge, as the upgrade is happening right now

PR #440, opened 2022-12-03
Fix: issue with `calculateStructMembers`

ElijahVlasov:
Fix: issue with `calculateStructMembers`

### DIFF
--- a/src/contract/default.ts
+++ b/src/contract/default.ts
@@ -338,18 +338,13 @@ export class Contract implements ContractInterface {
   }
 
   /**
-   * Deep parse of the object that has been passed to the method
+   * Number of (deep) felt members of a struct.
    *
-   * @param struct - struct that needs to be calculated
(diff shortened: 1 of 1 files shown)


janek26:
will this account for nested structs? and if so, could you put a test?

ElijahVlasov:
> will this account for nested structs? and if so, could you put a test?

This accounts for nested ones. Size of a struct in the abi is its actual word size

ElijahVlasov:
> will this account for nested structs? and if so, could you put a test?

Yes, I'll adapt my code for an example. Similar issue happens with `parseResponseStruct` method. It doesn't parse nested tuples correctly. Should I make a separate PR for that?

tabaktoni:
@ElijahVlasov Can you rebase and add a test for this one so we can close it?

tabaktoni:
Close in favor of https://github.com/0xs34n/starknet.js/pull/498

PR #441, opened 2022-12-04
fix: add deterministic tests

0xBenaparte:
Hey, 

Toni asked me to add deterministic random with a seed on my tests. 

Here it is. I used ``sinus`` to make a pseudo-random function that is deterministic.

I also changed the ``useEncode`` and ``useDecoded`` for better management of non-latin characters.


janek26 (approved):
thank you!

PR #445, opened 2022-12-06
fix: add testnet 2 compatibility to test

0xBenaparte:
Hi again, 

I created a fix considering [this](https://github.com/0xs34n/starknet.js/actions/runs/3624531902/jobs/6111618765) problem on testnet-2 testing. 

I hard-coded the signature cause I thought tests were only run locally, and of course as the account.address was different during testnet-2 tests so the signatures were also different. 

My bad. 

Now it works cause I'm recreating the signature myself with the account.address as an argument.

### DIFF
--- a/__tests__/account.test.ts
+++ b/__tests__/account.test.ts
@@ -2,7 +2,8 @@ import { isBN } from 'bn.js';
 
 import typedDataExample from '../__mocks__/typedDataExample.json';
 import { Account, Contract, Provider, number, stark } from '../src';
-import { feeTransactionVersion } from '../src/utils/hash';
+import { getKeyPair, sign } from '../src/utils/ellipticCurve';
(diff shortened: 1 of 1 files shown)


__tests__/account.test.ts
```diff
@@ -167,7 +168,7 @@ describe('deploy and test Wallet', () => {
       expect(declareTx.class_hash).toBeDefined();
     });
 
-    test('Get the stark name of the account and account from stark name (using starknet.id)', async () => {
+    test.only('Get the stark name of the account and account from stark name (using starknet.id)', async () => {
```
tabaktoni, line 171: I think you need to remove `.only` else other tests will not run ?

tabaktoni (changes requested):

__tests__/account.test.ts
0xBenaparte: Yep my bad, fixed

Issue #317, opened 2022-09-02
Add tests for RPC `getEstimateFee` and `invokeFunction` methods


ivpavici:
Missing in rpcProvider.test.ts

ivpavici:
https://github.com/penovicp test

PR #451, opened 2022-12-12
test: remove todo tests in rpcProvider

irisdv:
## Motivation and Resolution

test #317 
Deleted the todo tests for `getEstimateFee` and `invokeFunction` in `rpcProvider.test.ts` as they should not be called without an account. 

## Checklist:

- [ ] Documented the changes
- [ ] Updated the docs (www)
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/rpcProvider.test.ts
+++ b/__tests__/rpcProvider.test.ts
@@ -128,9 +128,5 @@ describeIfRpc('RPCProvider', () => {
       expect(contractClass).toHaveProperty('program');
       expect(contractClass).toHaveProperty('entry_points_by_type');
     });
-
-    test.todo('getEstimateFee');
-
-    test.todo('invokeFunction');
   });
 });


__tests__/rpcProvider.test.ts
```diff
@@ -128,9 +128,5 @@ describeIfRpc('RPCProvider', () => {
       expect(contractClass).toHaveProperty('program');
       expect(contractClass).toHaveProperty('entry_points_by_type');
     });
-
-    test.todo('getEstimateFee');
```
ivpavici, line 132: One more thing is just to add on these methods a comment to not use them outside of the `Account` cc: @tabaktoni

ivpavici (approved):

__tests__/rpcProvider.test.ts
tabaktoni: This should be added into src/provider/interface.ts and not RPC provider.
I think you can add to @deprecated note also this one:
`@deprecated Please use getInvokeEstimateFee or getDeclareEstimateFee instead. Should not be used outside of Account class`

irisdv: Ah ok, I just updated it

tabaktoni (approved):

PR #453, opened 2022-12-13
test: add ts compilation check to pretest step

penovicp:
## Motivation and Resolution
Adds a type check that also affects tests to improve consistency.

### DIFF
--- a/__tests__/defaultProvider.test.ts
+++ b/__tests__/defaultProvider.test.ts
@@ -1,11 +1,11 @@
-import { BlockNumber, GetBlockResponse, stark } from '../src';
+import { BlockNumber, GetBlockResponse, Provider, stark } from '../src';
 import { toBN } from '../src/utils/number';
 import { encodeShortString } from '../src/utils/shortString';
 import { compiledErc20, erc20ClassHash, getTestAccount, getTestProvider } from './fixtures';
 
 const { compileCalldata } = stark;
 
-const testProvider = getTestProvider();
+const testProvider = new Provider(getTestProvider());
 
 describe('defaultProvider', () => {
   let exampleTransactionHash: string;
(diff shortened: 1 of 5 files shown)


__tests__/defaultProvider.test.ts
```diff
@@ -1,11 +1,11 @@
-import { BlockNumber, GetBlockResponse, stark } from '../src';
+import { BlockNumber, GetBlockResponse, Provider, stark } from '../src';
 import { toBN } from '../src/utils/number';
 import { encodeShortString } from '../src/utils/shortString';
 import { compiledErc20, erc20ClassHash, getTestAccount, getTestProvider } from './fixtures';
 
 const { compileCalldata } = stark;
 
-const testProvider = getTestProvider();
+const testProvider = new Provider(getTestProvider());
```
tabaktoni, line 8: This translates to
`new Provider(new Provider(..))`
Why do we need it like this ?

tabaktoni (changes requested):

__tests__/defaultProvider.test.ts
penovicp: It actually doesn't, `getTestProvider` returns either an `RpcProvider` or `SequencerProvider`, not a `Provider` instance so it equates either to `new Provider(new RpcProvider(..))` or `new Provider(new SequencerProvider(..))`. The tests work since they all conform to the `ProviderInterface`, however, bugs are still possible, e.g. when `getClassHashAt` had flipped parameters between the specific providers and a mapping in the default provider.

We spoke about changing the default provider implementation to not be a wrapping proxy for the specific providers, but until that happens this is the more precise setup for the default provider tests.

tabaktoni: Yea, Good point!

ivpavici (approved):

PR #460, opened 2022-12-16
ci: prune the published module

penovicp:
## Motivation and Resolution

- limits the files included in the published module to only the necessary files
- updates the workflow tests to remove deprecation warnings and prunes unnecessary step elements


tabaktoni (approved):
In God We Thrust.
It's seems legit, if you tested it all it is fine to be merged.

Issue #388, opened 2022-11-09
Support estimate_fee_bulk API 


ivpavici:
Support estimate_fee_bulk API that computes the fee of multiple transactions that will be executed consecutively

It came out with cairo 0.10.2 

https://github.com/starkware-libs/cairo-lang/blob/f951a669e264796d53bf7a3cfd23d5bc1f6e54b8/src/starkware/starknet/services/api/feeder_gateway/feeder_gateway_client.py#L68

yoga-braavos:
@ivpavici , would be great to address https://github.com/0xs34n/starknet.js/issues/391 as part of this..

PR #461, opened 2022-12-16
Feat: Support estimate_fee_bulk API

irisdv:
## Motivation and Resolution

Feat #388
Adds support for `estimate_fee_bulk` API introduced in Cairo 0.10.2

## Usage related changes

- Adds a new function `estimateFeeBulk` in Account class, taking in arguments an array of transactions and returns an array of `EstimateFeeResponse`. Implemented only with sequencer, calling `estimateFeeBulk` with RPC will throw an error.  

## Checklist:

- [x] Documented the changes
- [x] Updated the docs (www)
- [x] Updated the tests
- [x] All tests are passing


ivpavici:
tests fail because we need to update devnet from 0.4.0 to 0.4.2

penovicp:
> tests fail because we need to update devnet from 0.4.0 to 0.4.2

Just to clarify, this means bumping `shardlabs/starknet-devnet:0.4.0-seed0` in the `.github/workflows` files so it can be done as part of this PR.

__tests__/account.test.ts
```diff
@@ -65,6 +65,40 @@ describe('deploy and test Wallet', () => {
     innerInvokeEstFeeSpy.mockClear();
   });
 
+  test('estimate fee bulk', async () => {
+    const res = await account.estimateFeeBulk([
+      {
+        type: 'DECLARE',
```
yoga-braavos, line 71: @irisdv I suggest that on top of this "generic" test, we add a test for the main scenario this bulk endpoint caters which is: 
Txn #1: DEPLOY_ACCOUNT
Txn #2: MULTI_CALL (of INVOKE-s)

src/types/lib.ts
```diff
@@ -77,6 +77,19 @@ export type InvocationsDetails = {
 
 export type InvocationsDetailsWithNonce = InvocationsDetails & { nonce: BigNumberish };
 
+export type TransactionBulk = { type: TransactionType } & (
```
yoga-braavos, line 80: @irisdv why are you calling it "Bulk" if it's a single object? I would move the wrapping in an Array to the type to justify it's "bulk" naming and change function signature to accept TransactionBulk rather than Array<TransactionBulk>

src/types/lib.ts
```diff
@@ -77,6 +77,19 @@ export type InvocationsDetails = {
...
+  | DeclareContractPayload
+  | AllowArray<Call>
+  | DeployAccountContractPayload
+);
+
+export type InvocationBulk = { type: TransactionType } & (
```
yoga-braavos, line 86: @irisdv same comment re "Bulk" semantics as above

src/account/default.ts
```diff
@@ -249,6 +251,162 @@ export class Account extends Provider implements AccountInterface {
...
 
+  public async estimateFeeBulk(
+    transactions: Array<TransactionBulk>,
+    { nonce: providedNonce, blockIdentifier }: EstimateFeeDetails = {}
+  ): Promise<Array<EstimateFee>> {
+    const nonce = toBN(providedNonce ?? (await this.getNonce()));
+    const version = toBN(feeTransactionVersion);
+    const chainId = await this.getChainId();
+
+    const invocations: any = await Promise.all(
```
yoga-braavos, line 262: @irisdv this is a big code dup with the specialized `estimate*Fee` functions, please consider some refactoring so each flavor is managed in a single place in the code and the wrapping functions use that place instead of implementing it on their own

__tests__/account.test.ts
irisdv: When adding this test I realised I have an issue when estimating fees for DEPLOY_ACCOUNT. I have this error when I’m trying to use the function `estimateAccountDeployFee` on its own : `Entry point 0x36fcbf06cd96843058359e1a75928beacfac10727dab22a3972f0af8aa92895 not found in contract with class hash 0x3fcbf77b28c96f4f2fb5bd2d176ab083a12a5e123adeb0de955d7ee228c9854`

Is there a specific way to make it work ? My understanding is that to deploy a new account I need to declare it, send eth to the address then deploy it.

yoga-braavos: The `0x36fcbf06cd96843058359e1a75928beacfac10727dab22a3972f0af8aa92895` selector is `__validate_deploy__` meaning that the contract you tried to deploy does not implement this endpoint and thus `DEPLOY_ACCOUNT` fails. In order for it to work, you need to implement/use a contract that have the `__validate_deploy__` entry-point - See example in OpenZeppelin's account contract:
https://github.com/OpenZeppelin/cairo-contracts/blob/e0e07555b319d8131a35259b55269ba7b19e744e/src/openzeppelin/account/presets/Account.cairo#L112

irisdv: Thanks, I updated the version of Account.json in the repo and added this testcase.

dhruvkelawala:
Let's fix these tests and include it in next v4 release please @irisdv

dhruvkelawala:
Is this PR ready for review?

irisdv:
Yes it's ready for review, I fixed the tests

src/provider/sequencer.ts
```diff
@@ -437,6 +439,51 @@ export class SequencerProvider implements ProviderInterface {
...
+        version: toHex(toBN(invocation?.version || 1)),
+        nonce: toHex(toBN(invocation.nonce)),
+      };
+    });
+
+    return this.fetchEndpoint('estimate_fee_bulk', { blockIdentifier }, params).then((result) => {
+      return [].concat(result as []).map((elem: any) => {
+        return this.responseParser.parseFeeEstimateResponse(elem);
+      });
+    });
```
dhruvkelawala, line 484: Add a separate ResponseParser

src/types/api/sequencer.ts
```diff
@@ -380,5 +381,12 @@ export namespace Sequencer {
       REQUEST: SimulateTransaction;
       RESPONSE: TransactionSimulationResponse;
     };
+    estimate_fee_bulk: {
+      QUERY: {
+        blockIdentifier: BlockIdentifier;
+      };
+      REQUEST: any;
```
dhruvkelawala, line 388: Let's come up with a better typing

src/types/lib.ts
```diff
@@ -77,6 +77,21 @@ export type InvocationsDetails = {
 
 export type InvocationsDetailsWithNonce = InvocationsDetails & { nonce: BigNumberish };
 
+export type TransactionBulk = Array<
+  { type: TransactionType } & {
+    payload: DeclareContractPayload | AllowArray<Call> | DeployAccountContractPayload;
+  }
```
dhruvkelawala, line 83: Needs better typing. This can allow passing `INVOKE_FUNCTION` with DeclareContractPayload arguement

src/types/lib.ts
```diff
@@ -77,6 +77,21 @@ export type InvocationsDetails = {
...
+>;
+
+export type InvocationBulk = Array<
+  { type: TransactionType } & (
+    | Invocation
+    | DeclareContractTransaction
+    | DeployAccountContractTransaction
+  ) &
+    InvocationsDetailsWithNonce & { blockIdentifier: BlockNumber | BigNumberish }
+>;
```
dhruvkelawala, line 93: Same

dhruvkelawala (changes requested):

src/provider/sequencer.ts
```diff
@@ -437,6 +440,49 @@ export class SequencerProvider implements ProviderInterface {
     ).then(this.responseParser.parseFeeEstimateResponse);
   }
 
+  public async getEstimateFeeBulk(
+    invocations: InvocationBulk,
+    blockIdentifier: BlockIdentifier = this.blockIdentifier
+  ): Promise<EstimateFeeResponseBulk> {
+    const params: any = [].concat(invocations as []).map((invocation: any) => {
```
dhruvkelawala, line 447:
```suggestion
    const params: Sequencer.EstimateFeeRequestBulk = invocations.map((invocation) => {
```

dhruvkelawala: Just a small change, we can merge after this

dhruvkelawala (changes requested):

dhruvkelawala:
Thanks for your help @irisdv . Appreciate this!

dhruvkelawala (approved):

Issue #434, opened 2022-12-02
Refactor: Optimise usage of getChainId


tabaktoni:
getChainId() will send an API request every time and store it in chainId, as RPC chainId getter was created to fix the initial chainId load, Account and Contract methods can use chained instead and reduce API requests.

PR #470, opened 2023-01-02
refactor: cache RPC provider chain id

penovicp:
## Motivation and Resolution

Issue #434


tabaktoni (changes requested):
chainId wasn't marked private because of usage in provider constructor.
Now that provider constructor do not depend on chainId, i think we can make it private.

tabaktoni (approved):

PR #483, opened 2023-01-10
fix: repair big number to uint256 conversion method

penovicp:


dhruvkelawala:
Let's close this in favour of #484 ?

PR #494, opened 2023-01-13
Feat: Support get_state_update in provider

### LINKED ISSUE #254: Support `get_state_update`
Provider is missing the `getStateUpdate` method, it should accept the block identifier. 

[RPC spec](https://github.com/starkware-libs/starknet-specs/blob/c18b7650a31b098a626935d82afdc603a5ad46f0/api/starknet_api_openrpc.json#L77)
[Sequencer API](https://github.com/starkware-libs/cairo-lang/blob/167b28bcd940fd25ea3816204fa882a0b0a49603/src/starkware/starknet/services/api/feeder_gateway/feeder_gateway_client.py#L75)

irisdv:
## Motivation and Resolution

Close #254 
Add support for `get_state_update` in Provider (default and sequencer)

## Checklist:

- [x] Documented the changes
- [x] Updated the docs (www)
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/defaultProvider.test.ts
+++ b/__tests__/defaultProvider.test.ts
@@ -82,9 +82,19 @@ describe('defaultProvider', () => {
         return expect(block).toHaveProperty('block_number');
       });
 
-      test('getStateUpdate()', async () => {
-        const stateUpdate = await testProvider.getStateUpdate('latest');
+      test(`getStateUpdate(blockHash=${exampleBlockHash}, blockNumber=undefined)`, async () => {
+        const stateUpdate = await testProvider.getStateUpdate(exampleBlockHash);
         expect(stateUpdate).toHaveProperty('block_hash');
+        expect(stateUpdate.block_hash).toBe(exampleBlockHash);
+        expect(stateUpdate).toHaveProperty('new_root');
+        expect(stateUpdate).toHaveProperty('old_root');
+        expect(stateUpdate).toHaveProperty('state_diff');
+      });
+
+      test(`getStateUpdate(blockHash=undefined, blockNumber=${exampleBlockNumber})`, async () => {
+        const stateUpdate = await testProvider.getStateUpdate(exampleBlockNumber);
+        expect(stateUpdate).toHaveProperty('block_hash');
+        expect(stateUpdate.block_hash).toBe(exampleBlockHash);
         expect(stateUpdate).toHaveProperty('new_root');
         expect(stateUpdate).toHaveProperty('old_root');
         expect(stateUpdate).toHaveProperty('state_diff');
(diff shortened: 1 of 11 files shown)


src/types/api/sequencer.ts
```diff
@@ -362,10 +391,10 @@ export namespace Sequencer {
     };
     get_state_update: {
       QUERY: {
-        blockHash: string;
+        blockIdentifier: BlockIdentifier;
```
penovicp, line 394: `get_state_update` should support `{ blockHash, blockNumber }` as query parameters, with `blockNumber` accepting both block numbers/heights and the pending/latest tags ([link](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/services/api/feeder_gateway/feeder_gateway_client.py#L146)). 

The tests pass on `devnet` because it ignores faulty parameters and returns as if the `latest` block was requested.

__tests__/defaultProvider.test.ts
```diff
@@ -81,6 +81,14 @@ describe('defaultProvider', () => {
         const block = await testProvider.getBlock('latest');
         return expect(block).toHaveProperty('block_number');
       });
+
+      test('getStateUpdate()', async () => {
```
penovicp, line 85: Adjust and expand the test to use the `exampleBlock` values from the `beforeAll` setup step to avoid false positives.

src/types/provider.ts
```diff
@@ -140,3 +145,17 @@ export interface TransactionSimulationResponse {
   trace: TransactionTraceResponse;
   fee_estimation: EstimateFeeResponse;
 }
+
+export interface StateUpdateResponse {
+  block_hash: string;
+  new_root: string;
+  old_root: string;
+  state_diff: {
+    storage_diffs: Array<{
```
penovicp, line 154: It might make sense to use the RPC format as the default and not the sequencer format since it is planned for the RPC methods to outlast the sequencer ones.

penovicp (changes requested):

src/provider/sequencer.ts
```diff
@@ -592,7 +596,19 @@ export class SequencerProvider implements ProviderInterface {
   public async getStateUpdate(
     blockIdentifier: BlockIdentifier = this.blockIdentifier
   ): Promise<StateUpdateResponse> {
-    return this.fetchEndpoint('get_state_update', { blockIdentifier }).then(
+    const args: { blockHash?: string; blockNumber?: BlockNumber } = {};
```
penovicp, line 599: Add a utility getter to the `Block` class, something along the lines of:
```
get sequencerIdentifier() {
    return this.hash !== null 
      ? { blockHash: this.hash } 
      : { blockNumber: this.number ?? this.tag };
  }
```

So it can be used like `const args = new Block(blockIdentifier).sequencerIdentifier` (will probably need some type fiddling) and the `setIdentifier` mapping reused.

penovicp (approved):

PR #496, opened 2023-01-18
feat: add public sequencer fetch method

### LINKED ISSUE #459: Make public Fetch request on Sequencer
Allow users to use starknetjs if missing features or changes to use direct fetch and deal with data themself

penovicp:
## Motivation and Resolution

Close #459

### DIFF
--- a/src/provider/sequencer.ts
+++ b/src/provider/sequencer.ts
@@ -45,10 +45,11 @@ import { GatewayError, HttpError, LibraryError } from './errors';
 import { ProviderInterface } from './interface';
 import { Block, BlockIdentifier } from './utils';
 
-type NetworkName = 'mainnet-alpha' | 'goerli-alpha' | 'goerli-alpha-2';
+export type NetworkName = 'mainnet-alpha' | 'goerli-alpha' | 'goerli-alpha-2';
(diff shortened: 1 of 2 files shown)


src/provider/sequencer.ts
```diff
@@ -202,43 +203,52 @@ export class SequencerProvider implements ProviderInterface {
     const baseUrl = this.getFetchUrl(endpoint);
     const method = this.getFetchMethod(endpoint);
     const queryString = this.getQueryString(query);
-    const headers = this.getHeaders(method);
     const url = urljoin(baseUrl, endpoint, queryString);
 
+    return this.fetch(url, {
+      method,
+      body: request,
```
tabaktoni, line 210: We need a header in fetchEndpoint. (ex. auth)

tabaktoni (changes requested):

src/provider/sequencer.ts
penovicp: Note that this is a call for the provider's `fetch` method not the `fetch` module, the provider's `getHeaders` resolution is applied within it.

tabaktoni: Noted!

tabaktoni (approved):

Issue #497, opened 2023-01-20
`stark.compileCalldata` with Array of struct


Lerwox:
In the documentation, we find the following behavior for compiling data of type Uint256
```js
const transferCallData = stark.compileCalldata({
  to: recipient,
  token_id: { type: 'struct', low: token.low, high: token.high },
});
```

I also read that you can directly pass an Array, and it automatically fills the `{}_len`.
```js
const arr = [1, 2 ,3];
const transferCallData = stark.compileCalldata({
  to: recipient,
  token_ids: arr,
});
```

So I naively thought that I could pass an Array of `{ type: 'struct' ...}`

```js
const arr = [
	{ type: 'struct', low: 1, high: 0 },
	{ type: 'struct', low: 2, high: 0 },
	{ type: 'struct', low: 3, high: 0 },
];
const transferCallData = stark.compileCalldata({
  to: recipient,
  token_ids: { type: 'struct', low: token.low, high: token.high },
});
```

But when I execute, I get an `Uncaught Error Error: Assertion failed` error.  
Am I doing it wrong? Maybe I should transform my list first, but `{}_len` must always be correct, So in the example : 
`{}_len` = 3 and not 6.
Imo, it would be cool to be able to pass directly the filled array 

EDIT : mb,  I misread [this part of the documentation](https://www.starknetjs.com/guides/define_call_message#object), it has all the info I need. But it would still be cool to pass an Array of struct.

tabaktoni:
Please take a look at this PR description and give u feedback (it will resolve your issue, but it is highly probable it will be in v5 because of breaking changes)
https://github.com/0xs34n/starknet.js/pull/498

tabaktoni:
Thank you for your feedback, I am closing now this issue as it is part of mentioned PR changes.

Issue #467, opened 2022-12-22
Exported `stark.compileCalldata` different from `CheckCallData.compileCalldata`


JaniAnttonen:
(Using version 4.17.1): Due to the currently exported `stark.compileCalldata` being different from `CheckCallData.compileCalldata` that gets used in the ContractFactory, deploying contracts using `ContractFactory.deploy` does not work as expected (in some cases).

**To Reproduce**
For us the error is given by running deploy using a ContractFactory with an ABI with a constructor that has the inputs `[{ name: 'signers_len', type: 'felt' }, { name: 'signers', type: 'felt*' }, { name: 'threshold', type: 'felt' }]` and passing the resulting `ContractFactory.deploy` the inputs
```
[
    "1",
    "1",
    "230113552613436439708946390704573322871127834961974877288439088301956112268",
    "1"
]
```
that were acquired by
```
stark.compileCalldata({
    "signers_len": "1",
    "signers": [
        "0x823d5a0c0eefdc9a6a1cb0e064079a6284f3b26566b677a32c71bbe7bf9f8c"
    ],
    "threshold": "1"
})
```

**Expected behavior**
Compiling calldata with `stark.compileCalldata` should result in the same calldata than `CheckCallData.compileCalldata` and pass validation when deploying a contract with a contract factory.

**Desktop (please complete the following information):**

- Brave & Chrome
- Node version 18.12.1
- StarkNet.js version 4.17.1
- Network Goerli testnet 1

NIC619:
I think the problem stems from that `stark.compileCalldata` actually handles the array length for you so when using `stark.compileCalldata` you should not need to pass in value for params like `XXX_len`:

```javascript
stark.compileCalldata({
    "signers": [
        "0x823d5a0c0eefdc9a6a1cb0e064079a6284f3b26566b677a32c71bbe7bf9f8c"
    ],
    "threshold": "1"
})
```

I feel like this is not intuitive at all. When people need to encode call data for the first time, they won't know that the length of an array is taken cared of by the function.

JaniAnttonen:
Nope, that does not fix the bug I'm having, but yes, it seems that `stark.compileCalldata` compiles the array length. The problem comes from the mismatch between input types – getting the error `Error: arg signers should be an Array` with signers_len and without. Problem is that the ContractFactory is checking the inputs against the ABI and trying to recompile them, even though they're already compiled and thus, all arrays are flattened.

tabaktoni:
[...]

Issue #431, opened 2022-12-01
`calculateStructMembers` fails if a struct has a tuple member


ElijahVlasov:
The [code](https://github.com/0xs34n/starknet.js/blob/511548b946f1b431389358d6ec69dffb86e15cd5/src/contract/default.ts#L346) 
```
 private calculateStructMembers(struct: string): number {
    return this.structs[struct].members.reduce((acc, member) => {
      if (member.type === 'felt') {
        return acc + 1;
      }
      return acc + this.calculateStructMembers(member.type);
    }, 0);
  } 
``` 

fails if the passed structure has members of a tuple type. For instance, if a `member.type == "(felt, felt)"` we get `this.structs["(fel, felt)"]` is `undefined`. 

**Expected behavior**
I think what the function essentially does is compute the size of a struct, so it should be safe to rewrite the func as
```
/**
   * Number of (deep) members of a struct. 
   *
   * @param struct - struct that needs to be calculated
   * @return {number} - number of members for the given struct
   */
  private calculateStructMembers(struct: string): number {
    return this.structs[struct].size;
  }
```
If you agree I can open a pr.

Issue #380, opened 2022-10-28
ETH transfer


Sandeep0322:
/Users/apple/Documents/starknet.js-workshop/node_modules/starknet/dist/provider/sequencer.js:394
                            error = new Error(message);
                                    ^

Error: REJECTED: TRANSACTION_FAILED
Error at pc=0:91:
Got an exception while executing a hint.
Cairo traceback (most recent call last):
Unknown location (pc=0:739)
Unknown location (pc=0:682)
Unknown location (pc=0:358)
Unknown location (pc=0:400)
Unknown location (pc=0:423)

Error in the called contract (0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7):
Error at pc=0:104:
Got an exception while executing a hint.
Cairo traceback (most recent call last):
Unknown location (pc=0:1678)
Unknown location (pc=0:1664)

Error in the called contract (0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7):
Error at pc=0:1344:
An ASSERT_EQ instruction failed: 10:3 != 10:2.
    at SequencerProvider.<anonymous> (/Users/apple/Documents/starknet.js-workshop/node_modules/starknet/dist/provider/sequencer.js:394:37)
    at step (/Users/apple/Documents/starknet.js-workshop/node_modules/starknet/dist/provider/sequencer.js:33:23)
    at Object.next (/Users/apple/Documents/starknet.js-workshop/node_modules/starknet/dist/provider/sequencer.js:14:53)
    at fulfilled (/Users/apple/Documents/starknet.js-workshop/node_modules/starknet/dist/provider/sequencer.js:5:58)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  response: [Object: null prototype] {
    tx_status: 'REJECTED',
    tx_failure_reason: [Object: null prototype] {
      code: 'TRANSACTION_FAILED',
      error_message: 'Error at pc=0:91:\n' +
        'Got an exception while executing a hint.\n' +
        'Cairo traceback (most recent call last):\n' +
        'Unknown location (pc=0:739)\n' +
        'Unknown location (pc=0:682)\n' +
        'Unknown location (pc=0:358)\n' +
        'Unknown location (pc=0:400)\n' +
        'Unknown location (pc=0:423)\n' +
        '\n' +
[...]

PR #498, opened 2023-01-20
refactor: contract interactions

tabaktoni:
## Motivation and Resolution
Fix https://github.com/0xs34n/starknet.js/issues/467, fix https://github.com/0xs34n/starknet.js/issues/497, fix https://github.com/0xs34n/starknet.js/issues/431, fix https://github.com/0xs34n/starknet.js/issues/380, fix https://github.com/0xs34n/starknet.js/issues/311
compileCallData and contratc.compileCallData with abi do not produce the same result because of non-standard type definitions (for example tuple type provided as an array without abi can't be distinguished from array type). To Fix it we must define how types are defined.
After digging inside Contract Class, and detecting multiple issues it came out that it requires refactoring in contract interaction.
Complex named and unnamed tuple structures are now supported

- [x] Document Changes
- [x] Test and Resolve Response Parser and added tuple parsing
- [x] Refactor and test contract.invoke, estimate, populate for raw and compiled data
- [x] Cleanup and finish Tests
- [x] Remove CalculateStructMembers (array tuples)
- [ ] Remove CompileCallData in favor of CallData.compile()
- [x] Contract Call and factory.deploy resolved
- [x] Test and Resolve Request (for all types) and Validation Parser (tuple validation is skipped)
- [x] Resolve Request Suppor for raw params and compiled data
- [x] Support Long Text

### ``Felt``
Most basic Cairo type. It is received from network as a type string and can represent text, number (big number), hex
When sending felt to network users can use ``felt()`` helper (alias of toFelt)
```ts
// this will be encoded
felt('someText');
felt(1); 
felt(toBN('1'))
felt('0x012')
// string number will stay string number '1' -> '1'
felt('1') 
```
When used in the contract method or compile it will be auto-detected
```ts
CallData.compile({
 p1: 'text',
 p2: 1,
 p3: '0x1',
 p4: toBN('1'),
})
// equal to 
CallData.compile({
 p1: felt('text'),
 p2: felt(1),
 p3: felt('0x1'),
 p4: felt(toBN('1')),
})
```
and 
```ts
contract.method('text', 1, '0x01', toBN('1'))
// equal to
contract.method(felt('text'), felt(1), felt('0x01'), toBN('1'))
```
})

### ``Struct``
Cairo Struct is composed Cairo type that can contain felt, nested structs and nested tuples
As a method parameter, it can't contain arrays as Cairo does not allow pointers inside.
Structs are defined as JS Objects
```ts
const struct1 = {
        a: '',
        b: '',
        c: ''
    }

CallData.compile({
    param1: struct1
})
// or 
contract.method(struct1)
```

#### ``uint256`` is a common struct in Cairo used for compatibility reasons with Solidity
user can use helper function uint256() in method parameters:
```ts
CallData.compile({
 p1: uint256('500000'),
})
//or
contract.method(uint256('500000'))
```
It can also be computed and defined manually as:
```ts
  {
    low: '',
    high: '',
  };
```

### ``Tuple``
Cairo Tuples are composed cairo types that can contain felt, nested struct or nested tuple.
In addition to than tuples can be names and unnamed.
As a method parameter, it can't contain arrays as Cairo does not allow pointers inside.
As a method parameter in JS you can use helper ``tuple()`` for unnamed tuple or it can be defined as JS Object for named tuple.
```ts
/* Unnamed tuple */
tuple('text', 125)

// equal to
{
  0: 'text',
  1: 125,
}

/* Named tuple */
{
  a: 'text',
  b: 125
}
```
in method call
```ts
// unnamed tuple
CallData.compile({
 p1: tuple('text', 125),
})
//or
contract.method(tuple('text', 125))

//named tuple
const tuple1 = {
  a: 'text',
  b: 125
 }

CallData.compile({
 p1: tuple1,
})
//or
contract.method(tuple1)

```

### ``Array``
Cairo Array can be array of felts, tuples or struct.
As a method parameter, it can't contain arrays as Cairo does not allow pointers inside, so no multi-dimensional (nested) arrays. When dealing with an array `len_` will be added internally
Also array can be of only one type, no mixing of types.
Cairo Array is defined as JS Array of tuple, felt or struct
```ts
const data = {
  p1: [1, 2, 3, 4, 5, 6],
  p1: [uint256(1000), uint256(2000), uint256(3000), uint256(4000)],
  p3: [
    { discount_fix_bps: 10, discount_transfer_bps: 11 },
    { discount_fix_bps: 20, discount_transfer_bps: 22 },
  ],
};
const compiledData = CallData.compile(data);
contract.method(compiledData);
//or
contract.method(data.p1, data.p2, data.p3);
```

### Contract Methods Arguments
When dealing with cairo contract we need to convert data sent to method from JS format to Cairo compatible format.
For executing contract method user can provide contract data arguments in two ways
As method arguments, and it will be compiled internality.
```ts
contract.method(arg1, arg2, arg3);
```
or compile them and than provide compiled data
```ts
const data = {
  p1: 1,
  p2: tuple(2,3),
  p3: {a: 'a', b: 'a'}
}
const compiledData = CallData.compile(data);
contract.method(compiledData);
```
When compiling with ``CallData.compile()`` response object will have private prop `compiled`
to differentiate from non-compiled data.
You can pass array with manualy compiled data to compile() to mark it as compiled (it will just add compiled prop).

#### Options
Contract method have options:
##### `parseRequest`
If for some reason internal compilation failed user can set `parseRequest: false`,
and provide manually compiled data. It will skip validation and parsing step.
##### `parseResponse`
By default Contract response data will be converted to JS type data with BN.
If user need raw data or response parsing failed it can set `parseResponse: false`
##### `formatResponse` !!Experimental Feature
An object containing a definition of how to convert response BN to JS types.

## Formatter
format response is an experimental feature and it will convert response JS type data with BN to defined types. support string, numbers, and functions (function argument is value to convert). if a type is not converted probably the response data key do not match formatResponse key.
For all unmatched keys data or omitted keys in formatResponse, original data as BN will be returned. As so if the user do not want to convert part of the response just omit it from the formatResponse schema.
For Arrays, it is enough to define one element type (as array can't have mixed types)
Examples ``arr: ['number']``, ``arr: [tuple('string','string')]``, ``arr: [{x: 'number', y: 'number'}]`` 
```ts
const formatResponse = {
      r1: 'string',
      r2: 'number',
      r3: {
        a: 'string',
        b: 'number'
      },
      uint: uint256ToBN,
      uint2: (value) => uint256ToBN(value).toNumber(),
}
const result = await contract.method(p1, p2, p3, { formatResponse });
// instead of { r1: BN(), r2: BN(), r3: {a: BN(), b:BN(), uint: { low:100, high:0 }, uint2: { low:100, high:0 } }}
// it will be { r1: 'text', r2: 1, r3: {a: 'text', b:2}, BN(100), 100}
```

Options:
```ts
const result = await contract.method(p1, p2, p3, {
      parseRequest: true,
      parseResponse: true,
      formatResponse: formatter
    });
```

## Long Text
Long text (more than 31 characters) can be passed to Cairo contract method as felt* (felt array).
From the perspective of a js developer working with short or long string looks the same.
In the case of using fromatter it is enough to define the response object prop as a string.
```ts
const formatResponse = {
  a: 'string',
  b: 'string',
}
const res = contract.method('some short text', 'some super long long long long text....', { formatResponse });
// res.a === 'some short text'
// res.b === 'some super long long long long text....'
```
How it works:
1. The validator will validate if the parameter is long text against abi and throw an exception if abi does not support long string, aka from Cairo contract side method do not accept felt*
2. both .compile() will produce an array of encoded strings
3. formatter will compare schema to response data and recreate the BN array to string.

## Additional Changes:

1. The response changed from Array-Like Object to Object:
The contract method returned an array-like object containing named fields and enumerated fields. The result should be an object the same as the request. Regarding enumerating fields is redundant and creates duplication of responses. Enumerated fields can be easily accessed with simple response conversion for users:
```ts
const response = await erc20.testFunction(params);
const idxResponse = Object.values(response);
```

2. Experimental Formatter:
Call contract method extended with option formatResponse, which formats response data to number and string based on provided schema of response data. See tests 'formatResponse'.

## Usage and Development related changes
- compileCallData is deprecated in favor of CallData.compile !!! compileCallData will be removed. Also `compile` is defined in CallData util with the rest of Cairo class parser, validator, response parser, tuple parser.
- Response From Contract method call changed from Object like array to Object and removed duplicated index data.
- For every cairo type there is now a definition for JS type. 
- Defining tuples as js array is deprecated 
- Full support for nested tuples and struct, check tests for complex examples
- CheckCallData class renamed to CallData. Methods renamed to singular purpose name. CallData split into modules for each singular purpose. (compileCallData -> compile)
- compileCallData moved to CallData with abi based compileCallData and renamed as static method ``CallData.compile``
- removed numbers/toFelt in favor of cairo/felt. felt can be a number or string and this also fixes circular references in strings.
- LongText implemented into contract interaction (validate, compile, parsers, formatter). (aka. longString but string can be string-hex string-number so long text is a better name for a string that is nor hex string nor whole number string).

## Checklist:

- [x] Documented the changes
- [ ] Updated the docs (www)
- [x] Updated the tests
- [x] All tests are passing


tabaktoni:
Dilemma:
As @PhilippeR26 mentioned and this PR contains too many changes (breaking) to be considered a minor release, but I think it also contains useful fixes and improvements for v4 users. As so we can do:
1. A minor release to v4 with a note of breaking changes and from there PR with bigInt to Beta
2. Rebase to the beta branch and implement bigInt for v5. (and expect new users to use v5, as old ones already managed to do contract interactions the old way).

- [ ] Rebase to beta and native BigInts 

@PhilippeR26 @ivpavici @penovicp @janek26 @dhruvkelawala

tabaktoni:
> Dilemma: As @PhilippeR26 mentioned and this PR contains some changes (minor breaking), but I think it also contains useful fixes and improvements for v4 users. As so we can do:
> 
> 1. A minor release to v4 with a note of breaking changes and from there PR with bigInt to Beta
> 2. Rebase to the beta branch and implement bigInt for v5. (and expect new users to use v5, as old ones already managed to do contract interactions the old way).
> 
> * [ ]  Rebase to beta and native BigInts
> 
> @PhilippeR26 @ivpavici @penovicp @janek26 @dhruvkelawala

My opinion is that changes are minor breaking (compile method name and import, contract response object, tuple as array) so we could merge it to develop and add a migration guide. As so when the same user migrates from this v4 to v5 there will be less of changes.

tabaktoni:
After discussion we decided to rebase on Beta and implement only on v5

src/utils/calldata/requestParser.ts
```diff
@@ -0,0 +1,122 @@
...
+  type: string,
+  structs: abiStructs
+): string | string[] {
+  if (element === undefined) {
+    throw Error(`Missing parameter for type ${type}`);
+  }
+  if (Array.isArray(element)) {
+    throw Error(`Array inside array (nD) are not supported by cairo. Element: ${element} ${type}`);
+  }
+  // checking if the passed element is struct
```
ivpavici, line 52: in general, like my two other comments, is just to think about if we can extract some of these portions into a separate function, remove these comments and name the function similar to the comment! 
Maybe it's overkill or unnecessary for most, but just to think about it

tabaktoni: Regarding all splitting  i think we can do additional module cleanup post cari 1 changes when we can do adaptations for cairo1

PR #502, opened 2023-01-27
modify account.md

kongtaoxing:
Fixed the description of the `transactions` variable data type in the `execute` section, and added the `execute` code example

## Checklist:

- [x] Updated the docs (www)

### DIFF
--- a/www/docs/API/account.md
+++ b/www/docs/API/account.md
@@ -153,7 +153,7 @@ The _estimateFeeDetails_ object may include any of:
 
 account.**execute**(transactions [ , abi , transactionsDetail ]) => _Promise < InvokeFunctionResponse >_
 
-Executes one or multiple calls using the account contract.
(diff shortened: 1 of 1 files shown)


www/docs/API/account.md
```diff
@@ -153,7 +153,7 @@ The _estimateFeeDetails_ object may include any of:
 
 account.**execute**(transactions [ , abi , transactionsDetail ]) => _Promise < InvokeFunctionResponse >_
 
-Executes one or multiple calls using the account contract.
+Executes one or multiple calls using the account contract. If there is only one call, _transactions_ will be an object contains parameters below. If there are multiple calls, _transactions_ will be an array contains several objects mentioned above.
```
ivpavici, line 156:
```suggestion
Executes one or multiple calls using the account contract. If there is only one call, _transactions_ will be an object that contains parameters below. If there are multiple calls, _transactions_ will be an array that contains several objects mentioned above.
```

ivpavici (approved):

PR #508, opened 2023-02-01
decodeShortString was not handling properly Integer Strings

### LINKED ISSUE #501: decodeShortString : wrong answer of the function
**Describe the bug**
Hello,
I noted a problem with the function `utils/shortstrings.ts/decodeShortString`. 
If I use it with an hex string ('0x321....456'), everything is fine.
But if the str:string is  an integer string ('1542233...56'), the function accept it, process without fail, and send back a wrong answer.
```typescript
export function decodeShortString(str: string) {
  return removeHexPrefix(str).replace(/.{2}/g, (hex) => String.fromCharCode(parseInt(hex, 16)));
}
```

**To Reproduce the problem :**
```typescript
const myText = "decimal number fail!";
console.log("myText =", myText);
const encodedText: string = shortString.encodeShortString(myText);
console.log("encodedText =", encodedText);
const myTextDecoded = shortString.decodeShortString(encodedText);
console.log("finalString =", myTextDecoded);
const decimalEncoded: string = BigInt(encodedText).toString(10);
console.log("decimalEncoded =", decimalEncoded);
const wrongTextDecoded : string= shortString.decodeShortString(decimalEncoded);
console.log("wrongTextDecoded =", wrongTextDecoded);
```
Result : 
```
myText = decimal number fail!
encodedText = 0x646563696d616c206e756d626572206661696c21
finalString = decimal number fail!
decimalEncoded = 573160112338783013729123589963347449102752902177
wrongTextDecoded = W1`#8x0r#Xc4tI'R
```

**Expected behavior**
This function should handle properly both formats.

**Screenshots**
N/A

- Browser & version : N/A
- Node version : "ts-node": "^10.9.1"
- StarkNet.js version : 4.21.0
- Network [ : N/A

**Additional context**
A PR is ready to send.

PhilippeR26:
Fix #501 

## Motivation and Resolution

The function decodeShortString returns a wrong answer  if the input is a string representing an integer.
See issue #501 for example of bug.

Now, the answer is correct

## Usage related changes

No change for users ; just a correct answer of the function.
Guides have been updated.

Test of this function has been updated. Please check that usage of `toMach` is correct (/__tests__/utils/shortString.test.ts, line 24).

## Checklist:

- [X] Linked the issues which this PR resolves
- [X] Updated the docs (www)
- [X] Updated the tests
- [X] All tests are passing


tabaktoni (approved):

tabaktoni:
Note: I think it would be cool to create encodeString &. decodeString that will auto-detect length and deal with short and long strings, so that end user has no worry about string length.

PhilippeR26:
> Note: I think it would be cool to create encodeString &. decodeString that will auto-detect length and deal with short and long strings, so that end user has no worry about string length.

It could be a problem on Cairo contract side ; some contracts are waiting only a felt (only shortString), some others are waiting only array of felt (long String).
Unfortunately, Cairo is not Typescript : it can not handle several types (as `felt | felt[]` ) and adapt itself in accordance.

www/docs/guides/define_call_message.md
```diff
@@ -202,3 +202,62 @@ When you perform a call, you have the result in an object :
 | Uint256 (256 bits max)                | `func getV()->(balance:Uint256)`           | BN                                | `const balance = uint256.uint256toBN(result.balance)`              |
 | array of felt                         | `func getV()->(list_len:felt, list:felt*)` | BN[]                              | `const list= result.list`                                          |
 | shortString (31 ASCII characters max) | `func getV()->(title:felt)`                | string                            | `const title:string = shortString.decodeShortString(result.title)` |
+
+## Handle Strings :
+
+In Javascript/Typecript, the max length of a string is nearly illimited. In Cairo, a string is limited to only 31 characters, and is called a ShortString.
```
penovicp, line 208: Nitpick, `limitless` is more commonly used in favour of `illimited`.

penovicp (changes requested):
Left a minor nitpick for the content changes.

Aside from that at least one of the commits should be reworded to conform to the [conventional commits format](https://www.conventionalcommits.org/en/v1.0.0/) so the patch number of the version is bumped by the CI. I recommend rebasing and squashing the commits into one with a message such as: `fix: repair decodeShortString for integer strings`.

tabaktoni:
> It could be a problem for encode and send to Cairo contract ; some contracts are waiting only a felt (only shortString), some others are waiting only array of felt (long String). Unfortunately, Cairo is not Typescript : it can not handle several types (as `felt | felt[]` ) and adapt itself in accordance. May be on decode side, if you are sure to recover a shortString or a longString, a function 'decodeString' could handle both cases.

This is more comment on the contract class implementation (I can implement it in a contract).
Agree, but when building call/invoke by contract class we read abi, so we know what parameter it expects felt or felt*, and at the same time we validate it against provided arguments. In case the argument is 'very long string.........' and abi element is felt* I can use a long string. In case it is felt and 'very long string...' i can throw error, in case it is felt and 'short string' all ok, in all other cases throw an error.

www/docs/guides/define_call_message.md
```diff
@@ -202,3 +202,62 @@ When you perform a call, you have the result in an object :
...
+
+the result is : "uri/pict/t38.jpg"
+
+### LongString
+
+How to handle a string with more than 31 characters :
+
+1. The Cairo contract has to manage this string as array of ShortString (array of felt).
+2. The JS code has to split/encode the string before call/invoke.
+3. The JS code hast to decode/merge the BNs received from a call.
```
tabaktoni, line 238:
```suggestion
3. The JS code has to decode/merge the BNs received from a call.
```

__tests__/utils/shortString.test.ts
```diff
@@ -4,19 +4,33 @@ describe('shortString', () => {
...
+  test('should throw if string to encode contains non ascii chars', () => {
     expect(() => encodeShortString('hello\uD83D\uDE00')).toThrowErrorMatchingInlineSnapshot(
       `"hello😀 is not an ASCII string"`
     );
   });
+
+  test('should convert hex number string to string', () => {
+    expect(decodeShortString('0x68656c6c6f')).toMatchInlineSnapshot(`"hello"`);
+  });
+  test('should convert decimal number string to string', () => {
```
ivpavici, line 23: nitpick: in general tests in other suites have new lines between them, so maybe to keep the same convention :)

ivpavici (approved):

PhilippeR26:
I think I have all implemented.

penovicp (approved):

PR #509, opened 2023-02-03
fix: remove hex prefix should work insensitive, for both 0x and 0X

tabaktoni:
## Motivation and Resolution
Fix https://github.com/0xs34n/starknet.js/pull/508
Mentioned PR contains change so that call on function decodeShortString can handle both hex and decimal string, usage-related change.
```decodeShortString(value.toString(16)) -> decodeShortString(value)```
But removeHexPrefix removes only `0x` and not `0X`. That produces a hidden character on the string that encodes, visible with JSON.stringify.
The issue is that removeHexPrefix should be able to remove both, not just in the mentioned PR but in case someone does an uppercase transformation before the call.
...

...

## Usage related changes
removeHexPrefix is case insensitive `0x` and `0X`

## Checklist:

- [x] Documented the changes
- [ ] Updated the docs (www)
- [ ] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/utils/shortString.test.ts
+++ b/__tests__/utils/shortString.test.ts
@@ -1,3 +1,4 @@
+import { removeHexPrefix } from '../../src/utils/encode';
 import { decodeShortString, encodeShortString } from '../../src/utils/shortString';
 
 describe('shortString', () => {
@@ -37,5 +38,10 @@ describe('shortString', () => {
     expect(() => decodeShortString('Test')).toThrowErrorMatchingInlineSnapshot(
       `"Test is not Hex or decimal"`
     );
+  });
+
+  test('explicitly test removeHexPrefix', () => {
+    expect(removeHexPrefix('0x01')).toBe('01');
+    expect(removeHexPrefix('0X01')).toBe('01');
   });
 });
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: __tests__/utils/shortString.test.ts [hunk 1]
```diff
@@ -20,11 +20,11 @@ describe('shortString', () => {
   });
 
   test('should convert hex number string to string', () => {
-    expect(decodeShortString('0x68656c6c6f')).toMatchInlineSnapshot(`"hello"`);
+    expect(JSON.stringify(decodeShortString('0x68656c6c6f'))).toBe('"hello"');
   });
 
   test('should convert decimal number string to string', () => {
-    expect(decodeShortString('448378203247')).toMatch(`hello`);
+    expect(JSON.stringify(decodeShortString('448378203247'))).toBe('"hello"');
```
ivpavici, line 27: Should we maybe add a new test case/cases with the example you mentioned (0X) that caused the bug?

tabaktoni: Not required as I left it as '0X' in decodeShortString and modified the test to detect hidden characters.
This test will not pass if there is an issue.

penovicp (approved):

__tests__/utils/shortString.test.ts
tabaktoni: On the other thought, I added it to have more transparent test.
ab4d341c3df6000ff4d0e14cd4d8788a8e07cb7f

PR #551, opened 2023-03-10
compiledClassHash &  classHash implementation

tabaktoni:
NEW:
computeSieraContractClassHash(siera) - classHash from siera (cairo 1 .json file)
computeCompiledClassHash(casm) - compiledClassHash from (cairo 1 .casm file)
computeContractClassHash(contract) - compute the class hash for both siera and legacy based on provided contract

CHANGE:
computeContractClassHash renamed to computeLegacyContractClassHash

- [x] pass test (get same CompiledClassHash as from cairo)
- [x] add noble with cjs when released

## Checklist:

- [x] Documented the changes
- [ ] Updated the docs (www)
- [x] Updated the tests
- [x] All tests are passing


paulmillr:
cjs is done: use https://github.com/paulmillr/micro-starknet 0.2.1; api is the same

src/utils/hash.ts
```diff
@@ -281,3 +292,139 @@ export function computeContractClassHash(contract: CompiledContract | string) {
     dataHash,
   ]);
 }
+
+// Cairo1 bellow
```
penovicp, line 296:
```suggestion
// Cairo1 below
```

package.json
```diff
@@ -92,6 +92,7 @@
     "ethereum-cryptography": "^1.0.3",
     "isomorphic-fetch": "^3.0.0",
     "json-bigint": "^1.0.0",
+    "json-keys-sort": "^2.1.0",
```
janek26, line 95:
```
const ordered = Object.keys(unordered).sort().reduce(
  (obj, key) => { 
    obj[key] = unordered[key]; 
    return obj;
  }, 
  {}
);
```

should we replace the dependency by a simple util function?

src/utils/hash.ts
```diff
@@ -212,40 +223,40 @@ function nullSkipReplacer(key: string, value: any) {
...
+      } else if (char === ',' && !insideQuotes) {
+        newString += ', ';
+      } else {
+        newString += char;
+      }
+      return [insideQuotes, newString];
+    },
+    [false, '']
+  )[1];
+}
```
janek26, line 248:
```
function formatSpaces(json: string) {
  let insideQuotes = false;
  let newString = '';

  for (const char of json) {
    if (char === '"' && newString.endsWith('\\') === false) {
      insideQuotes = !insideQuotes;
    }
    if (insideQuotes) {
      newString += char;
    } else {
      newString += (char === ':' ? ': ' : (char === ',' ? ', ' : char));
    }
  }

  return newString;
}
```

more readable maybe

janek26 (approved):

package.json
tabaktoni: I think this function will not do deep sorting, only top-level keys. I looked into this package source it has one file with validation and deep sort.

__tests__/fixtures.ts
```diff
@@ -2,14 +2,29 @@ import fs from 'fs';
...
-import { CompiledContract, waitForTransactionOptions } from '../src/types';
+import {
+  CompiledContract,
+  CompiledSiera,
+  CompiledSieraCasm,
+  waitForTransactionOptions,
+} from '../src/types';
 import { toHex } from '../src/utils/num';
 
 const readContract = (name: string): CompiledContract =>
```
penovicp, line 13:
```suggestion
const readContract = <T = CompiledContract>(name: string, extension: string = 'json'): T =>
  json.parse(
    fs.readFileSync(path.resolve(__dirname, `../__mocks__/${name}.${extension}`)).toString('ascii')
  )

...

export const compiledHelloSiera = readContract<CompiledSiera>('cairo/helloSiera/hello')
export const compiledHelloSieraCasm = readContract<CompiledSieraCasm>('cairo/helloSiera/hello', 'casm')
```

penovicp (approved):
Some minor nitpicks, lgtm

__tests__/fixtures.ts
tabaktoni: I like the generic approach (for permanent solutions) but this and the hash file will be cleaned up soon as we progress to remote compile and integrated hashes in compilation, so it is temporary and easy to remove.

Issue #52, opened 2021-11-09
Reimplement json-bigint


janek26:
One of our dependencies https://github.com/sidorares/json-bigint has seen it's best days.
It uses bignumber.js as a dependency, while this library uses bn.js. It also supports native BigInt which breaks support for older browsers and browserslists (see #37 ).

Lets reimplement this package in TS.
As a starting point we can use the test suites provided at the original repo and write an implementation against it.

We can ether make this repo a monorepo or start a new repo.

### Acceptance Criteria
- no use of native BigInt
- use BN.js as only dependency
- passes all test from the original repo

wraitii:
FYI I'm hitting [this issue](https://github.com/sidorares/json-bigint/issues/34) with json-bigint when using starknet.js .
My temp fix is to pull from `git+https://github.com/bushev/json-bigint.git#e3e93f71d633cd558bc7c57afcd9e1f54f669d0b` in my package.json

janek26:
on hold.
Maybe Starkware moves to using hex everywhere in compiled contracts which would solve this and remove the bitint json dependency

ivpavici:
Close with https://github.com/0xs34n/starknet.js/pull/560, if something is missing please reopen :)

PR #560, opened 2023-03-21
chore: swap json utility dependency

penovicp:


ivpavici:
@janek26 could this close https://github.com/0xs34n/starknet.js/issues/52 ?

PR #570, opened 2023-03-30
chore: fix call response type & Signature type

tabaktoni:
## Motivation and Resolution
Fix TS issue with contract.call response
Signature https://github.com/0xs34n/starknet.js/discussions/565

## Usage related changes
Type-defined but more flexible Signature:
If used as string[] it can be any implementation and default implementation use WeierstrassSignatureType 

## Development-related changes
No interface changes just redefined types
```ts
export type WeierstrassSignatureType = weierstrass.SignatureType;
export type ArraySignatureType = string[];
export type AnySignatureType = string | number | ArraySignatureType | WeierstrassSignatureType;
export type Signature = ArraySignatureType | WeierstrassSignatureType;
```

## Checklist:

- [ ] Documented the changes
- [ ] Updated the docs (www)
- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/types/lib/index.ts
+++ b/src/types/lib/index.ts
@@ -3,10 +3,10 @@ import { weierstrass } from '../../utils/ec';
 import type { BigNumberish } from '../../utils/num';
 import { CompiledContract, CompiledSierraCasm, ContractClass } from './contract';
 
-// Common Signature Type which needs to be imported from weierstrass
-// and imported at many places
-// This is because stark.ts doesn't export SignatureType
-export type Signature = weierstrass.SignatureType;
+export type WeierstrassSignatureType = weierstrass.SignatureType;
+export type ArraySignatureType = string[];
+export type AnySignatureType = string | number | ArraySignatureType | WeierstrassSignatureType;
+export type Signature = ArraySignatureType | WeierstrassSignatureType;
 
 export type RawCalldata = BigNumberish[];
 export type AllowArray<T> = T | T[];
--- a/src/types/lib/index.ts
+++ b/src/types/lib/index.ts
@@ -5,7 +5,6 @@ import { CompiledContract, CompiledSierraCasm, ContractClass } from './contract'
 
 export type WeierstrassSignatureType = weierstrass.SignatureType;
 export type ArraySignatureType = string[];
-export type AnySignatureType = string | number | ArraySignatureType | WeierstrassSignatureType;
 export type Signature = ArraySignatureType | WeierstrassSignatureType;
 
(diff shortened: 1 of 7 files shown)


src/utils/stark.ts
```diff
@@ -40,26 +42,32 @@ export function makeAddress(input: string): string {
...
   return bigNumberishArrayToHexadecimalStringArray(formatSignature(sig));
 }
 
-export function parseSignature(sig?: string[]) {
-  if (!sig) return undefined;
+export function parseSignature(sig?: ArraySignatureType): WeierstrassSignatureType {
+  if (!sig) throw Error('parseSignature: provided signature is undefined');
 
   const [r, s] = sig;
   return new Signature(toBigInt(r), toBigInt(s));
```
dhruvkelawala, line 73: This would not work if `sig = [r1, s1, r2, s2 ]`. I would suggest in this case to return `WeierstrassSignatureType[]` or get rid of this function completely.

src/types/lib/index.ts
```diff
@@ -3,10 +3,10 @@ import { weierstrass } from '../../utils/ec';
 import type { BigNumberish } from '../../utils/num';
 import { CompiledContract, CompiledSierraCasm, ContractClass } from './contract';
 
-// Common Signature Type which needs to be imported from weierstrass
-// and imported at many places
-// This is because stark.ts doesn't export SignatureType
-export type Signature = weierstrass.SignatureType;
+export type WeierstrassSignatureType = weierstrass.SignatureType;
+export type ArraySignatureType = string[];
+export type AnySignatureType = string | number | ArraySignatureType | WeierstrassSignatureType;
```
dhruvkelawala, line 8: I don't like this type tbh. This is too flexible. `ArraySignatureType | WeierstrassSignatureType` should be good for all cases.

tabaktoni: Agree, didn't like it either but wanted to make it smarter so users don't need to wrap it in an array. But as you suggested I will just drop it and force array for custom.

src/utils/stark.ts
tabaktoni: Yea this is not flexible at all, this is used only in parseGetTransactionResponse, but I am not sure this response will return a signature anymore, maybe if you query some old transactions.
I am in favor of dropping it and just returning raw data as it is in that case.

dhruvkelawala: That sounds okay to me

dhruvkelawala (approved):

PR #576, opened 2023-04-06
Feat/0.11.0/extend-tests

tabaktoni:
## Motivation and Resolution
Extended tests for 0.11.0 and minor type fixes

Wait for devnet fix& bump for GetClassAt test to pass

## Checklist:
- [x} Performed a self-review of the code
- [x] Documented the changes
- [ ] Updated the docs (www)
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/account.test.ts
+++ b/__tests__/account.test.ts
@@ -513,16 +513,24 @@ describeIfDevnetSequencer('not implemented for RPC', () => {
     const provider = getTestProvider() as SequencerProvider;
     const account = getTestAccount(provider);
     let classHash;
+    let contractAddress;
+    let declareV2Tx;
     initializeMatcher(expect);
 
-    test('Declare v2 - Hello Cairo 1 contract', async () => {
-      const declareTx = await account.declare({
+    beforeAll(async () => {
+      declareV2Tx = await account.declare({
         contract: compiledHelloSierra,
         casm: compiledHelloSierraCasm,
       });
-      classHash = declareTx.class_hash;
-      await provider.waitForTransaction(declareTx.transaction_hash);
-      expect(declareTx).toMatchSchemaRef('DeclareContractResponse');
+      classHash = declareV2Tx.class_hash;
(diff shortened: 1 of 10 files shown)


__tests__/account.test.ts
```diff
@@ -501,18 +509,56 @@ describe('deploy and test Wallet', () => {
...
+      const classResponse = await provider.getClassByHash(classHash);
+      expect(classResponse).toMatchSchemaRef('SierraContractClass');
+    });
+
+    test('GetClassAt', async () => {
+      const classResponse = await provider.getClassAt(contractAddress);
+      expect(classResponse).toMatchSchemaRef('SierraContractClass');
+    });
+
+    test('getCompiledClassByClassHash', async () => {
```
penovicp, line 559: Is this a duplicate of the third preceding/last test (the referenced schema as well)?

Nitpick: The `GetClassByHash` and `GetClassAt` tests could be changed to use `account` in place of `provider` so they align closer to the test file name.

src/utils/responseParser/sequencer.ts
```diff
@@ -211,4 +212,12 @@ export class SequencerAPIResponseParser extends ResponseParser {
       },
     };
   }
+
+  // TODO: Define response as new type as it diff from ContractClass
+  public parseSierraContractClassResponse(res: any): SierraContractClass {
```
penovicp, line 217: Nitpick: Not sure if it would make sense to put this somewhere else since `abi` property of the RPC response will also be a string.

__tests__/account.test.ts
tabaktoni: Yea,
I think I will extract this all to the new cairo1 test where we will update also contract interactions

src/utils/responseParser/sequencer.ts
tabaktoni: At the moment until protocol stabilized both seq. and rpc. had their own response parser.

Issue #579, opened 2023-04-10
Cairo 1 Contract Interaction


tabaktoni:
**Prerequisite**

Define the list of build-in types in Cairo 1
- [ ] felt -> felt252
- [ ] u8
- [ ] u16
- [ ] u32
- [ ] u64
- [ ] u128
- [ ] u256
- [ ] bool ?
- [ ] box ?
- [ ] nullable ?
- [ ] array (mix of types ?)
- [ ] ContractAddress
- [ ] Struct
- [ ] Tuple ??

Abi parsing 
- [ ] waiting for starknet update (abi methods missing types, no structures in abi ...)

**Implementation**

PhilippeR26:
bool and usize.
Not sure for box and option.

tabaktoni:
This look like to contain core types
https://github.com/starkware-libs/cairo/blob/3779d0db5968378061d40c30ab23932009e64b05/crates/cairo-lang-sierra/src/extensions/core.rs#L38

PR #581, opened 2023-04-11
Cairo 1 contract interaction

tabaktoni:
## Motivation and Resolution
Resolve https://github.com/0xs34n/starknet.js/issues/579
Bump devnet
Fix nested response, Update ``Result``: object, array, bool, string, bigint

Pass with validation and parsing
- [x] felt252
- [x] u8
- [x] u16
- [x] u32
- [x] u64
- [x] u128
- [x] u256
- [x] bool ?
- [x] ContractAddress (same as felt? what is diff ?)
- [x] unnamed flat tuples
- [x] array (echo, not store, simple flat)
- [ ] Struct (echo, not stored, simple flat)
- [ ] box ?
- [ ] nullable ?
- [ ] named Tuple ??
- [ ] unnamed nested tuple

Abi parsing 
- [ ] waiting for starknet update (abi methods missing types, no structures in abi ...) ?
- [ ] ABI: missing names parameter in output object only typer (intentional ?)

## Checklist:

- [ ] Documented the changes
- [ ] Updated the docs (www)
- [ ] Updated the tests
- [ ] All tests are passing


ivpavici (approved):
LGTM!

src/types/contract.ts
```diff
@@ -1,5 +1,10 @@
...
-  [key: string]: any;
-};
+export type Result =
+  | {
+      [key: string]: any;
+    }
+  | bigint
+  | string
+  | boolean
+  | any[];
```
penovicp, line 10: Would swapping `any` for `Result` make sense?

```
export type Result = Result[] | { [key: string]: Result }
```

PR #592, opened 2023-04-22
feat: make cairoVersion a property of Account

### LINKED ISSUE #591: Account @param cairoVersion
Now it is possible to declare a Contract. to be a Cairo1 contract using cairoVersion = '1': 
```js
constructor(abi: Abi, address: string, providerOrAccount?: ProviderInterface | AccountInterface, cairoVersion?: string);
```

But this is not possible for an Account:
```js
constructor(providerOrOptions: ProviderOptions | ProviderInterface, address: string, pkOrSigner: Uint8Array | string | SignerInterface);
```

It should be possible to specify it, so that the calldata are parsed using the new way of parsing data.

dhruvkelawala:
## Motivation and Resolution

Resolves #591  

## Checklist:

- [ ] Documented the changes
- [ ] Updated the docs (www)
- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/account/default.ts
+++ b/src/account/default.ts
@@ -6,6 +6,7 @@ import { Signer, SignerInterface } from '../signer';
 import {
   Abi,
   AllowArray,
+  CairoVersion,
   Call,
   DeclareAndDeployContractPayload,
   DeclareContractPayload,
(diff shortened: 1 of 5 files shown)


tabaktoni:
Problem with existing implementation is flag do no belong to InvocationsDetails (that's create some uncecesery usages and is not part of transaction)
I created PR to patch deploy when used with this flag as it will break (UDC contract is cairo 0)
Example dev try to use this expecting to work for deploying cairo 1 contract
```ts
const deploy = await account1.deployContract({
        classHash,
      }, {cairoVersion: '1'});
```
In this case with global flag, users will do
```ts
const account1 = getTestAccount(provider, '1');
      const deploy = await account1.deployContract({
        classHash,
      });
```
And this will break as deploy use execute that use now class prop and can't be overide.

I will try to figure out solution to resole both cases and do not change interface

tabaktoni:
At the moment this is the only approach and we will figure out edge cases later.

tabaktoni (approved):

PR #599, opened 2023-04-27
API doc autogenerated

PhilippeR26:
## Motivation and Resolution

API documentation is never up to date, and there are no links to the types involved in each method. Update this API doc is time consuming.
JS doc is already implemented in the code for main methods. With implementation of TypeDoc.js, the JS doc is collected and a full API doc is automatically generated in Docusaurus. In addition, the VScode of users will display in pop-ups the JS doc related to the position of their cursor.
So, update one time the JS doc --> automatic update of API doc, and update of VScode help popup.

## Usage related changes

Users will now have an up to data API doc, with links to easily find all the doc they needs.
They will have also all the up to date doc in their VScode. 

Each time the team want to update the API doc, they have to do : 
```bash
cd www
npm run build
```
It can be tested with
```bash
npm run serve
```
It can be probably automated at each release.
 
## Comments

- 2 JS doc have been updated in provider code : `{}` is prohibited in JSdoc. It has to be replaced by `\{` and `\}`
- The existing API doc has been moved in www/docs/guides/Old_API_doc, and can be consulted in the menu : `Guides/Starknet.js old API doc`
- In a second step, useful content of API old doc will be implemented in JS code.

## Checklist:

- [x] Updated the docs (www)
- [x] All tests are passing


ivpavici (approved):
LGTM, we will fix incrementally if something needs to be addressed. @penovicp we should add this also to the build pipeline

penovicp (approved):

PR #607, opened 2023-05-08
feat: declareIfNot, update declareAndDeploy with it, cleanup cairo1 test

tabaktoni:
## Motivation and Resolution
For Testing Cairo1 contracts we need to redeclare without failing.
This can be useful to 3rd parties also when dealing with cairo1 declaration.
Added as a new method to Account without extending the interface

## Usage related changes
account.declareIfNot will not fail when a contract is already declared but rather return a class hash with an empty string for transaction_hash
declareAndDeploy internal use declareIfNot so it will also not fail if the contract has already been declared

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


src/account/default.ts
```diff
@@ -327,6 +327,30 @@ export class Account extends Provider implements AccountInterface {
...
 
+  /**
+   * First check if contract is already declared, if not declare it
+   * If contract already declared returned transaction_hash is ''.
+   * Method will pass even if contract is already declared
+   * @param payload DeclareContractPayload
+   * @param transactionsDetail (optional) InvocationsDetails = {}
+   * @returns DeclareContractResponse
+   */
+  public async declareIfNot(
```
ivpavici, line 338: should we add tests for this method itself?

ivpavici (changes requested):

src/account/default.ts
tabaktoni: a18b26d8fa45cae4e6d3516f794916ca1e8f50e1

PR #609, opened 2023-05-08
handle arrays inside tuples and structs

cfal:
## Motivation and Resolution

- tuples that contain arrays eg `(core::array::Array<felt252>, bool)` are detected as arrays instead of tuples, because isTypeArray checks that the type includes the cairo array type, and because array detect happens before tuple detection here:

  https://github.com/0xs34n/starknet.js/blob/993fce6a18200c4e8126ed5d940743b1583cc363/src/utils/calldata/responseParser.ts#L94

  the first commit fixes this by checking the type prefix instead.  

- arrays inside tuples (same example as above) and structs aren't detected because they are handled in the `requestParser` switch statement instead of during `parseResponseStruct` recursion. i haven't tested with cairo 0.x, but this seems to also affect older contracts? eg `(felt*, bool)`.

  the second commit fixes this by moving the array handling inside `parseResponseStruct`.

...

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


tabaktoni:
Cairo 0 contract wasn't allowed pointers inside a struct, as such arrays with arrays or structs with array wasn't possible in response. I think I manage to cover all cases for cairo0 in contract.test.ts under Complex interactions but if there is another one I would like to extend the test for it also.

Same as another PR if you can add tests for this case, or message me peace of Cairo1 code that produces this req/res so I can extend tests with it.

Thanks

PR #610, opened 2023-05-08
utils/calldata/cairo.ts: find outermost bracket in getArrayType

cfal:
## Motivation and Resolution

arrays with more than one set of `<>` brackets would not have its inner type extracted properly since we look for the first instance of `>`.

...

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


tabaktoni:
Thanks for the contribution.
Can you provide a test for this case, or provide peace of cairo1 code that produces such a response, so I can extend the test with it? Thanks.

tabaktoni:
Fixed 2d arrays request and response, extended tests

PR #614, opened 2023-05-15
ensure valida c1 execute calls props order

tabaktoni:


src/utils/transaction.ts
```diff
@@ -59,15 +59,21 @@ export const transformCallsToMulticallArrays_cairo1 = (calls: Call[]) => {
...
-// TT: Can be removed ?
 /**
  * Transforms a list of calls in the full flattened calldata expected
  * by the __execute__ protocol.
  * @param calls
  * @returns Calldata
  */
 export const fromCallsToExecuteCalldata_cairo1 = (calls: Call[]) => {
-  return CallData.compile({ calls });
+  // ensure property order
```
penovicp, line 69: I'm fine with this code, just wanted to note that the property order being preserved is an incidental side-effect of the object implementation in the most widely used js engines, so while the behaviour probably won't change there is chance that it might.

penovicp (approved):

PR #627, opened 2023-05-30
refactor: group type exports

penovicp:
Aggregates exported types and interfaces into the `types` folder.

Moved types/interfaces with new default locations and legacy re-exports:
- `BigNumberish` from `utils/num` to `types/lib`
- `BlockIdentifier` from `provider/utils` to `types/lib`
- `Uint256` from `utils/uint256` to `types/lib`
- the `utils/typedData/types.ts` file renamed and moved to `types/`

Moved types/interfaces without location changes in the published module bundle:
- `ProviderOptions` from `provider/default` to `types/provider/configuration`
- `RpcProviderOptions` from `provider/rpc` to `types/provider/configuration`
- `SequencerHttpMethod` from `provider/sequncer` to `types/provider/configuration`
- `SequencerProviderOptions` from `provider/sequncer` to `types/provider/configuration`


tabaktoni (commented):
We need a list of changed imports for dev/user.

src/index.ts
```diff
@@ -1,12 +1,14 @@
...
  */
+export * from './account';
 export * from './contract';
-export * from './types';
 export * from './provider';
-export * from './account';
 export * from './signer';
 
+export * from './types';
+export * as types from './types';
```
tabaktoni, line 10: I guess the idea here is to have all types under type. And ``export * from './types';`` is for back compatibility reasons.
My suggestion is to mark it 'to be removed' and open an issue for future version X when it will be removed, so we don't forget it.

tabaktoni (approved):
Like this cleanup, LGTM.

- Just add notes on import changes in PR.
- As far as i can see no breaking changes are inroduced for the end user, only deprecated ones. (add note for new import)
- It will be nice to merge this one and RPC in the same new release.

PR #654, opened 2023-06-20
feat: isCairo1?

PhilippeR26:
## Motivation and Resolution

The way to construct the calldata, and the way to read the answer of a contract function, are not the same for Cairo 0 and Cairo 1.
That's why it's useful to have a method to have the origin of a Contract object, or of an abi variable.

## Usage related changes

Users have now : 

```typescript
import { cairo } from "starknet";
const isCairo1: boolean = myContract.isCairo1();
const isAbiCairo1: boolean = cairo.isCairo1abi(myAbi);
```

function `isCairo1abi` added in `utils/calldata/cairo/`.

## Checklist:

- [X] Documented the changes in code (API docs will be generated automatically)
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/contract.test.ts
+++ b/__tests__/contract.test.ts
@@ -1,6 +1,6 @@
 import { BigNumberish, Contract, ContractFactory, RawArgs, json, stark } from '../src';
 import { CallData } from '../src/utils/calldata';
-import { felt, tuple, uint256 } from '../src/utils/calldata/cairo';
+import { felt, isCairo1Abi, tuple, uint256 } from '../src/utils/calldata/cairo';
 import { getSelectorFromName } from '../src/utils/hash';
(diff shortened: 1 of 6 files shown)


penovicp (approved):

__tests__/contract.test.ts
```diff
@@ -50,6 +50,13 @@ describe('contract module', () => {
         );
       });
 
+      test('isCairo1', async () => {
```
ivpavici, line 53: this is a test for a negative case, can we add a additional positive test case?

ivpavici (approved):

tabaktoni:
UserNote: This could potentially be helpful to a dev who creates an app that works with both cairo0 and cairo1 contracts at the same time.
DevNote: This method should not be used internaly

PR #704, opened 2023-08-02
Feat/parse events

### LINKED ISSUE #697: Parsing of events
Not related to a problem per se but it would facilitate the usage of events in Starknet js

It's a hassle to read events coming from a contract for which we have the ABI available. I think it would be very nice to have a ParseEvent function available in the Contract class

For now I believe the only way of reading events is by parsing it manually: https://www.starknetjs.com/docs/guides/events

edisontim:
## Motivation and Resolution
Event parsing could be done since we have the event definitions in the abi and this helps a great deal when working with events.
This resolves #697

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


src/utils/calldata/responseParser.ts
```diff
@@ -18,7 +18,7 @@ import extractTupleMemberTypes from './tuple';
  * @param it iterator
  * @returns bigint | boolean
  */
-function parseBaseTypes(type: string, it: Iterator<string>) {
+export function parseBaseTypes(type: string, it: Iterator<string>) {
```
tabaktoni, line 21: I think we don't need to export it?

src/utils/calldata/responseParser.ts
```diff
@@ -42,7 +42,7 @@ function parseBaseTypes(type: string, it: Iterator<string>) {
  * @param structs - structs from abi
  * @return {any} - parsed arguments in format that contract is expecting
  */
-function parseResponseValue(
+export function parseResponseValue(
```
tabaktoni, line 45: Same as above

tabaktoni (changes requested):

tabaktoni (approved):

tabaktoni (approved):

Issue #759, opened 2023-09-15
provider method for iscairo1


PhilippeR26:
**Is your feature request related to a problem? Please describe.**
We have now Cairo 0 and Cairo 1 accounts. A DAPP can collect an address of account, without the information of Cairo version.
It is possible to find the Cairo version, but it needs several lines of code.

**Describe the solution you'd like**
To propose  a dedicated method :
```typescript
const isV1:boolean = provider.isContractV1(contractAddress);
```` 

**Describe alternatives you've considered**
Stay as it is, with no improvement of UX.

**Additional context**
It will be possible to remove this method after Regenesis of Starknet.
I would like to propose a small PR on this subject.

**Comment by tabaktoni** [comment]
I had thoughts about 'auto' ``CairoVersion`` but it can't be implemented backward compatible, and side effect free.
Dev needs to provide ``CairoVersion`` on the constructor and the constructor does not allow side effects (read async).
There is another way by optional chaining initialization like:
```ts
let acc = new Account(provider, address, key)
await acc.setVersion();
```
where setVetsion(version) if not provided version will do "auto detect".
Regarding "auto detect" method implementation, I will put it into utils and not provider.
Also calling it
```ts
// get contract Cairo version, fullname getContractCairoVersion
getContractCairoVersion(provider, address): VersionMetadata

type VersionMetadata {
 compilerVersion:  '0' | '1' | '2' | ....
 cairoVersion: '0' | '1'
 // potentially some more metadata
}
```
It will return Cairo version for any contract deployed on address.

This method can then be used inside Account setVersion.

I did not proceed with this implementation as I didn't see real value in it as it will not auto fix user issue with Account versions, as that will still need to read about it in documentation same as that need to do with Account CairoVersion param.
Regarding getContractCairoVersion method there were some discussions regarding providing compiler version in deployed artifacts so that it can be read out, but i don't know is it done and can it be done some other way in the maintime

PhilippeR26:
OK.
Will be detailed when rpc 0.5.0 will have to be implemented.

tabaktoni:
So we decided on by-weekly that this will be done so that async operation is called on usage and not on constructor.
So we can do as described just define CairoVersion on Execute when needed.

PhilippeR26:
Do I take the opportunity to change default `CairoVersion` to "1"?

PR #768, opened 2023-09-22
automatic evaluation of cairoVersion in Account

PhilippeR26:
## Motivation and Resolution
Solve account discussion from #759 
Account object is able to define itself the cairoVersion variable.

## Usage related changes

- At Account instance creation, if the account exists in the network, the cairoVersion parameter is not necessary. It will be automatically defined.
- If the account is not existing in the network, and if the cairoVersion parameter is not used, the default version is applied. 
- If the cairoVersion parameter is used, its value is used, whatever the status in the Network.
- version by default remains with "0".

If the cairoVersion isn't defined by the user at the creation of the Account instance, it will be defined once, just before it is necessary in an Account method (estimateFee, execute, ...)

## Checklist:

- [x] Documented the changes in code (API docs will be generated automatically) : updated guides
- [x] Updated the tests
- [x] All tests are passing


PhilippeR26:
Do I take the opportunity to change default CairoVersion to "1"?

PhilippeR26:
Ready for review.
default CairoVersion remains "0".

PR #792, opened 2023-10-17
[V6] WalletAccount

tabaktoni:
## Motivation and Resolution

https://community.starknet.io/t/snip-standardizing-starknet-wallet-connections-with-write-api-and-json-rpc-integration/98049#using-rpc-for-starknet-wallet-connections-1-6

- [x] invoke
- [x] declare
- [x] deploy
- [x] events

Declare had some error on extension side to discuss.

...

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


tabaktoni:
> @tabaktoni Just a feedback on the draft WalletAccount PR. Even though having class based approach for other starknet.js entities makes sense, for WalletAccount in particular I think we should go with functional approach. WalletAccount or WalletClient should just be an “interface” providing an high level interfacing between get-starknet and wallets, I like viem’s approach to this.

The initial idea is a drop-in replacement for Account Class. But we can eventually do both

src/wallet/account.ts
```diff
@@ -0,0 +1,188 @@
...
+    const declareContractPayload = extractContractHashes(payload);
+
+    // DISCUSS: HOTFIX: Adapt Abi format
+    const pContract = payload.contract as CompiledSierra;
+    const cairo1Contract = {
+      ...pContract,
+      abi: stringify(pContract.abi),
+    };
+
+    // Check FIx
```
ivpavici, line 146: is this checked or still a TODO?

src/wallet/account.ts
```diff
@@ -0,0 +1,188 @@
...
+    };
+
+    return addDeployAccountTransaction(this.walletProvider, params);
+  }
+
+  override signMessage(typedData: TypedData) {
+    return signMessage(this.walletProvider, typedData);
+  }
+
+  // MISSING ESTIMATES
```
ivpavici, line 187: maybe add a TODO: for easier searching?

src/wallet/connect.ts
```diff
@@ -0,0 +1,162 @@
...
+  AddStarknetChainParameters,
+  NetworkChangeEventHandler,
+  StarknetChainId,
+  StarknetWindowObject,
+  TypedData,
+  WatchAssetParameters,
+} from './getst/main';
+
+/**
+ * Request Permission for wallet account, return addresses that's allowed by user
```
ivpavici, line 15:
```suggestion
 * Request Permission for wallet account, return addresses that are allowed by user
```

src/wallet/connect.ts
```diff
@@ -0,0 +1,162 @@
...
+  });
+}
+
+/**
+ * Request adding custom Starknet chain
+ * @param chain AddStarknetChainParameters
+ * @returns boolean
+ */
+export function addStarknetChain(swo: StarknetWindowObject, chain: AddStarknetChainParameters) {
+  // Can this set custom RPC endpoint ?
```
ivpavici, line 54: also maybe TODO: here if still relevant

src/wallet/getst/localStorageStore.d.ts
```diff
@@ -0,0 +1,14 @@
+export interface IStorageWrapper {
+  set(value: string | null | undefined): boolean;
+  get(): string | null | undefined;
+  delete(): boolean;
+}
+export declare class LocalStorageWrapper implements IStorageWrapper {
+  #private;
```
ivpavici, line 7: is this a leftover?

src/wallet/types.ts
```diff
@@ -0,0 +1,8 @@
+import { RpcMessage, StarknetWindowObject } from './getst/main';
+
+// ---- TT Request Handler
+export type RpcCall = Omit<RpcMessage, 'result'>;
+
+// -- TT Better naming
```
ivpavici, line 6: maybe a TODO: here ?

src/wallet/account.ts
tabaktoni: Yes it is diff between declare and api-wallet declare, the one Dhruv mentioned on call before Denver.
I would leave it for now as reminder

src/wallet/connect.ts
tabaktoni: Yeah this one didn't work on the custom wallet build

src/wallet/getst/localStorageStore.d.ts
tabaktoni: It is a prop from get-starknet but now mentioning it, it looks like I need to update that one.

ivpavici (approved):

PR #794, opened 2023-10-25
Update deployFee estimation example

petscheit:
Before it seemed like the calldata cost would be derived via contact ABI automatically, this makes it more clear that the constructor calldata must be explicitly passed


PhilippeR26:
Hello,
In fact, the `constructorCalldata` is not mandatory :
image
The current doc is valid for a contract without constructor, and is not showing that it's necessary in case of constructor.
Your proposal do not show that it's not necessary in case of contract without constructor.

Could you add a comment? Like :

> `constructorCalldata` is not necessary if the contract to deploy as no constructor.

petscheit:
Hi Phillipe,
I was chasing an issue where the gas estimation for deployment was way too low and it took me a moment to figure out what is causing this. My assumption was that the estimation would fetch the constructor arguments from the contracts ABI, and adds this to the estimate. Having this in the example would have saved me a bit of investigating. 

I also see you point though, maybe it can be made a clearer in another way in the docs.

PhilippeR26:
Just insert in your PR this comment : 

> `constructorCalldata` is not necessary if the contract to deploy has no constructor.

tabaktoni (changes requested):
@PhilippeR26 suggestion

PhilippeR26:
> @PhilippeR26 suggestion

Yes, as I wrote earlier, I would like to add just after the code : 

> `constructorCalldata` is not necessary if the contract to deploy has no constructor.

PR #798, opened 2023-10-30
RPC v5, Test Setup Update

tabaktoni:
### Refactor & Implement RPC 0.5
fix https://github.com/starknet-io/starknet.js/issues/793
- [x] Rebuild RPC spec from zero
- [x] Refactor RPC spec into submodules
- [x] Added new methods
- [x] Checked all methods specified
- [x] Rpc Provider removed response types definition, used inheritance
- [ ] ~~Refactor waitForTransaction with getTransactionStatus~~ (Will wait RPC 0.5 to become standard)
- [ ] ~~Remove waitForTransaction Pathfinder fix~~
- [x] Remove buildTransaction type fix
- [x] Add tests for new methods
- [x] Check response parsers on all RPC Methods
- [ ] ~~Check all methods responses to conform to same type naming for less breaking types~~ (I think plane response show better typing)
- [x] Check exports for RPC module
- [x] RPC Provider Cleanup (renames, reorder, ...) and method props update
- [x] Refactor Jest Test Setup

### Refactor Jest Test Setup
fix https://github.com/starknet-io/starknet.js/issues/789

- [x] Introduce new discovery strategy, more at imp ([__tests__/jestGlobalSetup.ts](https://github.com/starknet-io/starknet.js/pull/798/files#diff-e0621975ab8c7e88db8475932de4a31552fb36b05310a2fb2c5bee5bf97ca7c3))
- [x] Test for remote devnet
- [x] Introduced better final env setup
- [x] Cleanup fixtures
- [x] Implement Final ENV Override by the user (advanced testing setups)

0.5

## Usage related changes
Potential RPC Provider Types

- jestGlobalSetup
- openrpc.ts -> .../rpcspec/* (components, methods, errors, nonspec -> for extending specification), rpc.ts -> for overriding specification
- TS Methods and Protocol Check (.../jsonrpc)
- RPC ``fetch()`` extended with jsonrpc id
- Updated RPC **errorHandler**
- RPC Methods removed forced

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


__tests__/jestGlobalSetup.ts
```diff
@@ -0,0 +1,255 @@
...
+/**
+ * Global Setup Fixtures
+ */
+
+/* Default test config based on run `starknet-devnet --seed 0` */
+const GS_DEFAULT_TEST_PROVIDER_URL = 'http://127.0.0.1:5050/';
+
+const overrideOrSet = (envName: string, setValue: string) => {
+  process.env[envName] = process.env[envName] ?? setValue;
+};
```
penovicp, line 28: I would suggest incorporating the boolean string ternary assignments made in the calls to `overrideOrSet` within the method itself so the calls can be made less verbose. I would also suggest renaming the method without the `override` fragment. E.g.:


```suggestion
const asBooleanString = (b: boolean) => (b ? 'true' : 'false');
const setIfNullish = (envName: string, setValue: string | boolean | undefined) => {
  process.env[envName] ??= typeof setValue === 'boolean' ? asBooleanString(setValue) : setValue;
};
```

__tests__/jestGlobalSetup.ts
```diff
@@ -0,0 +1,255 @@
...
+    const response = await fetch(`${GS_DEFAULT_TEST_PROVIDER_URL}is_alive`);
+    const body = await response.text();
+    if (body === 'Alive!!!') {
+      strategy.isDevnet = true;
+    } else {
+      return setup(strategy);
+    }
+  } catch (error) {
+    return setup(strategy);
+  }
```
penovicp, line 59: An approach I sometimes use to condense simple `try/catch` blocks is to use a promise chain, it might be worth considering here. Same thing could also be done for the RPC request.

```suggestion
  const devnetResult = await fetch(`${GS_DEFAULT_TEST_PROVIDER_URL}is_alive`)
    .then((response) => response.text())
    .catch(() => '');
  if (devnetResult !== 'Alive!!!') {
    return setup(strategy);
  }
  strategy.isDevnet = true;
```

__tests__/jestGlobalSetup.ts
```diff
@@ -0,0 +1,255 @@
...
+  } else if (providerType.rpc) {
+    console.log('Detected RPC');
+  }
+
+  const isAccountSet = await setAccount(devnetStrategy);
+  if (isAccountSet) {
+    console.log('Detected Account');
+  }
+
+  ready = await verifySetup(true);
```
penovicp, line 242: `verifySetup` is sync so could be:
```suggestion
  ready = verifySetup(true);
```
or:
```suggestion
  return verifySetup(true);
```

penovicp (commented):
Made some minor code style suggestions for the test setup.

In general the PR lgtm

__tests__/jestGlobalSetup.ts
tabaktoni: Good sugestion, I updated it using nullish assignment with toString handling bools.

__tests__/jestGlobalSetup.ts
tabaktoni: Nice one

__tests__/rpcProvider.test.ts
```diff
@@ -210,9 +237,15 @@ describeIfRpc('RPCProvider', () => {
 
       test('getTransactionByHash', async () => {
         const transaction = await rpcProvider.getTransactionByHash(transaction_hash);
+
+        // const b: types.RPC. = 2;
```
ivpavici, line 241: leftover?

src/types/api/jsonrpc/index.ts
```diff
@@ -0,0 +1,29 @@
+// TODO: Provjerit ovo i 'mergat' sa methods
```
ivpavici, line 1: leftover? or comment in english if not :D

__tests__/rpcProvider.test.ts
tabaktoni: Yep

src/types/api/jsonrpc/index.ts
tabaktoni: leftover

penovicp (commented):
Accumulated a bunch of minor suggestions which I felt became impractical so I opened PR #808. It has two commits, one contains minor tsdoc suggestions relating to the auto-generated documentation appearance. The other commit resolves an issue that arises in the build result.

penovicp (approved):

PR #804, opened 2023-11-02
refactor: parse int as number or big int

NguyenTuanCanh:

1.Use Number.isInteger to check if the input can be parsed as an integer, and if not, we parse it as a float.
2.Use Number(x) to convert the string to a number for checking integer type.
3.Split the condition and result into separate lines for improved readability.
4.Use Number.isSafeInteger to check if the parsed integer is within the safe integer range.
4.If it's a safe integer, we return it as a regular number; otherwise, we return it as a BigInt.

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/utils/json.ts
+++ b/src/utils/json.ts
@@ -6,9 +6,17 @@ import * as json from 'lossless-json';
  * Convert string to number or bigint based on size
  */
 const parseIntAsNumberOrBigInt = (x: string) => {
-  if (!json.isInteger(x)) return parseFloat(x);
+  if (!Number.isInteger(Number(x))) {
+    return parseFloat(x);
+  }
+
   const v = parseInt(x, 10);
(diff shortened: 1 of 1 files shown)


tabaktoni (changes requested):
Thank you for a nice try.

- I do not see significant gain between regex and number solution, other than using the lib method, what can be removed by doing, or creating local ```isInteger(str: String)```

```ts
/^-?[0-9]+$/.test(x)
```
https://www.measurethat.net/Benchmarks/Show/28243/0/compare-number-vs-regex-test

- Regarding readability more compact method is preferred as more convenient to read.

Please provide a convenient argument for the change
If you would like to contribute please take some of the open Issues.

penovicp:
Also didn't notice a performance change with a benchmark that uses the different approaches within the lossless-json parser override method.

PR #808, opened 2023-11-06
chore: remove TS namespaces to resolve build issue

penovicp:


src/provider/rpc.ts
```diff
@@ -648,7 +640,7 @@ export class RpcProvider implements ProviderInterface {
 
   /**
    * Returns all events matching the given filter
-   * @returns RPC.Events - events and the pagination of the events
```
tabaktoni, line 651: Here i left type because for some reason response type shown is Promise<RPC.SPEC.EVENTS_CHUNK> instead of Promise<RPC.Events>

penovicp: The notation for the return type should use braces e.g. `@return {RPC.Events}`, however it would also be ignored. Not sure if there is a way to force TS to use the `RPC.Events` alias and not the base type it unspools to.

tabaktoni: Yea not sure, we can go back to this question in calmer times.

tabaktoni (approved):

PR #809, opened 2023-11-06
Deprecations & Default RPC Provider

tabaktoni:
## Motivation and Resolution
fix https://github.com/starknet-io/starknet.js/issues/803
- [x] Set default provider to RPC Provider
- [x] Update RPC Provider with option 'empty nodeUrl' - will use random public goerli nodeUrl from the known nodes
- [x] Update RPC Provider with option 'NetworkName for nodeUrl' - provide NetworkName and random public nodeUrl will be used for specific network
- [x] Updated test setup for default RPC Provider and new options

0.5

## Usage related changes
mentioned

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/__tests__/fixtures.ts
+++ b/__tests__/fixtures.ts
@@ -1,7 +1,7 @@
 import fs from 'node:fs';
 import path from 'node:path';
 
-import { Account, ProviderInterface, RpcProvider, SequencerProvider, json } from '../src';
+import { Account, ProviderInterface, RpcProvider, json } from '../src';
 import {
   CompiledSierra,
   CompiledSierraCasm,
@@ -47,10 +47,7 @@ export const compiledC210 = readContractSierra('cairo/cairo210/cairo210.sierra')
 export const compiledC210Casm = readContractSierraCasm('cairo/cairo210/cairo210');
 
 export const getTestProvider = (): ProviderInterface => {
-  const provider = process.env.TEST_RPC_URL
-    ? new RpcProvider({ nodeUrl: process.env.TEST_RPC_URL })
-    : new SequencerProvider({ baseUrl: process.env.TEST_PROVIDER_BASE_URL || '' });
-
+  const provider = new RpcProvider({ nodeUrl: process.env.TEST_RPC_URL });
   if (process.env.IS_LOCALHOST_DEVNET) {
     // accelerate the tests when running locally
     const originalWaitForTransaction = provider.waitForTransaction.bind(provider);
--- a/__tests__/fixtures.ts
+++ b/__tests__/fixtures.ts
@@ -1,7 +1,7 @@
 import fs from 'node:fs';
 import path from 'node:path';
 
-import { Account, ProviderInterface, RpcProvider, json } from '../src';
+import { Account, ProviderInterface, RpcProvider, SequencerProvider, json } from '../src';
 import {
   CompiledSierra,
   CompiledSierraCasm,
@@ -47,7 +47,10 @@ export const compiledC210 = readContractSierra('cairo/cairo210/cairo210.sierra')
 export const compiledC210Casm = readContractSierraCasm('cairo/cairo210/cairo210');
 
 export const getTestProvider = (): ProviderInterface => {
-  const provider = new RpcProvider({ nodeUrl: process.env.TEST_RPC_URL });
+  const provider = process.env.TEST_RPC_URL
+    ? new RpcProvider({ nodeUrl: process.env.TEST_RPC_URL })
+    : new SequencerProvider({ baseUrl: process.env.TEST_PROVIDER_BASE_URL || '' });
+
   if (process.env.IS_LOCALHOST_DEVNET) {
     // accelerate the tests when running locally
     const originalWaitForTransaction = provider.waitForTransaction.bind(provider);
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -48,11 +48,11 @@ export const UDC = {
 export const RPC_GOERLI_NODES = [
   'https://starknet-testnet.public.blastapi.io/rpc/v0.5',
   'https://rpc.starknet-testnet.lava.build/rpc/v0.5',
-  // 'http://limited-rpc.nethermind.io/goerli-juno/v0_5',
+  'https://limited-rpc.nethermind.io/goerli-juno/v0_5',
 ];
 
 export const RPC_MAINNET_NODES = [
   'https://starknet-mainnet.public.blastapi.io',
   'https://rpc.starknet.lava.build',
-  // 'http://limited-rpc.nethermind.io/mainnet-juno/v0_5',
+  'https://limited-rpc.nethermind.io/mainnet-juno/v0_5',
 ];
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -47,12 +47,10 @@ export const UDC = {
 
 export const RPC_GOERLI_NODES = [
   'https://starknet-testnet.public.blastapi.io/rpc/v0.5',
-  'https://rpc.starknet-testnet.lava.build/rpc/v0.5',
   'https://limited-rpc.nethermind.io/goerli-juno/v0_5',
 ];
 
 export const RPC_MAINNET_NODES = [
   'https://starknet-mainnet.public.blastapi.io/rpc/v0.5',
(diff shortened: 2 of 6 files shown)


src/provider/default.ts
```diff
@@ -44,21 +42,15 @@ export class Provider implements ProviderInterface {
...
     } else if (providerOrOptions && 'rpc' in providerOrOptions) {
       // providerOrOptions is rpc option
       this.provider = new RpcProvider(<RpcProviderOptions>providerOrOptions.rpc);
-    } else if (providerOrOptions && 'sequencer' in providerOrOptions) {
-      // providerOrOptions is sequencer option
-      this.provider = new SequencerProvider(<SequencerProviderOptions>providerOrOptions.sequencer);
     } else {
       // providerOrOptions is none, create SequencerProvider as default
-      this.provider = new SequencerProvider();
+      this.provider = new RpcProvider();
```
penovicp, line 53: Maybe change `ProviderOptions` to something like `type ProviderOptions = { rpc: RpcProviderOptions } | RpcProviderOptions;`, then the last `else if` can be removed and the `else` can have something like `new RpcProvider(providerOrOptions.rpc ?? providerOrOptions)`.

src/provider/default.ts
```diff
@@ -44,21 +42,15 @@ export class Provider implements ProviderInterface {
     if (providerOrOptions instanceof Provider) {
       // providerOrOptions is Provider
       this.provider = providerOrOptions.provider;
-    } else if (
-      providerOrOptions instanceof RpcProvider ||
-      providerOrOptions instanceof SequencerProvider
-    ) {
+    } else if (providerOrOptions instanceof RpcProvider) {
       // providerOrOptions is SequencerProvider or RpcProvider
```
penovicp, line 46:
```suggestion
      // providerOrOptions is RpcProvider
```

src/provider/default.ts
```diff
@@ -44,21 +42,15 @@ export class Provider implements ProviderInterface {
...
       this.provider = <ProviderInterface>providerOrOptions;
     } else if (providerOrOptions && 'rpc' in providerOrOptions) {
       // providerOrOptions is rpc option
       this.provider = new RpcProvider(<RpcProviderOptions>providerOrOptions.rpc);
-    } else if (providerOrOptions && 'sequencer' in providerOrOptions) {
-      // providerOrOptions is sequencer option
-      this.provider = new SequencerProvider(<SequencerProviderOptions>providerOrOptions.sequencer);
     } else {
       // providerOrOptions is none, create SequencerProvider as default
```
penovicp, line 52:
```suggestion
      // providerOrOptions is none, create RpcProvider as default
```

src/provider/rpc.ts
```diff
@@ -40,6 +43,17 @@ import { ProviderInterface } from './interface';
...
 
+export const getDefaultNodeUrl = (networkName?: NetworkName): string => {
+  // eslint-disable-next-line no-console
+  console.warn('Using default public node url, please provide nodeUrl in provider options!');
+  if (networkName && NetworkName.SN_MAIN === networkName) {
+    const randIdx = Math.floor(Math.random() * RPC_MAINNET_NODES.length);
+    return RPC_MAINNET_NODES[randIdx];
+  }
+  const randIdx = Math.floor(Math.random() * RPC_GOERLI_NODES.length);
+  return RPC_GOERLI_NODES[randIdx];
```
penovicp, line 54:
```suggestion
  const nodes = networkName === NetworkName.SN_MAIN ? RPC_MAINNET_NODES : RPC_GOERLI_NODES;
  const randIdx = Math.floor(Math.random() * nodes.length);
  return nodes[randIdx];
```

src/provider/rpc.ts
```diff
@@ -59,9 +73,18 @@ export class RpcProvider implements ProviderInterface {
 
   private chainId?: StarknetChainId;
 
-  constructor(optionsOrProvider: RpcProviderOptions) {
-    const { nodeUrl, retries, headers, blockIdentifier, chainId } = optionsOrProvider;
-    this.nodeUrl = nodeUrl;
+  constructor(optionsOrProvider?: RpcProviderOptions) {
+    const { nodeUrl, retries, headers, blockIdentifier, chainId } = optionsOrProvider || {};
+    if ((<any>Object).values(NetworkName).includes(nodeUrl)) {
```
penovicp, line 78: Could also do:
```suggestion
    if (Object.values(NetworkName as any).includes(nodeUrl)) {
```
or 
```suggestion
    if (Object.values(NetworkName).includes(nodeUrl as NetworkName)) {
```

penovicp (commented):
Guessing there are more test changes planned for this PR? Or will that come in a separate one?

src/constants.ts
```diff
@@ -44,3 +44,15 @@ export const UDC = {
   ADDRESS: '0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf',
   ENTRYPOINT: 'deployContract',
 };
+
+export const RPC_GOERLI_NODES = [
```
avimak, line 48: Maybe we shouldn't include the `/rpc/v0.5` part in the string and have only the base URL there. Then, in `"RpcProviderOptions"`, we could allow clients to specify the expected RPC version. This would make our API future-proof, assuming these endpoints will also be relevant for v0.6 (and beyond?)

avimak: We should probably request Juno to maintain path compatibility with other nodes, for example, `/rpc/v0.X` instead of `/v0_X`. Then snjs could simply append the expected version (or use the default if no version is specified) based on user-provided options.

tabaktoni: Regarding /rpc/v0.X it was my initial idea also, but it looks like it will not be possible to sync all services.
But this pulls also another question, at the moment with the latest version we support only the latest RPC Spec. To change that we would need to:
Case1 -  Maintain in parallel multiple versions each one related to specific RPC.
Case 2 - Support Two RPC Versions as the same codebase until the next one comes, then remove first and add the latest.

Case 2 is what you are proposing and this could be done when we:
1. Remove Sequencer Provider (it is now only deprecated)
2. Remove the common interface Between RPC and Sequencer (my plan was to first deprecate it, then let it exist for some time after the removal of Sequencer provider for the migration period, then remove it completely)
3. Move the default provider from using a common interface to using the RPC Provider interface
4. And then we have space to introduce two RPC Provider Default implementations one for 0.N-1 and another for 0.N RPC Specification.

src/provider/rpc.ts
```diff
@@ -59,9 +73,18 @@ export class RpcProvider implements ProviderInterface {
...
+      // Network name provided for nodeUrl
+      this.nodeUrl = getDefaultNodeUrl(nodeUrl as NetworkName);
+    } else if (nodeUrl) {
+      // NodeUrl provided
+      this.nodeUrl = nodeUrl;
+    } else {
+      // non provided fallback to default testnet
```
ivpavici, line 85:
```suggestion
      // none provided, fallback to default testnet
```

src/provider/default.ts
tabaktoni: In v6

__tests__/fixtures.ts
```diff
@@ -47,10 +47,7 @@ export const compiledC210 = readContractSierra('cairo/cairo210/cairo210.sierra')
 export const compiledC210Casm = readContractSierraCasm('cairo/cairo210/cairo210');
 
 export const getTestProvider = (): ProviderInterface => {
-  const provider = process.env.TEST_RPC_URL
-    ? new RpcProvider({ nodeUrl: process.env.TEST_RPC_URL })
-    : new SequencerProvider({ baseUrl: process.env.TEST_PROVIDER_BASE_URL || '' });
-
+  const provider = new RpcProvider({ nodeUrl: process.env.TEST_RPC_URL });
```
penovicp, line 50: Might be worth considering moving this change to a future PR so the Sequencer tests can pass (one last time
:crossed_fingers:).

penovicp (approved):
One small side note, lgtm

__tests__/fixtures.ts
tabaktoni: True

tabaktoni: Even doe we can do additional fixes and tests on the final v5 when devnet release

Issue #589, opened 2023-04-17
Detect and write missing unit tests


ivpavici:
Detect any missing unit tests to improve testing coverage!

Just one example is the `public async getEvents(eventFilter: RPC.EventFilter)` method

tabaktoni:
Unit Tests for CallData.compile

ivpavici:
Missing:
`getInvokeEstimateFee`

financegeek-org:
I'll take the test for getInvokeEstimateFee

tabaktoni:
@ivpavici Can we close this generic Issue as we track by code coverage so we can have specific ones?

PR #834, opened 2023-11-14
test: add unit test for util/event.ts

aliening:
## Motivation and Resolution

add unit test for util/event 

none

## Usage related changes

none

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


tabaktoni (changes requested):
This is a good test to have, but it should read actual events from the network not the mocked one to be considered as a valid test. 
This parsing will fail only if the network changes the event structure, and this test will not catch such event.

Bitcex:
this is good

penovicp:
Ask and await approval in #589 before contributing unit tests.

PR #836, opened 2023-11-15
fix: errors when testing in Testnet with a Cairo 1 account

PhilippeR26:
## Motivation and Resolution
As no Devnet is currently supporting rpc0.5.0, I tried to use a local Pathfinder node (Testnet) with a Cairo1 account (ArgentX).
Several problems have been discovered and solved (see hereunder).

## Usage related changes

- If you try `myAccount.execute()` with a Cairo 1 account, it works fine only if you use the type `Calldata`. With `RawArgs` type, an error occurs. It's now solved. (2 people reported this problem in Discord, but I thought they have made errors in the creation of their calldata...).

- Problem of `myAccount.execute()` solved in `src/utils/transaction.ts`.
- In account.test.ts, several `declare()` have been replaced by `declareIfNot()`.
- Funding of new accounts has been reduced to minimize the gETH needs.
- Simulate or estimateFee for Declare are failing if this class is already declared (it's the case in Testnet). These 2 tests are executed only in Devnet.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/account.test.ts
+++ b/__tests__/account.test.ts
@@ -24,6 +24,7 @@ import {
   compiledOpenZeppelinAccount,
   compiledStarknetId,
   compiledTestDapp,
+  describeIfDevnet,
   describeIfDevnetSequencer,
   erc20ClassHash,
   getTestAccount,
@@ -130,7 +131,7 @@ describe('deploy and test Wallet', () => {
         entrypoint: 'transfer',
         calldata: {
           recipient: tobeAccountAddress,
-          amount: uint256(1_000_000_000_000_000),
+          amount: uint256(5 * 10 ** 15),
         },
       });
       await account.waitForTransaction(transaction_hash);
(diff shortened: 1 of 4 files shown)


src/utils/transaction.ts
```diff
@@ -66,7 +74,10 @@ export const fromCallsToExecuteCalldata_cairo1 = (calls: Call[]) => {
   const orderCalls = calls.map((call) => ({
     contractAddress: call.contractAddress,
     entrypoint: call.entrypoint,
-    calldata: call.calldata,
+    calldata:
+      Array.isArray(call.calldata) && '__compiled' in call.calldata
```
penovicp, line 78:
```suggestion
      Array.isArray(call.calldata) && '__compiled__' in call.calldata
```

The compilation flag check is missing two underscores, so the ternary will always go through the second expression.

penovicp (approved):

__tests__/account.test.ts
```diff
@@ -228,15 +229,19 @@ describe('deploy and test Wallet', () => {
...
-    test('simulate DECLARE - Cairo 0 Contract', async () => {
-      const res = await account.simulateTransaction([
-        {
-          type: TransactionType.DECLARE,
-          contract: compiledErc20,
-        },
-      ]);
-      expect(res).toMatchSchemaRef('SimulateTransactionResponse');
+
+    describeIfDevnet('declare tests only on devnet', () => {
```
tabaktoni, line 233: It is strange this tests pass from Cairo0 Acc and fail from Cairo1 Acc

tabaktoni (approved):

Issue #796, opened 2023-10-27
Create Helper methods for TransactionExecutionStatus response


tabaktoni:
**Is your feature request related to a problem? Please describe.**
It is fairly complicated for a dev to detect what repose type it should get, as it needs to know the parameter-specific diff between response types.

**Describe the solution you'd like**
I imagine a solution to be in two parts, helper methods:
```ts
function isSuccess(response: TransactionExecutionStatus): SuccessfulTransactionReceiptResponse;
function isRejected(response: TransactionExecutionStatus): RejectedTransactionReceiptResponse;
function isReverted(response: TransactionExecutionStatus): RevertedTransactionReceiptResponse;
```
And one combined callback to use internally those 3 methods
```ts
function isResponse(response: TransactionExecutionStatus, onSuccessCallback, onRevertedCallback, onRejectedCallback)
```

**Additional context**
Final user would use it like:
```ts
const response = await account.waitForTransaction(declareAccount.transaction_hash);
isResponse(response, (success) => {
// here success should be type SuccessfulTransactionReceiptResponse
// do what you do with success...
}, (reverted) => {
// here reverted should be type RevertedTransactionReceiptResponse
// do what you do with reverted...
}, (rejected) => {
// here rejected should be type RejectedTransactionReceiptResponse
// do what you do with rejected...
})
```

**Alternative solution**
Feel free to propose better UX solution.

PhilippeR26:
I have started to work on this subject.
Any Rust devs would immediately say : use Rust enums ; it's made for a such case.
So, my idea is to code something that mimic a Rust Enum :
```typescript
const responseTx = TransactionResponse.of(await account.waitForTransaction(declareAccount.transaction_hash));
responseTx.switch({
    success(txR: SuccessfulTransactionReceiptResponse) { console.log("success =", txR) },
    rejected(txR: RejectedTransactionReceiptResponse) { console.log("rejected =", txR) },
    reverted(txR: RevertedTransactionReceiptResponse) { console.log("reverted =", txR) },
});
console.log("status = ", responseTx.status); // string
console.log("Is a success = ", responseTx.isSuccess); //boolean
console.log("Is rejected = ", responseTx.isRejected);
console.log("Is reverted = ", responseTx.isReverted);

```
As In Rust it's mandatory to code all cases, I will try to force the devs to handle the 3 switch cases.

PhilippeR26:
[...]

PR #851, opened 2023-11-22
feat: transactionResponse

PhilippeR26:
## Motivation and Resolution
Solve issue #796 
Create a new helper to detect what response type the user get, as it's rather complicated today.

## Usage related changes
Users can now use a transaction receipt as input for a `transactionResponse` function, and obtain an object that mimics a Rust Enum.
Example :
```typescript
const result = await account.execute(myCall);
const txReceipt = await provider.waitForTransaction(result.transaction_hash);
const txR = transactionResponse(txReceipt);
console.log(txR.match({
    Success: () => "success!",
    _: () => "unsuccess"
}));

console.log(txR.content);
console.log(txR.status);
console.log(txR.isSuccess());
console.log(txR.isRejected());

txR.match({
    Success: (txR: SuccessfulTransactionReceiptResponse) => { console.log("Success =", txR) },
    Rejected: (txR: RejectedTransactionReceiptResponse) => { console.log("Rejected =", txR) },
    Reverted: (txR: RevertedTransactionReceiptResponse) => { console.log("Reverted =", txR) },
    Error: (err: Error) => { console.log("An error occured =", err) },
});
```

- A new `txResponse` directory has been created in `utils`.
- A new class has been created : `TxResponseVariant`
- A proxyFactory is created, and when the variant is defined, the Proxy creates an instance of `TxResponseVariant`
 - tests : As I didn't found how to force the node/devnet to provide a `Rejected` response, I didn't tested this case. As the test duration is currently of 62s on my PC, I didn't added a `declare Success` test (can be easily added if necessary). 

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


PhilippeR26:
> I opened [PR#1 ](https://github.com/PhilippeR26/starknet.js/pull/1)with some modifications.
> 
> It boils down to two main alterations: changed the names so it is more apparent that the new entities are related to each other; merged the logic from `src/utils/txResponse/index.ts` and `src/utils/txResponse/txResponseVariant.ts` to simplify them.

PR merged.
Small question : is it still useful to perform checks
 on sequencer?

tabaktoni:
> > I opened [PR#1 ](https://github.com/PhilippeR26/starknet.js/pull/1)with some modifications.
> > It boils down to two main alterations: changed the names so it is more apparent that the new entities are related to each other; merged the logic from `src/utils/txResponse/index.ts` and `src/utils/txResponse/txResponseVariant.ts` to simplify them.
> 
> PR merged. Small question : is it still useful to perform checks on sequencer?

Not now, navbe we will have Seqeuncer as a Service later

tabaktoni:
All looks great, nice effort, the only issue I have with this is a diversion from the original idea, How do we now add this method to the response object? Originally we would map it as a non-enumerable prop.

PhilippeR26:
> All looks great, nice effort, the only issue I have with this is a diversion from the original idea, How do we now add this method to the response object? Originally we would map it as a non-enumerable prop.

Not sure to have understood.
I see 2 possibilities :
- As `waitForTransaction` is an old and a very common method, do not change anything to its output. Then create something that will analyse this output. It's what I proposed.
- Break the compatibility with the current interface, and the response of `waitForTransaction` is a complex object/class including some analysis functions.
Do you see a third way?

tabaktoni:
> > All looks great, nice effort, the only issue I have with this is a diversion from the original idea, How do we now add this method to the response object? Originally we would map it as a non-enumerable prop.
> 
> Not sure to have understood. I see 2 possibilities :
> 
> * As `waitForTransaction` is an old and a very common method, do not change anything to its output. Then create something that will analyse this output. It's what I proposed.
> * Break the compatibility with the current interface, and the response of `waitForTransaction` becomes a complex object/class including some analysis functions.
>   Do you see a third way?
>   Or may be return a such type as response (will be clash with `status` property) :
> 
> ```ts
> export type GetTransactionReceiptResponse = (
>   | SuccessfulTransactionReceiptResponse
>   | RevertedTransactionReceiptResponse
>   | RejectedTransactionReceiptResponse
> ) &
>   TransactionReceiptUtility;
> ```

Adding none enumerable methods would not break anything, you would have same object props, and the object will inherit from its proto methods like isSuccess, isRejected, isReverted, and isError.

with this one can do:

```ts
const txReceipt = await provider.waitForTransaction(result.transaction_hash);
txReceipt.isSuccess(() => {
  ... whatever on success do
})
```
  
But enumerable props from txReceipt will be the same.

PhilippeR26:
OK.
What I made : I modified the Receipt class constructor, setting all existing keys to `enumerable: false`, and adding all the properties of the received data, with `enumerable: true`.

I have added :
```typescript
export type GetTransactionReceiptResponse = GetTransactionReceiptResponseWoHelper & Receipt;
```

and getTransactionReceipt() ends with :
```typescript
return new Receipt(txReceiptWoHelper) as GetTransactionReceiptResponse;
``` 
and it works as you requested.

tabaktoni:
Is it possible to refactor this to target beta, and drop all sequencer stuff?

PR #857, opened 2023-11-30
docs: use only RpcProvider

PhilippeR26:
## Motivation and Resolution

Following removal of Starknet sequencers, users shall only use `RpcSequencer`.
The guides are here updated in accordance.
It's also the opportunity to migrate in the guides from Pythonic Devnet to Starknet-devnet-rs.

## Usage related changes
N/A

N/A

## Checklist:

- [x] All tests are passing

### DIFF
--- a/www/docs/guides/connect_contract.md
+++ b/www/docs/guides/connect_contract.md
@@ -11,7 +11,7 @@ You need 2 pieces of data:
 - the address of the contract
 - the ABI file of the contract (or the compiled/compressed contract file, that includes the abi)
 
-> If you don't have the abi file, the `provider.getClassAt()` and `provider.getClassByHash()` commands will recover the compressed contract file. As these methods generate a significant workload to the sequencer/node, it's recommended to store the result in your computer, to be able to reuse it later without using the provider each time:
+> If you don't have the abi file, the `provider.getClassAt()` and `provider.getClassByHash()` commands will recover the compressed contract file. As these methods generate a significant workload for the sequencer/node, it's recommended to store the result on your computer to be able to reuse it later without using the provider each time:
 
 ```typescript
 import fs from "fs";
--- a/www/docs/guides/connect_contract.md
+++ b/www/docs/guides/connect_contract.md
@@ -11,10 +11,11 @@ You need 2 pieces of data:
 - the address of the contract
 - the ABI file of the contract (or the compiled/compressed contract file, that includes the abi)
 
-> If you don't have the abi file, the `provider.getClassAt()` and `provider.getClassByHash()` commands will recover the compressed contract file. As these methods generate a significant workload to the sequencer/node, it's recommended to store the result in your computer, to be able to reuse it later without using the provider:
+> If you don't have the abi file, the `provider.getClassAt()` and `provider.getClassByHash()` commands will recover the compressed contract file. As these methods generate a significant workload to the sequencer/node, it's recommended to store the result in your computer, to be able to reuse it later without using each time the provider:
 
 ```typescript
 import fs from "fs";
+
 const compressedContract = await provider.getClassAt(addrContract);
 fs.writeFileSync('./myAbi.json', json.stringify( compressedContract.abi, undefined, 2));
 ```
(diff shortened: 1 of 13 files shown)


www/docs/guides/connect_contract.md
```diff
@@ -11,10 +11,11 @@ You need 2 pieces of data:
 - the address of the contract
 - the ABI file of the contract (or the compiled/compressed contract file, that includes the abi)
 
-> If you don't have the abi file, the `provider.getClassAt()` and `provider.getClassByHash()` commands will recover the compressed contract file. As these methods generate a significant workload to the sequencer/node, it's recommended to store the result in your computer, to be able to reuse it later without using the provider:
+> If you don't have the abi file, the `provider.getClassAt()` and `provider.getClassByHash()` commands will recover the compressed contract file. As these methods generate a significant workload to the sequencer/node, it's recommended to store the result in your computer, to be able to reuse it later without using each time the provider:
```
ivpavici, line 14:
```suggestion
> If you don't have the abi file, the `provider.getClassAt()` and `provider.getClassByHash()` commands will recover the compressed contract file. As these methods generate a significant workload to the sequencer/node, it's recommended to store the result in your computer, to be able to reuse it later without using the provider each time:
```

penovicp (commented):
I would suggest documenting that `new RpcProvider({ nodeUrl: constants.NetworkName.SN_GOERLI })` and `new RpcProvider()` are also possible and note that when using them a random public node will be selected, but also advise that using a specific `nodeUrl` is the better approach.

It might also make sense to mostly use one of those two options in most of the documentation or something like ``new RpcProvider({ nodeUrl: `${someNodeUrl}` })``  and use specific `nodeUrl` values within `www/docs/guides/connect_network.md` to simplify documentation maintenance (e.g. the Nethermind URLs will soon change).

ivpavici:
> I would suggest documenting that `new RpcProvider({ nodeUrl: constants.NetworkName.SN_GOERLI })` and `new RpcProvider()` are also possible and note that when using them a random public node will be selected, but also advise that using a specific `nodeUrl` is the better approach.
> 
> It might also make sense to mostly use one of those two options in most of the documentation or something like `` new RpcProvider({ nodeUrl: `${someNodeUrl}` }) `` and use specific `nodeUrl` values within `www/docs/guides/connect_network.md` to simplify documentation maintenance (e.g. the Nethermind URLs will soon change).

This is the new Juno link, but didn't test it yet if it's live: https://free-rpc.nethermind.io/mainnet-juno/v0_5

PhilippeR26:
> This is the new Juno link, but didn't test it yet if it's live:` https://free-rpc.nethermind.io/mainnet-juno/v0_5`
It's live (I have a script to batch check quickly a list of nodes). Is it a difference with 
`https://limited-rpc.nethermind.io/mainnet-juno/v0_5`?

PhilippeR26:
I completely rewrote `connect_network.md`, and i modified all node name as proposed.

www/docs/guides/what_s_starknet.js.md
```diff
@@ -25,7 +25,7 @@ Some important topics that have to be understood:
 
 > Understand what is Starknet and how it works is necessary. Then, you can learn how to interact with it using Starknet.js. So, at this stage, you should be aware of the content of the [Starknet official doc](https://docs.starknet.io/documentation/) and [the Starknet Book](https://book.starknet.io/).
 
-- Only the `Provider` object is talking directly to the network - your DAPP will talk mainly to `Account` and `Contract` objects. You will define with the `Provider` with which network you want to work. You can ask the Provider for some low-level data of the network (block, timestamp, ...).
+- Only the `RpcProvider` object is talking directly to the network - your DAPP will talk mainly to `Account` and `Contract` objects. You will define with the `RpcProvider` with which network you want to work. You can ask the provider for some low-level data of the network (block, timestamp, ...).
```
penovicp, line 28:
```suggestion
- Only the `RpcProvider` object communicates directly with the network; your DAPP will mainly interact with `Account` and `Contract` objects. You will define with the `RpcProvider` with which network you want to work. You can use the provider to access some low-level data from the network (block, timestamp, ...).
```

penovicp (approved):

PR #878, opened 2023-12-07
fix: apply bound for contract address from hash calculation

### LINKED ISSUE #876: calculateContractAddressFromHash
**Describe the bug**
I think the implementation is wrong. The cairo implementation calculates the hash but the modulus it by `const ADDR_BOUND = 2 ** 251 - MAX_STORAGE_ITEM_SIZE;` where `const MAX_STORAGE_ITEM_SIZE = 256;`

penovicp:
## Motivation and Resolution

Resolves #876 

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/src/utils/hash.ts
+++ b/src/utils/hash.ts
@@ -3,6 +3,7 @@
 import { poseidonHashMany } from '@scure/starknet';
 
 import {
+  ADDR_BOUND,
   API_VERSION,
   BN_FEE_TRANSACTION_VERSION_1,
   BN_FEE_TRANSACTION_VERSION_2,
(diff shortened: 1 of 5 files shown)


src/utils/hash.ts
```diff
@@ -205,13 +206,14 @@ export function calculateContractAddressFromHash(
...
 
-  return computeHashOnElements([
+  const hash = computeHashOnElements([
     CONTRACT_ADDRESS_PREFIX,
     deployerAddress,
     salt,
     classHash,
     constructorCalldataHash,
   ]);
+  return toHex(BigInt(hash) % ADDR_BOUND);
```
tabaktoni, line 216: I am not sure about this return if it will behave the same as Network, but I guess it will also return it to bounded value?

penovicp: I was surprised when the issue was reported since it never popped up, so I checked with [cairo-lang](https://github.com/starkware-libs/cairo-lang/blob/v0.12.3/src/starkware/starknet/common/storage.cairo#L8), [starknet.py](https://github.com/software-mansion/starknet.py/blob/v0.18.3/starknet_py/hash/address.py#L41), and [starknet-rs](https://github.com/xJonathanLEI/starknet-rs/blob/starknet-core/v0.7.2/starknet-core/src/utils.rs#L190). They use the same approach, it just seems the out of bounds values are rare so nobody noticed.

Issue #886, opened 2023-12-10
Cairo 2.4.0 new types : bytes31 & ByteArray


PhilippeR26:
**Is your feature request related to a problem? Please describe.**
I have analyzed the new Cairo v2.4.0 release, and searched what could be the impacts on Starknet.js.
I have found 2 new types that have been added in the abi :
- bytes31
- ByteArray

**Describe the solution you'd like**
Cairo team explained me how they are serialized :
### bytes31
is just a Bigint limited to 31 bytes, representing a ShortString.
```
"type": "core::bytes_31::bytes31"
```
It will be straightforward to implement.

### ByteArray
Here, it's  more complicated. This type is an object representing a LongString.
In Cairo, it's used as a normal string ; nobody see the low level object.
But this complex object is used in the abi (similar to enums, but slightly more complicated).
I have coded/compiled a small cairo example. The result in the abi is :
```typescript
{
      "type": "struct",
      "name": "core::byte_array::ByteArray",
      "members": [
        {
          "name": "data",
          "type": "core::array::Array::<core::bytes_31::bytes31>"
        },
        {
          "name": "pending_word",
          "type": "core::felt252"
        },
        {
          "name": "pending_word_len",
          "type": "core::integer::u32"
        }
      ]
    },
```
I think that on Starknet.js side, users want just to provide/receive a LongString in the calldata.
It means that we will not be in accordance with the abi for data input. 
Will probably be interesting also for @haroune-mohammedi .
Cairo team has explained me how to read/write this struct ; I coded something that is conform.
> __One question is remaining :__ without Abi, what should be the behavior of CallData.compile in case of a string in the input data?
> If it's a short string, deliver a Bigint, an array of Bigint, or a serialized ByteArray?
> If it's a long string, deliver an array of Bigint, or a serialized ByteArray?

**Describe alternatives you've considered**
N/A

**Additional context**
As Cairo v2.4.0 is implemented for the first time in Starknet v0.13.0, it could be a good thing to implement these 2 new types in Starknet.js v6.0.

PR #891, opened 2023-12-11
Cairo 2.4.0 new types : bytes31 & ByteArray

PhilippeR26:
## Motivation and Resolution
implement issue #886 

## Usage related changes

For older Cairo versions, the default response of` CallData.compile("fghdfghdfghdfghdfghfdghfgsdfgsdfgsdfgsdfg)` is modified. It's now necessary to use : `CallData.compile(shortString.splitLongString("gfddfghdfhdfghdfghdfghfdghdfghgfh");` to have an array of felts.

For Cairo 2.4.0, the result of CallData.compile("fghdfghdfghdfghdfghfdghfgsdfgsdfgsdfgsdfg)` is a serialized ByteArray

(see last message in the issue : https://github.com/starknet-io/starknet.js/issues/886#issuecomment-1850617944)

Added handling of Cairo  `bytes31` and `ByteArray` types.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


PhilippeR26:
tests where OK in Beta3.
In Beta 6, it seems that `rpcProvider.test.ts` is in an infinite loop.

tabaktoni:
> tests were OK in Beta3. In Beta 6, it seems that `rpcProvider.test.ts` is in an infinite loop.

Yes, Fixed

tabaktoni:
I would open discussion regarding this I don't think we should do it this way.
1. I want to eradicate the naive approach and leave it only for backward compatibility.
Also including last abi one call.

haroune-mohammedi:
Let's not merge this PR until we add typed tests since abiwan support latest cairo version too as per https://github.com/keep-starknet-strange/abi-wan-kanabi/pull/63

tabaktoni (approved):
Solution is lgtm, we should rethink how we handle Cairo types in js later on

Issue #858, opened 2023-11-30
account.verifyMessage() has an unclear response


PhilippeR26:
**Describe the bug**
The method `account.verifyMessage()` is performing OnLine the verification of an EIP712 message, with the following function signature :
```typescript
public async verifyMessage(typedData: TypedData, signature: Signature): Promise<boolean>
```
In some cases, this method can answer `false` even if the signature is valid.
Boolean response can handle only 2 states, but this method can encounter 3 states :
- Node is operating, account has been correctly defined (in particular the address), Signature is valid --> returns `true`.
- Node is operating, account has been correctly defined, Signature is not valid --> returns `false`.
- Node is not operating, or account has not been correctly defined (in particular the address), Signature is valid or is not valid --> returns `false` whatever the signature validity.
This last case is generating an answer that can be wrong.

Users have reported such problems in [Discord](https://discord.com/channels/793094838509764618/927918707613786162/1179557789631266937)

Same problem with `account.verifyMessageHash()`

**To Reproduce**
Use this script with a wrong provider or a wrong account address : https://github.com/PhilippeR26/starknet.js-workshop-typescript/blob/main/src/scripts/signature/4b.signEIP712Nested.ts 

**Expected behavior**
The method should inform the caller if the node reported a problem not related to the signature verification (for example something similar in the spirit to Rust Result).
I see 2 solutions :
1. To have a response only if no problem occurred, else Starknet.js throw an Error.
2. To have a response in any case, with such type of response format :
```typescript
type SignatureVerifResult = {
   isVerificationProcessed: boolean;
    isSignatureValid?: boolean;
   error?: Error    
}
```
As I am not a fan to throw errors when it can be avoided, I vote for the second proposal.

**Screenshots**
N/A

**Desktop (please complete the following information):**

- Browser & version [e.g. chrome, safari, webworker] N/A
- Node version [e.g. 16.0.1]
- Starknet.js version : v5.24.3
- Network [devnet, testnet] any

**Additional context**

PhilippeR26:
I can propose a PR for this subject.

PhilippeR26:
I just need a response about the 2 solutions proposed.

PhilippeR26:
Thanks for the proposal, but the solution code is ready.
Just that Starknet 0.13.0 is priority one for the moment.

PR #895, opened 2023-12-13
fix: solve wrong response for account.verifyMessage

PhilippeR26:
## Motivation and Resolution
implement issue #858 about wrong answers of `account.verifyMessage()`
See issue for details.

## Usage related changes

Breaking change : the result myAccount.verifyMessage(typedMessage, signature) is no more a boolean. It's now : 
```typescript
type SignatureVerifResult = {
  isVerificationProcessed: boolean;
  isSignatureValid?: boolean;
  error?: Error;
};
```

Example of usage :
```typescript
const myAccount = new Account(provider, accountAddress, "0x0123"); // fake private key
const result = await myAccount.verifyMessage(typedMessage, signature);
if (result.isVerificationProcessed) {
    console.log("Result (boolean) =", result.isSignatureValid);
} else {
    console.log("verification failed :", result.error);
}
```

There is no standard in Starknet for Cairo `IsValidSignature` account function response.
- OpenZeppelin 0.5.0 (cairo 0) and 0.8.0 (Cairo 2.3.1) are responding `0` in case of wrong signature.
- ArgentX 0.3.0 (Cairo 2.0.0) is reverting with a specific message.
- Same thing for Braavos (Cairo 0), with a different message.

The code is adapted in accordance.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing (in beta 3)


src/account/default.ts
```diff
@@ -573,23 +574,59 @@ export class Account extends Provider implements AccountInterface {
...
+        // Braavos Cairo 0 invalid signature
+        return {
+          isVerificationProcessed: true,
+          isSignatureValid: false,
+        } as SignatureVerifResult;
+      }
+      return {
+        isVerificationProcessed: false,
+        error: new Error('Signature verification request is rejected by the network.'),
+      } as SignatureVerifResult;
```
tabaktoni, line 622: isVerificationProcessed seems redundant with an error?
We should include err in Error?

I am more aligned with throwing errors:
1. This could silently break someone's code by not breaking after the update.
2. If the user does not want to break on network issue it can try-catch, as a user of networking libs I would assume it will throw on network issue.

tabaktoni, line 622:
```suggestion
      if (BigInt(resp.result[0]) === 0n) {
        // OpenZeppelin 0.8.0 invalid signature
        return false;
      }
      // OpenZeppelin 0.8.0, ArgentX 0.3.0 & Braavos Cairo 0 valid signature
      return true;
    } catch (err) {
      if (['argent/invalid-signature','is invalid, with respect to the public key'].includes((err as Error).message)) {
        // ArgentX 0.3.0 invalid signature, Braavos Cairo 0 invalid signature
        return false;
      }
      throw Error(`Signature verification request is rejected by the network: ${err}`)
```

tabaktoni: Also, this seems like a future issue for maintaining, are there any standards for this (if not we should propose one)? We can't cover all wallets and devs' Acc implementations in lib.

PhilippeR26: SNIP-6 is existing : https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-6.md
I don't know if this SNIP is accepted, but nobody is conform to it, neither for the function name, not for the return type.
So, I covered the 3 main account classes (ArgentX, Braavos, Openzeppelin).

PhilippeR26: Throwing Error implemented.

tabaktoni (approved):
Looking good for now, 
we can rethink-update support for C. Acc later on.

PR #908, opened 2023-12-22
[VIP] V6 saitama

tabaktoni:
## Motivation and Resolution
We have this structure in the project:

Starknet | <-> 1. Channel -> 2. Provider -> 3.Account -> 4.Contract

We have this type of structure:
RPC (namespace: types.RPC and RPC) (none in top namespace)
Provider (all in top namespace) (some inherited from RPC)
Account (all in top namespace)
Contract (all in top namespace)

Every level has a specific diff from prev one. inheriting prev. structure + diff
The channel will change with every RPC Release
Provider needs to stay in the same structure until the next major release.
Account and Contract are based on Provider so it will also stay + extend in the same structure until next major.

So it is clear that stability needs to be lined on Provider as so RPC types res and response used by provider need to be re-exported on the top level as "Provider types"
and so to the top.

This is a vertical structure, on the other hand, we have a horizontal structure like in Account - Signer -Hashing
An account provides loose types so that UX can be flexible on the other hand as you go deeper into signer it transforms it into strict union discriminatory types and at the bottom Hashing works only with the strictly defined types.

As I see it, I will need to export RPC Enums to the top level as Provider Enums, and req/res types as provider types. This would be the first step.
The second step will be to check horizontal types and polish them as we can.

The second issue is that we have floating definitions of classes and some types in lib and some other files that are floating around at the moment, they were historically union between sequencer and RPC in some unstructured mess. This would need to be checked and aligned with now only SSOT that is RPC Spec in such a way that it can be extended from 2 RPC Specs as we could potentially have in the future when changes are not in addition like v5 v6 but also restructured.

...

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


src/account/default.ts
```diff
@@ -73,9 +73,12 @@ export class Account extends Provider implements AccountInterface {
     address: string,
     pkOrSigner: Uint8Array | string | SignerInterface,
     cairoVersion?: CairoVersion,
-    transactionVersion: ETransactionVersion.V2 | ETransactionVersion.V3 = ETransactionVersion.V2 // TODO: Discuss this, set to v2 for backward compatibility
+    transactionVersion:
+      | ETransactionVersion.V2
+      | ETransactionVersion.V3 = DEFAULT_TRANSACTION_VERSION,
+    instance?: boolean
```
penovicp, line 79: This `instance` parameter and the new `AccountInstance` class feels like it could be confusing to users. I suggest evaluating if a different term might be a better fit, maybe something like `inheritable` or `inheritableProvider`.

Adding @ivpavici since he's had success with naming things.

src/channel/rpc_0_6.ts
```diff
@@ -52,22 +52,31 @@ export class RpcChannel {
...
+        optionsOrProvider?.default
+      );
     }
     this.retries = retries || defaultOptions.retries;
     this.headers = { ...defaultOptions.headers, ...headers };
     this.blockIdentifier = blockIdentifier || defaultOptions.blockIdentifier;
     this.chainId = chainId;
     this.waitMode = waitMode || false;
     this.requestId = 0;
+    this.speckVersion = speckVersion;
```
penovicp, line 79: `speckVersion` should be renamed to `specVersion`.

src/provider/interface.ts
```diff
@@ -34,7 +34,7 @@ import type {
 } from '../types';
 
 export abstract class ProviderInterface {
-  public abstract channel: RpcChannel;
+  public abstract channel: RpcChannel | undefined;
```
penovicp, line 37: This change doesn't seem necessary because of the `@ts-expect-error` in `src/provider/rpc.ts` on line 44.

src/provider/rpc.ts
```diff
@@ -34,8 +35,22 @@ export class RpcProvider implements ProviderInterface {
...
-  constructor(optionsOrProvider?: RpcProviderOptions | ProviderInterface | RpcProvider) {
-    if (optionsOrProvider && 'channel' in optionsOrProvider) {
+  constructor(
+    optionsOrProvider?: RpcProviderOptions | ProviderInterface | RpcProvider,
+    instance?: boolean
+  ) {
+    if (optionsOrProvider && instance) {
+      // stop ts complains
+      // @ts-expect-error
+      this.channel = null;
```
penovicp, line 45: Could also be done with a cast.

```suggestion
      this.channel = null as any;
```

src/types/index.ts
```diff
@@ -3,8 +3,11 @@ export * from './calldata';
...
 export * from './lib';
 export * from './provider';
+export * from './channel';
 export * from './signer';
 export * from './typedData';
 export * from './cairoEnum';
+export * from './api/forward';
 
 export * as RPC from './api';
+// export * from './api';
```
penovicp, line 13: Straggler?

src/types/lib/index.ts
```diff
@@ -155,52 +161,54 @@ export enum TransactionType {
   INVOKE = 'INVOKE_FUNCTION',
 }
 
+// TODO: All this should be removed
```
penovicp, line 164: Remove the commented blocks?

src/types/provider/response.ts
```diff
@@ -70,7 +71,7 @@ export type Storage = RPC.Felt;
 
 export type Nonce = string;
 
-export type SimulationFlags = RPC.SimulationFlags;
+// export type SimulationFlags = RPC.SimulationFlags;
```
penovicp, line 74: Straggler?

penovicp (commented):
Bunch of smaller nitpicks. In general lgtm.

src/provider/interface.ts
tabaktoni: I wasn't sure will ti effect external implementations

penovicp (approved):

tabaktoni:
TODO: Will close this one once extracted important parts.

tabaktoni:
Time to close this one as we did partial implementation already of some stuff here

PR #920, opened 2024-01-15
feat: expand structured data hashing in line with SNIP-12

penovicp:
## Motivation and Resolution

Implementation for [SNIP-12](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md).

The idea is for this PR that targets v5 to expand the available types and switch to the Poseidon hash, for v6 the validation will be overhauled to make it more comprehensive.

## Usage related changes

The changes should be backward compatible.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


ivpavici (approved):
Looks good from my perspective! Only thing we could discuss is it better to split some tests into couple of smaller ones, but it's not that important

src/utils/typedData.ts
```diff
@@ -107,130 +152,183 @@ function getMerkleTreeType(types: TypedData['types'], ctx: Context) {
...
-      .includes(type)
-  ) {
-    const structHashes: string[] = (data as unknown[]).map((struct) => {
-      // eslint-disable-next-line @typescript-eslint/no-use-before-define
-      return getStructHash(types, type.slice(0, -1), struct as Record<string, unknown>);
-    });
-    return [type, computeHashOnElements(structHashes)];
+  if (presetTypes[type]) {
+    return [type, getStructHash(presetTypes, type, data as TypedData['message'], revision)];
   }
```
sgc-code, line 212: IMO presets shouldn't be overridden by the other types

sgc-code, line 212: revision 0 has no presets

src/utils/typedData.ts
```diff
@@ -107,130 +152,183 @@ function getMerkleTreeType(types: TypedData['types'], ctx: Context) {
...
-      // eslint-disable-next-line @typescript-eslint/no-use-before-define
-      return encodeValue(types, merkleTreeType, struct as Record<string, unknown>)[1];
-    });
-    const { root } = new MerkleTree(structHashes as string[]);
-    return ['felt', root];
-  }
+  if (type.endsWith('*')) {
+    const isStructArray = Object.keys(types)
+      .map((x) => `${x}*`)
+      .includes(type);
```
sgc-code, line 217: not sure but looks like it's taking presets into account

src/utils/typedData.ts
```diff
@@ -107,130 +152,183 @@ function getMerkleTreeType(types: TypedData['types'], ctx: Context) {
...
-  if (type === 'selector') {
-    return ['felt', prepareSelector(data as string)];
+    const hashes: string[] = (data as Array<TypedData['message']>).map(mappingMethod);
+    return [type, revisionConfiguration[revision].hashMethod(hashes)];
   }
 
-  return [type, getHex(data as string)];
-};
+  switch (type) {
+    case 'enum': {
```
sgc-code, line 228: enums are revision 1 only

src/utils/typedData.ts
```diff
@@ -107,130 +152,183 @@ function getMerkleTreeType(types: TypedData['types'], ctx: Context) {
...
+      if (revision === Revision.Active) {
+        // TODO: should this be skipped for v5 and added as ByteArray in v6?
+        const elements = splitLongString(data as string).map(getHex);
+        return [type, revisionConfiguration[revision].hashMethod(elements)];
+      }
+      return [type, getHex(data as string)];
+    }
+    default: {
+      return [type, getHex(data as string)];
+    }
```
sgc-code, line 266: this is accepting many types that are not defined in the spec, it also allows all revision > 0 types for revision 0

sgc-code: also some types like shortstring need special treatment (length check)

src/utils/typedData.ts
```diff
@@ -20,18 +72,13 @@ function getHex(value: BigNumberish): string {
 
 /**
  * Validates that `data` matches the EIP-712 JSON schema.
- *
- * @param {any} data
- * @return {boolean}
  */
-const validateTypedData = (data: unknown): data is TypedData => {
+function validateTypedData(data: unknown): data is TypedData {
```
sgc-code, line 76: it would be nice to have some extra checks like:
- no types with empty name
- no types ending in `*`
- not types like in parenthesis
- not types that try to override the presets
- no duplicated types
- all enum variants types must be enclosed in parenthesis
- no unused types (types that are not being used by any other type)

src/utils/typedData.ts
```diff
@@ -1,12 +1,64 @@
...
+  ),
+  NftId: JSON.parse(
+    '[{ "name": "collection_address", "type": "ContractAddress" }, { "name": "token_id", "type": "u256" }]'
+  ),
+};
+
+function identifyRevision({ types, domain }: TypedData) {
+  if (revisionConfiguration[Revision.Active].domain in types && domain.revision !== Revision.Legacy)
+    return Revision.Active;
+
```
sgc-code, line 55: the revision field is mandatory in rev1, would be nice to check compliance

src/utils/typedData.ts
```diff
@@ -107,130 +152,183 @@ function getMerkleTreeType(types: TypedData['types'], ctx: Context) {
...
+    case 'selector': {
+      return ['felt', prepareSelector(data as string)];
+    }
+    case 'string': {
+      if (revision === Revision.Active) {
+        // TODO: should this be skipped for v5 and added as ByteArray in v6?
+        const elements = splitLongString(data as string).map(getHex);
+        return [type, revisionConfiguration[revision].hashMethod(elements)];
+      }
+      return [type, getHex(data as string)];
```
sgc-code, line 262: in revision v0 it would be great to enforce that the string fits into a single felt

src/utils/typedData.ts
```diff
@@ -107,130 +152,183 @@ function getMerkleTreeType(types: TypedData['types'], ctx: Context) {
...
 
-  if (type === 'selector') {
-    return ['felt', prepareSelector(data as string)];
+    const hashes: string[] = (data as Array<TypedData['message']>).map(mappingMethod);
+    return [type, revisionConfiguration[revision].hashMethod(hashes)];
   }
 
-  return [type, getHex(data as string)];
-};
+  switch (type) {
```
sgc-code, line 227: basic types like felt should no be overridden by the provided types

src/utils/typedData.ts
```diff
@@ -107,130 +152,183 @@ function getMerkleTreeType(types: TypedData['types'], ctx: Context) {
...
+ * Get the EIP-191|EIP-712 encoded message to sign, from the typedData object.
  */
-export const getMessageHash = (typedData: TypedData, account: BigNumberish): string => {
+export function getMessageHash(typedData: TypedData, account: BigNumberish): string {
   if (!validateTypedData(typedData)) {
     throw new Error('Typed data does not match JSON schema');
   }
 
+  const revision = identifyRevision(typedData) as Revision;
+  const { domain, hashMethod } = revisionConfiguration[revision];
```
yoga-braavos, line 324: Consider asserting that `domain` is well defined and throw a proper error in case it isn't
i.e. it have all the required fields
Currently I think it will only throw a non indicative error in `encodeData` during reduce in case `domain` is undefined?

src/utils/typedData.ts
penovicp: More robust validation will definitely be added. The only question is whether it will be within this PR or a separate one.

src/utils/typedData.ts
penovicp: Presets should be taken into account through the mapped encodeValue calls.

src/utils/typedData.ts
penovicp: Plan is to have it checked when the validation is reworked.

src/utils/typedData.ts
penovicp: Changed with the latest refactoring.

src/utils/typedData.ts
penovicp: It fails if it doesn't which is in line with the current behaviour. Should be made more robust with the validation rework.

src/utils/typedData.ts
penovicp: If this PR is merged to v5 there will be a validation rework done for v6. 

It shouldn't be possible for `domain` to end up as `undefined` since `identifyRevision` is called within `validateTypedData`.

penovicp (commented):
Modified the PR partially in line with the suggestions from the comments to reduce the revision overlap. 

The idea is for this PR that targets v5 to expand the available types and switch to the Poseidon hash, for v6 the validation will be overhauled to make it more comprehensive which would cover all the suggestions from the comments.

__tests__/utils/typedData.test.ts
```diff
@@ -1,54 +1,98 @@
...
+    encoded = encodeType(typedDataExample.types, 'Mail', TypedDataRevision.Active);
+    expect(encoded).toMatchInlineSnapshot(
+      JSON.stringify(
+        '"Mail"("from":"Person","to":"Person","contents":"felt")"Person"("name":"felt","wallet":"felt")'
+      )
+    );
+    encoded = encodeType(typedDataStructArrayExample.types, 'Mail', TypedDataRevision.Active);
+    expect(encoded).toMatchInlineSnapshot(
+      `"\\"Mail\\"(\\"from\\":\\"Person\\",\\"to\\":\\"Person\\",\\"posts_len\\":\\"felt\\",\\"posts\\":\\"Post*\\")\\"Person\\"(\\"name\\":\\"felt\\",\\"wallet\\":\\"felt\\")\\"Post\\"(\\"title\\":\\"felt\\",\\"content\\":\\"felt\\")"`
+    );
```
sgc-code, line 43:
```suggestion
    expect(encoded).toMatchInlineSnapshot(
      `"Mail"("from":"Person","to":"Person","posts_len":"felt","posts":"Post*")"Person"("name":"felt","wallet":"felt")"Post"("title":"felt","content":"felt")`
    );
```
this would be easier to read

src/utils/typedData.ts
```diff
@@ -107,130 +174,218 @@ function getMerkleTreeType(types: TypedData['types'], ctx: Context) {
...
+        ];
+      } // else fall through to default
+      return [type, getHex(data as string)];
+    }
+    case 'merkletree': {
+      const merkleTreeType = getMerkleTreeType(types, ctx);
+      const structHashes: string[] = (data as Array<TypedData['message']>).map((struct) => {
+        return encodeValue(types, merkleTreeType, struct, undefined, revision)[1];
+      });
+      const { root } = new MerkleTree(structHashes as string[]);
```
sgc-code, line 290: merkle tree should use poseidon in revision 1

src/utils/typedData.ts
```diff
@@ -107,130 +174,218 @@ function getMerkleTreeType(types: TypedData['types'], ctx: Context) {
...
+          ...byteArray.data,
+          byteArray.pending_word,
+          byteArray.pending_word_len,
+        ];
+        return [type, revisionConfiguration[revision].hashMethod(elements)];
+      } // else fall through to default
+      return [type, getHex(data as string)];
+    }
+    case 'felt':
+    case 'bool':
```
sgc-code, line 310: could have some check for booleans to avoid passing things like "3" here

src/utils/typedData.ts
sgc-code: IMO its easy to accomplish now, we just reorder to:
- check if preset
- check if array
- check other basic types
- check user defined types

and we can later add the extra validation with nice error message, but at least the priorities will be correct from the beginning

src/utils/typedData.ts
sgc-code: can you double check that?
I tried `TokenAmount*` and `isStructArray` was set to false instead of true

src/utils/typedData.ts
```diff
@@ -107,130 +174,218 @@ function getMerkleTreeType(types: TypedData['types'], ctx: Context) {
...
+        const targetType =
+          t.type === 'enum' && revision === Revision.Active
+            ? (t as StarkNetEnumType).contains
+            : t.type;
+        // parentheses handling for enum variant types
+        const typeString = targetType.match(/^\(.*\)$/)
+          ? `(${targetType
+              .slice(1, -1)
+              .split(',')
+              .map((e) => (e ? esc(e.trim()) : e))
```
sgc-code, line 199: i wouldn't trim trim it, regarding the spec `Thing` and `Thing ` are two separate types.
We could choose to reject request that are not trimmed but automatically trimming might be dangerous

src/utils/typedData.ts
sgc-code: fixed now

__tests__/utils/typedData.test.ts
penovicp: Snapshots are autogenerated.

src/utils/typedData.ts
penovicp: It would be a change for the existing behaviour, it could be split by revision but I want to avoid adding too many of those conditionals. Idea is to get the correct hash in this version and enforce rules in the following version.

src/utils/typedData.ts
penovicp: Will be added for v6 along with boundary checks for the other base types.

src/utils/typedData.ts
penovicp: It was getting taken into account by the recursive `encodeValue` call. I noticed the code could be simplified so that both custom and preset types rely on it so it should be clearer now.

src/utils/typedData.ts
penovicp: Good point, I added the trim so the `(u128, u128*) `example from the SNIP could pass. 

I removed the trimming for this PR, it would probably be best to re-add it after updating the SNIP so whitespace characters are excluded from string borders like the parenthesis rule. It would also make sense to exclude commas from the allowed characters for custom type names since they are used as enum delimiters.

tabaktoni (approved):
LGTM

src/utils/hash.ts
```diff
@@ -50,16 +50,30 @@ export function getVersionsByType(versionType?: 'fee' | 'transaction') {
     : { v1: transactionVersion, v2: transactionVersion_2 };
 }
 
+export function computePedersenHash(a: BigNumberish, b: BigNumberish): string {
```
tabaktoni, line 53: Maybe add some method descriptions

tabaktoni (approved):

PR #940, opened 2024-01-26
Feat CairoUint256

tabaktoni:
## Motivation and Resolution
`bnToUint256` and `cairo.uint256` should throw when provided with a negative param value.

### Feat

```ts 
new CairoUint256()
```

## Usage related changes
Throw on negative uint

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


src/utils/calldata/cairo.ts
```diff
@@ -102,6 +102,7 @@ export function getAbiContractVersion(abi: Abi): ContractVersion {
  */
 export const uint256 = (it: BigNumberish): Uint256 => {
   const bn = BigInt(it);
+  if (bn < 0) throw Error('uint256 must be positive number');
```
ivpavici, line 105: 3 small nitpicks:
1. `bn` could be renamed to `bi` as is down lower in `bnToUint256`
2. `uint256 must be positive number` could be extracted as constant and added here and in tests instead of c/p
3. I would maybe write: `uint256 must be **a** positive number`

penovicp: Could also call this method within `bnToUint256` and wrap the result properties with `toHex` to remove the duplication.

tabaktoni: 1. Let's be specific and call it bigInt.
2.  dep circle

tabaktoni: To Fix the dep circle I needed to clean up/restructure (hotfix with the parent file was out of sapient option).
Introduced new Structure Cairo data type, made external optional internal mandatory, backward comp.
b64abccf3888eba49460124a4d7aabbdca1592d1

I imagine every Cairo type to be like this.

src/utils/cairoDataTypes/uint256.ts
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  /**
+   * Validate if low and high can be represented as Unit256
+   */
+  static validateProps(low: BigNumberish, high: BigNumberish) {
+    const bigIntLow = BigInt(low);
+    const bigIntHigh = BigInt(high);
+    if (bigIntLow < UINT_256_LOW_MIN || bigIntLow > UINT_256_LOW_MAX) {
+      throw new Error('low is our of range UINT_256_LOW_MIN - UINT_256_LOW_MAX');
```
penovicp, line 74:
```suggestion
      throw new Error('low is out of range UINT_256_LOW_MIN - UINT_256_LOW_MAX');
```

src/utils/cairoDataTypes/uint256.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+    if (bigIntHigh < UINT_256_HIGH_MIN || bigIntHigh > UINT_256_HIGH_MAX) {
+      throw new Error('high is our of range UINT_256_HIGH_MIN - UINT_256_HIGH_MAX');
```
penovicp, line 77:
```suggestion
      throw new Error('high is out of range UINT_256_HIGH_MIN - UINT_256_HIGH_MAX');
```

src/utils/uint256.ts
```diff
@@ -1,36 +1,31 @@
...
-
-export const UINT_128_MAX = (1n << 128n) - 1n;
-export const UINT_256_MAX = (1n << 256n) - 1n;
+// Legacy support Export
+export { UINT_128_MAX, UINT_256_MAX };
 
 /**
  * Convert Uint256 to bigint
+ * Legacy support Export
+ * @
```
penovicp, line 11: `@` straggler.

Could also prefix `Legacy support Export` with `@deprecated`.

penovicp (approved):
Typo in two error messages. Otherwise, I like the rework, lgtm.

PR #943, opened 2024-01-30
docs: add migration to V6

PhilippeR26:
## Motivation and Resolution
In the frame of Starknet.js v6 pre-release, added in the migration guide the recommendations to migrate from v5 to v6.

## Usage related changes

Changed the start page : migration guide from v5

Do not forget to create a last V5 documentation variant, before the pre-release.

## Checklist:

### DIFF
--- a/www/docs/guides/migrate.md
+++ b/www/docs/guides/migrate.md
@@ -2,6 +2,84 @@
 sidebar_position: 18
 ---
 
+# Migrate from v5 to v6
+
+This document only covers the features present in v5 which have changed in some significant way in v6.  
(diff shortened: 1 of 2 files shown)


ivpavici (approved):
LGTM

tabaktoni (commented):
Account Class TransactionVersion V3 setup and note it works only with RPC 0.6+.
Default is v2 that works with 0.5+

This is not a breaking change but it is important for one to use v3 tx, so not sure should it be added here.

tabaktoni (approved):
Not sure we cover all but good starting point.

PR #967, opened 2024-02-17
Test: EstimateFee Bulk with Cario1 Acc 

RolandoDrRobot:
## Motivation and Resolution
This ia solution for this issue: https://github.com/starknet-io/starknet.js/issues/904

## Checklist:
- [x] Updated the tests
- [x] All tests are passing

### Test
Captura de pantalla 2024-03-28 a la(s) 8 29 21 a  m

### DIFF
--- a/__tests__/account.test.ts
+++ b/__tests__/account.test.ts
@@ -802,26 +802,6 @@ describe('deploy and test Wallet', () => {
       // expect(innerInvokeEstFeeSpy.mock.calls[0][1].version).toBe(feeTransactionVersion);
       // innerInvokeEstFeeSpy.mockClear();
     });
-  });
-
-  describe('EstimateFeeBulk with Cario1 Acc', () => {
-    test('estimateInvokeFee Cairo 1', async () => {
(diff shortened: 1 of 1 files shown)


tabaktoni (commented):
In a nutshell, during the transition period, we needed to test both cairo0 and cairo1 acc's. It should be done true assigning an account on a test run as an account defined c0 or c1 to be used.

We already have bulk tests that are proposed here.
At the current time, I believe we can dedicate to Cairo1 only test so no switch is required.

wdyt @penovicp @ivpavici

__tests__/account.test.ts
```diff
@@ -803,6 +803,51 @@ describe('deploy and test Wallet', () => {
       // innerInvokeEstFeeSpy.mockClear();
     });
   });
+
+  describe('EstimateFeeBulk with Cario1 Acc', () => {
+    test('estimateInvokeFee Cairo 1', async () => {
```
ivpavici, line 808: we already have this test above

RolandoDrRobot: u right! updated

PR #979, opened 2024-02-23
Move isomorphic-fetch to dev dependency

penovicp:
## Motivation and Resolution

Removes `isomorphic-fetch` as a production dependency because a fetch fallback should no longer be necessary in the majority of environments.

## Usage related changes

Most users should not be affected. If there are some edge cases that still need a fetch fallback, a `baseFetch` override is enabled for the `RpcProvider` and `RpcChannel` classes through which it can be set.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


tabaktoni (commented):
@dhruvkelawala @janek26 @yoga-braavos Can you check this one, will this still work for you ?

src/channel/rpc_0_6.ts
```diff
@@ -40,20 +40,22 @@ export class RpcChannel {
...
 
-  private speckVersion?: string;
+  private specVersion?: string;
 
-  readonly waitMode: Boolean; // behave like web2 rpc and return when tx is processed
+  private baseFetch: NonNullable<RpcProviderOptions['baseFetch']>;
 
   constructor(optionsOrProvider?: RpcProviderOptions) {
-    const { nodeUrl, retries, headers, blockIdentifier, chainId, waitMode } =
+    const { nodeUrl, retries, headers, blockIdentifier, chainId, waitMode, baseFetch } =
```
avimak, line 58: @ivpavici pls extract `specVersion` and assign it to `this` to avoid the extra rpc call for each new instance

avimak (changes requested):
i'm good with the main change, but pls note the degradation related to `specVersion`; generally speaking, we should have it assigned locally next to `this.chainId` in any constructor

src/channel/rpc_0_6.ts
ivpavici: thanks! this was addressed in this [PR](https://github.com/starknet-io/starknet.js/pull/1052) so we need to rebase and fix conflicts!

PR #980, opened 2024-02-23
feat: add fallback nodes to provider

### LINKED ISSUE #872: Add support for multiple RPC providers
Currently supports a single choice of RPC provider. To add support for failing over from a primary to a secondary RPC provider, based on an error response/no response from the primary

piwonskp:
## Motivation and Resolution
Resolves #872 

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/src/channel/rpc_0_6.ts
+++ b/src/channel/rpc_0_6.ts
@@ -19,6 +19,7 @@ import {
   waitForTransactionOptions,
 } from '../types';
 import { ETransactionVersion } from '../types/api';
+import assert from '../utils/assert';
 import { CallData } from '../utils/calldata';
 import { isSierra } from '../utils/contract';
 import fetch from '../utils/fetchPonyfill';
@@ -36,8 +37,6 @@ const defaultOptions = {
 };
 
 export class RpcChannel {
-  public nodeUrl: string;
-
   public headers: object;
 
   readonly retries: number;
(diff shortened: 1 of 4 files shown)


tabaktoni (commented):
1. Maybe we can keep nodeUrl, and pass param, if it is singular it would be non-fallback and, if it is an array it would have fallbacks. The logical order would be 0 then 1 ... if there are fallbacks. In that case, it would be the same as before.
2. I would rather add setNode and getNode on the channel than pre-pending fetch.
This way you intentionally change channel source rather than hacking into one method.

My initial idea for this was to dedicate to one and switch to the next when it fails, this solution proposes to use all nodes per request without dedication. This seems like an interesting idea and it could also work.

wdyt @penovicp @ivpavici

ivpavici:
> My initial idea for this was to dedicate to one and switch to the next when it fails

I thought of the same approach as you said here, I think Ethers works like this, but I have to double check.
I think if we want to go with this approach it wouldn't be a big change in this PR

src/channel/rpc_0_6.ts
```diff
@@ -52,38 +51,88 @@ export class RpcChannel {
...
+    const lastNode = nodes.pop();
+    assert(lastNode !== undefined);
+    let response;
+    for (let i = 0; i < nodes.length - 1; i += 1) {
+      try {
+        // eslint-disable-next-line no-await-in-loop
+        response = await this.fetch(nodes[i], method, params);
+
+        if (response.ok) {
+          this.setPrimaryNode(nodes[i], i);
```
ivpavici, line 115: I don't quite understand why we need this primaryNode here

tabaktoni:
TODO: I would like to refactor this approach and will do on the same PR

tabaktoni:
Closing in favor of the user managing it

Issue #987, opened 2024-02-28
account.verifyMessage() is not working with last Braavos account contract


PhilippeR26:
**Describe the bug**
Normally, all account contracts should be conform to [SNIP-6](https://github.com/starknet-io/SNIPs/blob/f6998f779ee2157d5e1dea36042b08062093b3c5/SNIPS/snip-6.md?plain=1#L61) about verification of message.
In fact it's not the case : some accounts are not using the right function name, some other are not providing the right response in case of success. And as SNIP-6 isn't defining a standard response in case of non success, everybody is responding his own way.

Here was the situation one month ago :
|account contract| camel case function name|snake case function name|response if success|response if fail|
|---:|:---:|:---:|:---:|:---:|
|Braavos <= 0.0.11| x | x| 0x01| Reverted, without specific message|
|ArgentX <= 0.3.0|x|x|0x01|Reverted, with "argent/invalid-signature" message|
|OpenZeppelin <= 0.6.1|x| |"VALID"|0x00|
|OpenZeppelin 0.7.0-0.8.0|x|x|"VALID"|0x00|

To handle all these account contracts, issue #858 and PR #895 have been created and implemented.
As the camel case function name `isValidSignature` is the only one common to all contracts, it has been used in Starknet.js

Recently, 3 new account contracts have been released :
- Braavos v1.0.0
- ArgentX v0.3.1
- OpenZeppelin v0.9.0

Now the situation is the following :
|account contract| camel case function name|snake case function name|response if success|response if fail|
|---:|:---:|:---:|:---:|:---:|
|Braavos <= 0.0.11| x | x| 0x01| Reverted, without specific message|
|Braavos 1.0.0|  | x| "VALID"| Reverted, with "INVALID_SIG" message|
|ArgentX <= 0.3.1|x|x|0x01|Reverted, with "argent/invalid-signature" message|
|OpenZeppelin <= 0.6.1|x| |"VALID"|0x00|
|OpenZeppelin 0.7.0-0.9.0|x|x|"VALID"|0x00|

There is no more 100% of account contracts that handles camel case function name, so the new Braavos contract is throwing an error, complaining that the function name is unknown.

**To Reproduce**
This code will fail :
```typescript
const signatureBR = await BRaccount.signMessage(typedMessage) as WeierstrassSignatureType;
console.log("BR signature = \n",num.toHex(signatureBR.r),"\n",num.toHex(signatureBR.s));
[...]

PR #989, opened 2024-03-01
fix: verification signature error with new braavos account

PhilippeR26:
## Motivation and Resolution
Solve issue #987 .
Accounts have now 2 possible names of message verification function. So it's necessary to search this name in the Abi prior of message verification.

## Usage related changes
No change

A special attention has been taken to minimize the interactions with the node to get the Abi.
At first use of an account instance, Starknet.js is asking the sierra file to the node, to find in the Abi the cairo version. 
Now, at the same time, the code try to find the name of the verification function. If it's an account behind a proxy, the request to the node of the implementation class Abi is performed only if the `account.verifyMessage()` is called.

- `getContractVersion()` is now also trying to get the verification function name ; so, it has been renamed `getContractSpecificities()`.
- `ContractVersion` type has been renamed `ContractSpecificities` has the function name has been included.
- add property `signatureVerifFunctionName` in Account.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/account.test.ts
+++ b/__tests__/account.test.ts
@@ -421,6 +421,10 @@ describe('deploy and test Wallet', () => {
     const signature = await account.signMessage(typedDataExample);
     const verifMessageResponse: boolean = await account.verifyMessage(typedDataExample, signature);
     expect(verifMessageResponse).toBe(true);
+    const messageVerificationFunctionName = account.signatureVerifFunctionName;
+    expect(
+      ['isValidSignature', 'is_valid_signature'].includes(messageVerificationFunctionName!)
+    ).toBe(true);
   });
 
   describe('Contract interaction with Account', () => {
--- a/src/account/default.ts
+++ b/src/account/default.ts
@@ -89,7 +89,6 @@ export class Account extends Provider implements AccountInterface {
       this.cairoVersion = cairoVersion.toString() as CairoVersion;
     }
     this.transactionVersion = transactionVersion;
-    this.signatureVerifFunctionName = undefined;
   }
 
   // provided version or contract based preferred transactionVersion
@@ -117,6 +116,8 @@ export class Account extends Provider implements AccountInterface {
    * Retrieves the Cairo version from the network and sets `cairoVersion` if not already set in the constructor.
    * Retrieves also the name of the function that verify a message signature.
    * @param classHash if provided detects Cairo version from classHash, otherwise from the account address
+
+   * Retrieves also the name of the function that verifies a message signature.
    */
   public async getCairoVersion(classHash?: string) {
     if (!this.cairoVersion) {
@@ -555,7 +556,7 @@ export class Account extends Provider implements AccountInterface {
   }
 
   public async verifyMessageHash(hash: BigNumberish, signature: Signature): Promise<boolean> {
-    // Accounts should be conform to SNIP-6 ( https://github.com/starknet-io/SNIPs/blob/f6998f779ee2157d5e1dea36042b08062093b3c5/SNIPS/snip-6.md?plain=1#L61 ), but they are not always conform, and SNIP do not standardize the response if the signature isn't valid.
+    // Accounts should conform to SNIP-6 (https://github.com/starknet-io/SNIPs/blob/f6998f779ee2157d5e1dea36042b08062093b3c5/SNIPS/snip-6.md?plain=1#L61), but they don't always conform. Also, the SNIP doesn't standardize the response if the signature isn't valid.
     if (typeof this.signatureVerifFunctionName === 'undefined') {
       await this.getCairoVersion();
     }
--- a/src/account/default.ts
+++ b/src/account/default.ts
@@ -40,6 +40,7 @@ import {
 } from '../types';
 import { ETransactionVersion, ETransactionVersion3, ResourceBounds } from '../types/api';
 import { CallData } from '../utils/calldata';
+import { createAbiParser } from '../utils/calldata/parser';
 import { extractContractHashes, isSierra } from '../utils/contract';
 import { starkCurve } from '../utils/ec';
 import { parseUDCEvent } from '../utils/events';
@@ -66,6 +67,8 @@ export class Account extends Provider implements AccountInterface {
 
   public cairoVersion: CairoVersion;
 
+  public signatureVerifFunctionName: string | undefined;
+
   readonly transactionVersion: ETransactionVersion.V2 | ETransactionVersion.V3;
 
   constructor(
@@ -86,6 +89,7 @@ export class Account extends Provider implements AccountInterface {
       this.cairoVersion = cairoVersion.toString() as CairoVersion;
     }
     this.transactionVersion = transactionVersion;
+    this.signatureVerifFunctionName = undefined;
   }
 
   // provided version or contract based preferred transactionVersion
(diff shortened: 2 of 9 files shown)


PhilippeR26:
Some tests have been performed (with success) with various types of accounts, outside of the test suite :
Capture d’écran du 2024-03-01 10-42-00

tabaktoni (commented):
In general, looking good.
I would suggest keeping ContractVersion instead of ContractSpecificities.
Even tho it is not a numeric string it determines the contract version.
It is also more intuitive (ContractSpecificities to me sound a little bit cryptic and non-deterministic)

@ivpavici @penovicp

PhilippeR26:
Name of variable modified.

src/provider/rpc.ts
```diff
@@ -203,18 +204,18 @@ export class RpcProvider implements ProviderInterface {
       .then(this.responseParser.parseContractClassResponse);
   }
 
-  public async getContractVersion(
+  public async getContractSpecificities(
```
ivpavici, line 207: @tabaktoni did you mean to rename also all instances of "specificities" in the PR?

tabaktoni: Yea

PhilippeR26: Modified

__tests__/account.test.ts
```diff
@@ -421,6 +421,10 @@ describe('deploy and test Wallet', () => {
     const signature = await account.signMessage(typedDataExample);
     const verifMessageResponse: boolean = await account.verifyMessage(typedDataExample, signature);
     expect(verifMessageResponse).toBe(true);
+    const messageVerificationFunctionName = account.signatureVerifFunctionName;
+    expect(
+      ['isValidSignature', 'is_valid_signature'].includes(messageVerificationFunctionName!)
+    ).toBe(true);
```
penovicp, line 427: Could also use the `toContain()` matcher: `expect([...]).toContain(functionName)`.

src/account/default.ts
```diff
@@ -86,6 +89,7 @@ export class Account extends Provider implements AccountInterface {
       this.cairoVersion = cairoVersion.toString() as CairoVersion;
     }
     this.transactionVersion = transactionVersion;
+    this.signatureVerifFunctionName = undefined;
```
penovicp, line 92: Shouldn't need to be explicitly set for `undefined`.

src/account/default.ts
```diff
@@ -110,15 +114,17 @@ export class Account extends Provider implements AccountInterface {
   }
 
   /**
-   * Retrieves the Cairo version from the network and sets `cairoVersion` if not already set in the constructor
+   * Retrieves the Cairo version from the network and sets `cairoVersion` if not already set in the constructor.
+   * Retrieves also the name of the function that verify a message signature.
```
penovicp, line 118:
```suggestion
   * Retrieves also the name of the function that verifies a message signature.
```

src/account/default.ts
```diff
@@ -549,28 +555,58 @@ export class Account extends Provider implements AccountInterface {
   }
 
   public async verifyMessageHash(hash: BigNumberish, signature: Signature): Promise<boolean> {
+    // Accounts should be conform to SNIP-6 ( https://github.com/starknet-io/SNIPs/blob/f6998f779ee2157d5e1dea36042b08062093b3c5/SNIPS/snip-6.md?plain=1#L61 ), but they are not always conform, and SNIP do not standardize the response if the signature isn't valid.
```
penovicp, line 558:
```suggestion
    // Accounts should conform to SNIP-6 (https://github.com/starknet-io/SNIPs/blob/f6998f779ee2157d5e1dea36042b08062093b3c5/SNIPS/snip-6.md?plain=1#L61), but they don't always conform. Also, the SNIP doesn't standardize the response if the signature isn't valid.
```

src/provider/rpc.ts
```diff
@@ -231,14 +232,37 @@ export class RpcProvider implements ProviderInterface {
       throw Error('getContractVersion require contractAddress or classHash');
     }
 
+    // Take the opportunity of class reading, to get the name of the message signature verification name
+    // or an empty string in case of proxy.
+    let messageVerifFunctionName: string | undefined;
+    if ('signatureVerifFunctionName' in this) {
+      const parser = createAbiParser(contractClass.abi);
```
penovicp, line 239: Looks like there is enough overlap here with the code added in the Account `verifyMessageHash` method that could be extracted to a utility method, it would do a superfluous proxy check for that method but it shouldn't be too costly.

src/provider/rpc.ts
```diff
@@ -231,14 +232,37 @@ export class RpcProvider implements ProviderInterface {
...
+      const parsedAbi = parser.getLegacyFormat();
+      const functionsList = parsedAbi.filter((abiElement) => abiElement.type === 'function');
+      if (functionsList) {
```
penovicp, line 242: This conditional can be removed, it will always be true. Or it could be modified to check the array length.

src/types/lib/index.ts
```diff
@@ -268,13 +268,15 @@ export interface CallStruct {
...
 /**
- * Represent Contract version
+ * Represent Contract version & account verification message function name
  */
 export type ContractVersion = {
   /** version of the cairo language */
   cairo: CairoVersion;
   /** version of the cairo compiler used to compile the contract */
   compiler: CompilerVersion;
+  /** If the contract is an account, the name of the function that verify a message signature. Otherwise returns '' */
```
penovicp, line 278:
```suggestion
  /** If the contract is an account, the name of the function that verifies a message signature. Otherwise returns '' */
```

penovicp (commented):
Left some tweak suggestions, in general lgtm.

It might be worth to consider a shorter approach of forgoing the function name detection and having a boolean `useLegacySignatureValidator` configuration property that determines between using `isValidSignature` or the snake case equivalent. The snake case would be the default and for unsuccessful validation for older OZ accounts an additional warning would be logged to retry with `useLegacySignatureValidator` set to `true`.

__tests__/account.test.ts
PhilippeR26: Do you want it to be modified?

src/provider/rpc.ts
PhilippeR26:
```suggestion
      if (functionsList.length > 0) {
```

PhilippeR26: Something like that?

src/provider/rpc.ts
PhilippeR26: Are you sure?
In detail, seems rather different...

PhilippeR26:
With an account that has been instantiated with CairoVersion as parameter, the `signatureVerifFunctionName` is never defined. So, I have added just this code : 
https://github.com/starknet-io/starknet.js/pull/989/commits/9c2726105955e26f60d036a7c1bf85932d0d64df
A consequence is that this test is now failing at line 649 : 
https://github.com/starknet-io/starknet.js/blob/a85d48ee73acb1365da6bef3f9d3a65153f9a422/__tests__/account.test.ts#L641-L679

```
LibraryError: RPC: starknet_getClassAt with params {
      "block_id": "pending",
      "contract_address": "0x311c7b427829ff25aac8a7a3f1332f4baef1c5d3069d5eca5808a905910a9ad"
    }
     
            20: Contract not found: undefined
```
In the JSdoc, it's written that for estimateFeeBulk, the account  has to be deployed :
https://github.com/starknet-io/starknet.js/blob/a85d48ee73acb1365da6bef3f9d3a65153f9a422/src/account/interface.ts#L148
But the test is trying to use account.estimateFeeBulk() with a non deployed account.
So, either the JSDoc of account.estimateFeeBulk() is wrong, or its test should use a deployed account. But there is an error somewhere.

What do you think is wrong?

tabaktoni:
Closing in favor of PR https://github.com/starknet-io/starknet.js/pull/989 based on this branch

PR #992, opened 2024-03-03
feat: implement RPC v0.7.0-rc2 specification

penovicp:
## Motivation and Resolution

Updates support to the latest RPC specification release.

[v0.7.0-rc1](https://github.com/starkware-libs/starknet-specs/releases/tag/v0.7.0-rc1)

## Usage and development related changes

For low-level interactions the RPC specification types have been split to 0.6 and 0.7 with corresponding `RpcChannel` classes for each. The `RpcProvider` uses unified types that should be compatible, however, the unification mapping might have some unforeseen edge cases.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


penovicp:
Taking into account that the specification updates are mostly expansions I kept going back and forth between merging the supported specifications and having them separated (as it is in the PR). The downside of the former is it necessitates tracking the manual changes, while the latter is significantly more verbose and requires complex type unification for the `RpcProvider` class. Another approach, now that the `Merge` mapped typed is available, might be to use it for only the divergent base types from the specification.

That choice is the main thing I would like to draw attention to in the PR and welcome fresh pairs of eyes taking a look.

ivpavici (approved):
LGTM... in a sense that we will for sure be able to simplify this a lot in the future... once RPC becomes more stable and all rpc providers hopefully support only one version

tabaktoni:
Provider Mapping looks troublesome to me, i would rather leave v6 rpc 0.6 as the original mapping and do response parsing.

PR #995, opened 2024-03-04
feat: don't sign when skipValidate is true

dhruvkelawala:
## Motivation and Resolution

Signing should be skipped if skipValidate is true. There is no need for signing in this case, because we send empty signature anyways.

For example, on ledger you would need to sign a transaction even if skipValidate is true for feeEstimation which shouldn’t be the case

...

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/account/default.ts
+++ b/src/account/default.ts
@@ -134,7 +134,12 @@ export class Account extends Provider implements AccountInterface {
     calls: AllowArray<Call>,
     details: UniversalDetails = {}
   ): Promise<EstimateFee> {
-    const { nonce: providedNonce, blockIdentifier, version: providedVersion } = details;
+    const {
+      nonce: providedNonce,
+      blockIdentifier,
+      version: providedVersion,
+      skipValidate = true,
+    } = details;
 
     const transactions = Array.isArray(calls) ? calls : [calls];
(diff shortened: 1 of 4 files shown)


tabaktoni (approved):
LGTM!

To be clear, this is an optimization of execution, removing unnecessary signing calculations when not needed.

penovicp (approved):
lgtm

Did a rebase and added a test fix so it can be merged.

src/account/default.ts
```diff
@@ -673,13 +687,15 @@ export class Account extends Provider implements AccountInterface {
...
-      classHash,
-      compiledClassHash: compiledClassHash as string, // TODO: TS Nekuzi da v2 nemora imat a v3 mora i da je throvano ako nije definiran
-      senderAddress: details.walletAddress,
-    });
+    const signature = !details.skipValidate
+      ? await this.signer.signDeclareTransaction({
+          ...details,
+          ...v3Details(details),
+          classHash,
+          compiledClassHash: compiledClassHash as string, // TODO: TS Nekuzi da v2 nemora imat a v3 mora i da je throvano ako nije definiran
```
ivpavici, line 695: @tabaktoni what about this Croatia comment :D

Issue #997, opened 2024-03-06
Bugs in response parser, for tuple type


PhilippeR26:
**Describe the bug**
Some bugs have been identified when a smart-contract response is including a tuple (now very common with Cairo 1) :
- If the tuple includes a `Result` enum, the parse is wrong (because the Result type includes a `,` : `Result<u64, u8>`, and the code is then lost.
- if it's nested tuples (more than 2 levels), it's not handled.

**To Reproduce**
Some examples of types that are not handled correctly :
```
((u256,(u16,Order2)), u8)
(Result<u64, u8>, u8)
(tupleResponse::tupleResponse::Direction, core::integer::u8) // where Direction type can be a Result type
```

**Expected behavior**
To handle correctly complex tuples responses, at least for Cairo 1.

**Screenshots**
N/A

**Desktop (please complete the following information):**

- Browser & version [e.g. chrome, safari, webworker]
- Node version [e.g. 16.0.1]
- Starknet.js version v6.1.4
- Network [devnet, testnet]

**Additional context**
N/A

PR #998, opened 2024-03-06
fix: tuples response parser, for nested tuples, or tuple including Result

PhilippeR26:
## Motivation and Resolution
Solve issue #997 

- If the tuple includes a `Result` enum, the parse is wrong (because the Result type includes a `,` : `Result<u64, u8>`
- if it's nested tuples (more than 2 levels), it's not handled by the parser.

## Usage related changes
Such Abi outputs are now handled (in Cairo 1):
```
((u256,(u16,Order2)), u8)
(Result<u64, u8>, u8)
(tupleResponse::tupleResponse::Direction, core::integer::u8) // where Direction type can be a Result type
```

- the tuple parser was splitting the abi output string, using `,` as splitter, and was not able to handle more than 2 levels of tuples. This code has been completely rewritten.
- Tests : a dedicated test contract has been added, and tests have been added in `Cairo1v2.test.ts`

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


penovicp (commented):
I wanted to test a refactoring suggestion and I noticed that some of the tests don't fully check the code.

On this [branch](https://github.com/starknet-io/starknet.js/commit/38e21a22a54cc7fb60df30186919aeca08c82225#diff-f7dca081cb62e38c337dcdf23e048124e55164c90ca6212d0a920bceb9a347d7) you can see that part of the code is commented out but the [tests still pass](https://github.com/penovicp/starknet.js/actions/runs/8177837252). The code does get passed through when it is not removed, but when it is removed the fallback conditional is sufficient to produce the expected result for the corresponding tests.

PhilippeR26:
> I wanted to test a refactoring suggestion and I noticed that some of the tests don't fully check the code.
> 
> On this [branch](https://github.com/starknet-io/starknet.js/commit/38e21a22a54cc7fb60df30186919aeca08c82225#diff-f7dca081cb62e38c337dcdf23e048124e55164c90ca6212d0a920bceb9a347d7) you can see that part of the code is commented out but the [tests still pass](https://github.com/penovicp/starknet.js/actions/runs/8177837252). The code does get passed through when it is not removed, but when it is removed the fallback conditional is sufficient to produce the expected result for the corresponding tests.

It should be better now.
I have already tried a refactor of this part of the code, using `.some()`, but i had an error message due to variable `posStart`.

PR #1012, opened 2024-03-14
Adjust max amount and make fee margins configurable

penovicp:
## Motivation and Resolution

The PR alters the resource bound max amount calculation in line with the recommendation from the Starknet v0.13.1 pre-release notes. It also makes the fee margins configurable at Provider/Account class initialization.

## Usage related changes

- The max amount bound calculation is modified
- Users can fine-tune fee margins through the `ProviderOptions` constructor parameter
- The `overhead` parameter of the `estimatedFeeToMaxFee` utility function is changed from using a decimal notation to using a percentage notation (e.g. 0.2 to 20) to be in line with the `estimateFeeToBounds` function

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/src/provider/rpc.ts
+++ b/src/provider/rpc.ts
@@ -31,15 +31,17 @@ import { isSierra } from '../utils/contract';
 import { RPCResponseParser } from '../utils/responseParser/rpc';
 
 export class RpcProvider implements ProviderInterface {
-  private responseParser = new RPCResponseParser();
+  private responseParser: RPCResponseParser;
 
   public channel: RPC07.RpcChannel | RPC06.RpcChannel;
 
   constructor(optionsOrProvider?: RpcProviderOptions | ProviderInterface | RpcProvider) {
     if (optionsOrProvider && 'channel' in optionsOrProvider) {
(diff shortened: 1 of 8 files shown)


src/provider/rpc.ts
```diff
@@ -176,7 +178,7 @@ export class RpcProvider implements ProviderInterface {
     // can't be named simulateTransaction because of argument conflict with account
     return this.channel
       .simulateTransaction(invocations, options)
-      .then(this.responseParser.parseSimulateTransactionResponse);
+      .then((r) => this.responseParser.parseSimulateTransactionResponse(r));
```
tabaktoni, line 181: Any specific reason for changing to implicit providing params ?

tabaktoni (approved):
LGTM!

src/provider/rpc.ts
penovicp: With the changes in this PR `RPCResponseParser` has internal `this` references in the parsing methods so their usage needed to be adapted for the proper `this` to be accessible and changing it to arrow functions was shorter than doing `.bind(this.responseParser)`.

PR #1019, opened 2024-03-15
fix: repair i128 typed data encoding and add typed data range checks

### LINKED ISSUE #1002: `encodeType` of typed data does not encode preset types correctly
**Describe the bug**
`encodeType` of type data does not encode preset types correctly. 

You can verify this from the SNIP:
- [X is an u256](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md#when-x-is-a-nft-id)
- [X is a TokenAmount](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md#when-x-is-a-token-amount)
- [X is a Nft ID](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md#when-x-is-a-nft-id)

Where for every preset type, it says:
> It will be encoded as the following object.

And per the [X is an object](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md#when-x-is-an-object), every object should encoded as follows:
> `escape(name) || "(" || escape(param1_name) || ":" || escape(param1_type) || "," || ... || escape(paramN_name) || ":"|| escape(paramN_type) || ")"`
> If the object references other objects/enum which can also reference other objects/enums, the set of referenced objects/enums is collected, sorted by name, and appended to the encoding.

The spec also refers to [this repo](https://github.com/argentlabs/starknet-off-chain-signature) for examples, where preset type `u256` is [included along with its properties](https://github.com/argentlabs/starknet-off-chain-signature/blob/19ad573c0def1dc2f0a08d3d63fabdbccab1d6f6/src/struct_with_u256.cairo#L12).

**To Reproduce**
Run this test:
https://github.com/starknet-io/starknet.js/blob/fdcc7fbf5957c14df765fd78508db6539b6124e3/__tests__/utils/typedData.test.ts#L50

**Expected behavior**
All preset types are encoded as objects:
```
"Example"("n0":"TokenAmount","n1":"NftId")"NftId"("collection_address":"ContractAddress","token_id":"u256")"TokenAmount"("token_address":"ContractAddress","amount":"u256")"u256"("low":"u128","high":"u128")
```
**Actual behavior**
All preset types are encoded as basic types:
```
"Example"("n0":"TokenAmount","n1":"NftId")
```

- Starknet.js version cb2059039e3b42501cdfefec7802da83eb73645b

penovicp:
## Motivation and Resolution

Resolves #1002 (i128 typed data encoding). Also adds typed data range checks for primitive types and a "StarkNet" casing update for typed data types.

## Usage related changes

The encoding changes should not affect users that use valid data, if the data is outside the permitted range it will generate errors with the code from this PR. If users are using some of the updated types they will need to update them to their new name.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/src/utils/typedData.ts
+++ b/src/utils/typedData.ts
@@ -3,9 +3,9 @@ import { PRIME, RANGE_FELT, RANGE_I128, RANGE_U128 } from '../constants';
 import {
   BigNumberish,
   TypedDataRevision as Revision,
-  StarkNetEnumType,
-  StarkNetMerkleType,
-  StarkNetType,
+  StarknetEnumType,
+  StarknetMerkleType,
+  StarknetType,
   TypedData,
 } from '../types';
 import assert from './assert';
@@ -107,7 +107,7 @@ export function prepareSelector(selector: string): string {
   return isHex(selector) ? selector : getSelectorFromName(selector);
 }
 
-export function isMerkleTreeType(type: StarkNetType): type is StarkNetMerkleType {
+export function isMerkleTreeType(type: StarknetType): type is StarknetMerkleType {
   return type.type === 'merkletree';
 }
 
@@ -142,7 +142,7 @@ export function getDependencies(
 
   return [
     type,
-    ...(types[type] as StarkNetEnumType[]).reduce<string[]>(
+    ...(types[type] as StarknetEnumType[]).reduce<string[]>(
       (previous, t) => [
         ...previous,
         ...getDependencies(types, t.type, previous, t.contains, revision).filter(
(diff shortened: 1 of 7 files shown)


src/utils/typedData.ts
```diff
@@ -305,15 +312,42 @@ export function encodeValue(
...
+      } // else fall through to default
+      return [type, getHex(data as string)];
+    }
     case 'felt':
-    case 'bool':
-    case 'u128':
-    case 'i128':
-    case 'ContractAddress':
+    case 'shortstring': {
+      // TODO: should 'shortstring' diverge into directly using encodeShortString()?
```
penovicp, line 332: I want to draw attention to the approach taken with the `shortstring` encoding. It was implemented by reusing the existing `felt` encoding. 

As a basic example, a provided value `"hello"` will be encoded as `0x68656c6c6f`, this I believe to not be contentious. The part where I believe there might be issues with is the encoding of numerical strings, the existing `felt` encoding treats both numbers and numerical strings as the same value and such behaviour might be undesirable/unexpected for the `shortstring` type. As an example `"2"` will be encoded as `0x2` while someone coming from a Cairo background will probably assume it to be encoded as `0x32`.

tabaktoni (approved):
IGWT!

PR #1021, opened 2024-03-18
simplify extends for the account class

0xknwn:
## Motivation and Resolution

I am working with Starknet abstract account. To support our work, I am extending the starknet.js `Account` class to add some additional behaviours on the application side. I am **not** changing the way the transaction version is used internally and would still need to implement the account deployment.

You have defined 3 `private` methods in the class which makes sense to not expose to the outside. However in the context of extending that class we would prefer to access those 3 methods instead of rewriting them. Obviously, this is your decision to make. For that particular scenario, I think having those 3 `private` methods turned into `protected` makes more sense.

- not applicable

## Usage related changes

- It will not affect any user that is not extending the account class (i.e. most developers)
- It will simplify the ability to extend the `Account`

- We did not add any test as the protected functions can still not be accessed from the outside and the code can easily be copied/pasted into a new `private` function anyway.
- If you think something should be done to enhanced my request, let me know.

## Checklist:

- [ ] Updated the tests
- [x] All tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+# [6.5.0](https://github.com/starknet-io/starknet.js/compare/v6.4.2...v6.5.0) (2024-03-14)
+
+### Bug Fixes
+
+- adjust max amount bound calculation for RPC v0.7.0 ([dd34cdb](https://github.com/starknet-io/starknet.js/commit/dd34cdb8b9817a55a16a97d960b1544d75c0059a))
+
+### Features
+
+- make fee margins configurable ([cedd984](https://github.com/starknet-io/starknet.js/commit/cedd984e1106db5b73d17630e282eb956d344a97))
+
 ## [6.4.2](https://github.com/starknet-io/starknet.js/compare/v6.4.1...v6.4.2) (2024-03-14)
 
 ### Bug Fixes
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,8 @@ Or run tests in watch mode:
 npm test --watch
 ```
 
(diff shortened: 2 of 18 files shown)


tabaktoni (approved):
LGTM!

tabaktoni:
If you can please do share an example of your AA implementation it could be a valuable resource for others.

ivpavici (approved):

0xknwn:
> If you can please do share an example of your AA implementation it could be a valuable resource for others.

I will for sure! Both the Cairo and the Starknet.js implementation. I just need some time to get it clean

Also thank to the 2 of you @tabaktoni and @ivpavici for this lightning merge!

0xknwn:
> 🎉 This PR is included in version 6.6.0 🎉

Actually, no it is not. Can you double-check? It could be because of the error in the CI. Thank you

ivpavici:
> > 🎉 This PR is included in version 6.6.0 🎉
> 
> Actually, no it is not. Can you double-check? It could be because of the error in the CI. Thank you

yeah the github-actions post here is wrong, it will be included in the next version! 🙏

penovicp:
The notification does look like a bug. My guess is that it might stem from an overlap arising from the `next-version` branch name, that when identifying PRs to notify the plugin expands the set with PRs that have been merged into branches that have themselves been merged into the configured release branch without or with faulty verification of the included commits. Essentially my presumption is the following sequence: `next-version` merged to `develop`; this PR merged to `next-version`; manual release; faulty notification for all merges into `next-version` and not just from the first step.

I haven't found an existing corresponding issue and would need to do a deep dive into the release plugin code to corroborate my assumption.

Issue #1030, opened 2024-03-21
signTransaction missing abi arg in Signer class


stanleyyconsensys:
**Describe the bug**

In Account class, the execute method will ask the signer to sign the transaction via signTransaction

it pass in transactions, signerDetails, abis

however in Signer Class, signTransaction method only contains 2 args

in https://github.com/starknet-io/starknet.js/blob/v6.6.0/src/signer/default.ts#L46

**Desktop (please complete the following information):**

- Browser & version [e.g. chrome, safari, webworker]
- Node version : 18.18.2
- Starknet.js version: 6.60
- Network [devnet, testnet]: testnet

dhruvkelawala:
But we don't need abis on starknet. So it's optional. We can remove it at this point. I don't think it solves any purpose

stanleyyconsensys:
i agree, can we remove the abi from all interface for signTransaction?

PhilippeR26:
I think that we can remove it from the Signer interface, but we have to keep it as parameter in `account.execute()`  to not create a breaking change.

PhilippeR26:
If agreed this way, I have a PR to propose.

ivpavici:
@tabaktoni a breaking PR could still be done now since v6 is not officially released?

PhilippeR26:
The syntax
```typescript
const res = await account.execute(myCall, undefined, { maxFee: 1 * 10 ** 15 });
```
is very common.
I think we should not break it.

penovicp:
I think we should defer to Phil's perspective on this.

A thing that could be done is that the method could be refactored to allow for truncated parameters (`execute(calls, details)`) together with the current notation (`execute(calls, abis, details)`) and mark it for future deprecation.

tabaktoni:
So let's remove it from the signer and overload execute

PR #1045, opened 2024-03-26
fix: remove unnecessary abis parameter from signer and .execute

PhilippeR26:
## Motivation and Resolution
Solve issue #1030 .
`Account.execute()` uses an optional `abis` parameter that is useless. Same for `Signer` interface.

- Signer optional `abis` parameter is deleted.
- `Account.execute()` accepts both signatures with and without `abis` parameter (to not break the current syntax).

## Usage related changes
All these syntaxes are valid :
```typescript
const res1 = await account0.execute(myCall,undefined,{maxFee:10n**17n});
const res2 = await account0.execute(myCall);
const res3 = await account0.execute(myCall,{maxFee:10n**17n});
```

- Abis removed of Signer interface
- Account interface with execute method overloaded.
- Added a test of the new syntax.

**Nota** : Guides are automatically modified by CI during commit. I think that this has to be deactivated. Documentation content has to be handled only by the author.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/account.test.ts
+++ b/__tests__/account.test.ts
@@ -370,17 +370,16 @@ describe('deploy and test Wallet', () => {
     expect(balance.low).toStrictEqual(toBigInt(990));
   });
 
-  test('execute with syntax without abi parameter', async () => {
-    const { transaction_hash } = await account.execute(
-      {
-        contractAddress: erc20Address,
-        entrypoint: 'transfer',
-        calldata: [erc20.address, '10', '0'],
-      },
-      { maxFee: 6n * 10n ** 15n }
-    );
+  test('execute with and without deprecated abis parameter', async () => {
+    const transaction = {
+      contractAddress: erc20Address,
+      entrypoint: 'transfer',
+      calldata: [erc20.address, '10', '0'],
+    };
+    const details = { maxFee: 0n };
 
-    await provider.waitForTransaction(transaction_hash);
+    await expect(account.execute(transaction, details)).rejects.toThrow(/zero/);
+    await expect(account.execute(transaction, undefined, details)).rejects.toThrow(/zero/);
   });
 
   test('execute with custom nonce', async () => {
--- a/__tests__/account.test.ts
+++ b/__tests__/account.test.ts
@@ -370,6 +370,19 @@ describe('deploy and test Wallet', () => {
     expect(balance.low).toStrictEqual(toBigInt(990));
   });
 
+  test('execute with syntax without abi parameter', async () => {
+    const { transaction_hash } = await account.execute(
+      {
+        contractAddress: erc20Address,
+        entrypoint: 'transfer',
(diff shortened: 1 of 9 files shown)


__tests__/account.test.ts
```diff
@@ -370,6 +370,19 @@ describe('deploy and test Wallet', () => {
...
+      {
+        contractAddress: erc20Address,
+        entrypoint: 'transfer',
+        calldata: [erc20.address, '10', '0'],
+      },
+      { maxFee: 6n * 10n ** 15n }
+    );
+
+    await provider.waitForTransaction(transaction_hash);
+  });
```
penovicp, line 384:
```suggestion
  test('execute with and without deprecated abis parameter', async () => {
    const transaction = {
      contractAddress: erc20Address,
      entrypoint: 'transfer',
      calldata: [erc20.address, '10', '0'],
    };
    const details = { maxFee: 0n };

    await expect(account.execute(transaction, details)).rejects.toThrow(/zero/);
    await expect(account.execute(transaction, undefined, details)).rejects.toThrow(/zero/);
  });
```
The suggestion makes the test explicitly about the parameter notation.

src/account/default.ts
```diff
@@ -305,11 +305,34 @@ export class Account extends Provider implements AccountInterface {
...
+    transactionsDetail?: UniversalDetails
+  ): Promise<InvokeFunctionResponse>;
+  public async execute(
+    transactions: AllowArray<Call>,
+    abis?: Abi[],
+    transactionsDetail?: UniversalDetails
+  ): Promise<InvokeFunctionResponse>;
+  public async execute(
+    transactions: AllowArray<Call>,
+    arg2: Abi[] | UniversalDetails | undefined = undefined,
```
penovicp, line 318:
```suggestion
    arg2?: Abi[] | UniversalDetails,
```

src/account/default.ts
```diff
@@ -305,11 +305,34 @@ export class Account extends Provider implements AccountInterface {
...
+        break;
+      case typeof arg2 === 'object':
+        details = arg2;
+        break;
+      case typeof arg2 === 'undefined':
+        details = transactionsDetail;
+        break;
+      default:
+        throw Error(`wrong input parameters: ${arg2}, ${transactionsDetail}`);
+    }
```
penovicp, line 334:
```suggestion
    const details = arg2 === undefined || Array.isArray(arg2) ? transactionsDetail : arg2;
```
This omits the explicit error but parameter validation is rarely used in the library and the code will fail anyway for faulty parameters.

penovicp (commented):
Left some suggestions, in general lgtm

PhilippeR26 (commented):
All done.

ivpavici (approved):

penovicp (approved):

PR #1051, opened 2024-03-28
refactor: create wallet namespace

PhilippeR26:
## Motivation and Resolution
Wallet API functions have to be imported one by one `from "starknet"`
Create a `wallet` namespace is solving this problem, and will group all methods related to this subject.

## Usage related changes

image

`connect.ts` renamed `wallet.ts`

## Checklist:

- [x] All tests are passing


src/index.ts
```diff
@@ -43,6 +43,7 @@ export * from './utils/calldata/enum';
 export * from './utils/contract';
 export * from './utils/events';
 export * from './utils/transactionReceipt';
+export * as wallet from './wallet/wallet';
```
ivpavici, line 46: I know there was a discussion about this on Discord, but if we have a better idea not to have `wallet/wallet` ? I know naming is the hardest thing in development and I'm nitpicking xD
cc @tabaktoni

PhilippeR26: Do you prefer a file called `walletAPI.ts` ?

penovicp: One possibility is to use `export * as wallet from './wallet'` in `src/wallet/index.ts` in place of `export * from './connect'` instead of removing it (could even keep the `connect` file name).

PhilippeR26: I went back to `connect.ts`, keeping `export * as wallet`.

tabaktoni (approved):
Ok, lgtm

PR #1061, opened 2024-04-03
Test RPC default Providers Services

tabaktoni:
## Motivation and Resolution
This issue https://github.com/starknet-io/starknet.js/issues/1058 motivated me to add a test for the default RPC version defined by lib and the default RPC version server by RPC.

This test does not ensure any security just print results.
Could we maybe add some CI warning or action for testing RPC versions?
We could set expectations in this two tests like
- we expect 0.7 as the default version from RPC on base route
- we expect 0.7 as the lib default version and all rpc should have 0.7 implemented

Some feedback @ivpavici @penovicp 

Ensure and Test RPC Providers

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/__tests__/defaultNodes.test.ts
+++ b/__tests__/defaultNodes.test.ts
@@ -1,0 +1,46 @@
+import { Provider, constants } from '../src';
+
+describe('Default RPC Nodes', () => {
+  test('Default lib configured version', async () => {
+    const result = await Promise.all(
+      Object.keys(constants.RPC_NODES).map(async (network: any) => {
+        return Promise.all(
+          constants.RPC_NODES[network as keyof typeof constants.RPC_NODES].map(async (it: any) => {
+            const provider = new Provider({ nodeUrl: it });
+            return {
+              network,
+              nodeUrl: provider.channel.nodeUrl,
+              version: await provider.getSpecVersion(),
+            };
+          })
+        );
+      })
+    );
+
+    console.table(result.flat());
+  });
+  test('Default RPC configured version on base route', async () => {
+    function getBaseUrl(url: string) {
+      const pathArray = url.split('/');
+      // Hotfix for juno but not generic
+      const net = pathArray[3].includes('juno') ? `${pathArray[3]}` : '';
+      return `${pathArray[0]}//${pathArray[2]}/${net}`;
+    }
+    const result = await Promise.all(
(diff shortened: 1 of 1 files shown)


PhilippeR26:
There are 42 commits, with 36 files impacted!
Where is the code related to this verification of rpc spec?

__tests__/defaultNodes.test.ts
```diff
@@ -0,0 +1,46 @@
...
+      const pathArray = url.split('/');
+      // Hotfix for juno but not generic
+      const net = pathArray[3].includes('juno') ? `${pathArray[3]}` : '';
+      return `${pathArray[0]}//${pathArray[2]}/${net}`;
+    }
+    const result = await Promise.all(
+      Object.keys(constants.RPC_NODES).map(async (network: any) => {
+        return Promise.all(
+          constants.RPC_NODES[network as keyof typeof constants.RPC_NODES].map(async (it: any) => {
+            const provider = new Provider({ nodeUrl: getBaseUrl(it) });
```
ivpavici, line 33: except for `getBaseUrl(it)` the test here and above are the same, so we can extract it maybe and send as param!

ivpavici:
> We could set expectations in this two tests like
> 
> we expect 0.7 as the default version from RPC on base route
> we expect 0.7 as the lib default version and all rpc should have 0.7 implemented

I agree... also am curious for more thoughts!

penovicp:
I think it's a good idea.

Might be worth considering doing this as a separate script since it would bypass the environment setup for the regular tests.

PhilippeR26:
Just for information, Blast is responding to `provider.getSpecVersion()`  with : `0.7.0-rc2`

PhilippeR26:
I understand that you want to check if all default networks (list hard coded in the project) are really using default_url_version.
All 4 url includes at the end the default_url_version :
https://github.com/starknet-io/starknet.js/blob/b3a2946a3d1e89fd16c542cf19e4fcedb6777ef5/src/constants.ts#L75-L82
Logically, it should be OK by definition, but it's never bad to verify.
But I don't understand why you want to check the rpc versions of the base urls ... as we are not using them.

tabaktoni:
Closing this with resolution on spec version with Provider.create auto detect specification.
As so no need for this testings.

PR #1075, opened 2024-04-11
docs: types format fix for generated docs

ivpavici:


src/types/lib/index.ts
```diff
@@ -209,7 +211,9 @@ export type BlockNumber = BlockTag | null | number;
 
 /**
  * hex string and BN are detected as block hashes
+ *
```
tabaktoni, line 214: Can you drop this one it is in my PR

tabaktoni (approved):

Issue #1073, opened 2024-04-10
export function parseCalldataField


PhilippeR26:
**Is your feature request related to a problem?**
Today, we can decode a Starknet response by type (struct or enums) with  a method of the CallData class :
```typescript
decodeParameters(typeCairo: AllowArray<string>, response: string[]): AllowArray<Result>
 ```
 but we have nothing exported to perform the opposite action : encode for a type.
 
 In the frame of a DAPP I am creating, I need this function.

**Describe the solution you'd like**
A quick solution is to just export the function `parseCalldataField` from `requestParser.ts`

**Describe alternatives you've considered**
Something more refined could be done, by adding a method in the CallData class, with more user friendly inputs.
Not sure it will be useful by many devs, so not sure we should spend many time on it.

**Additional context**
If agreed, I have a PR ready for this.

PR #1078, opened 2024-04-11
Export function parseCalldataField()

PhilippeR26:
## Motivation and Resolution
Solve issue #1073

## Usage related changes
can now use :
```typescript
import { calldata } from "starknet";
const res = calldata.parseCalldataField(...);
```

Export the function `parseCalldataField` from `requestParser.ts`
Exported in `calldata` namespace.

## Checklist:

- [x] All tests are passing

### DIFF
--- a/src/utils/calldata/index.ts
+++ b/src/utils/calldata/index.ts
@@ -37,6 +37,7 @@ import validateFields from './validate';
 
 export * as cairo from './cairo';
 export * as byteArray from './byteArray';
+export * as calldata from './requestParser';
 
 export class CallData {
   abi: Abi;


src/utils/calldata/index.ts
```diff
@@ -37,6 +37,7 @@ import validateFields from './validate';
 
 export * as cairo from './cairo';
 export * as byteArray from './byteArray';
+export * as calldata from './requestParser';
```
tabaktoni, line 40:
```suggestion
export { parseCalldataField } from './requestParser';
```

tabaktoni (approved):
As temp solution this is fine, just refine existing Feature issue to refactor this eventually.

Another thing is that requestParser expose only parseCalldataField method so I would rather do as suggested. Another point is that we have too many calldata names already.

penovicp (approved):

Issue #1077, opened 2024-04-11
Replace ENUM with companion pattern approach


lukasaric:
**Is your feature request related to a problem? Please describe.**
Currenlty,  TS `enum` is used across the codebase. I would be good to be replaced with the `companion pattern` approach to increase the type safety.
To have better understanding, here's a good thread why it would be good to avoid enums in TS
[Don't use Enums in Typescript, they are very dangerous](https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh#:~:text=Excessive%20use%20of%20regular%20ENUMs,new%20values%20can%20be%20added.)

**Describe the solution you'd like**
BEFORE
```ts
export enum NetworkName {
  SN_MAIN = 'SN_MAIN',
  SN_GOERLI = 'SN_GOERLI',
  SN_SEPOLIA = 'SN_SEPOLIA',
}
```
AFTER
```ts
export const NetworkName = {
  SN_MAIN: 'SN_MAIN',
  SN_GOERLI: 'SN_GOERLI',
  SN_SEPOLIA: 'SN_SEPOLIA',
} as const;

// eslint-disable-next-line @typescript-eslint/no-redeclare -- intentionally naming the variable the same as the type
export type NetworkName = (typeof NetworkName)[keyof typeof NetworkName];
```

**Describe alternatives you've considered**
Pros:
- Increased type safety
- Less prone to mistakes and errors (especially for numerics)
- Decreased output size and improved performance

Cons:
- Eslint `re-declare` rule does not have exception for declaring the same name type and variable. Either, the rule should be disabled or we'll add eslint ignore for these cases which will be annoying

**Additional context**

ivpavici:
@kateberryd sorry, Luka reached out earlier to solve this (when he opened the issue), but I forgot to assign him... 
Keep an eye on other issues, there will soon be more!

PR #1083, opened 2024-04-17
Chore/replace enums with companion pattern approach

lukasaric:
Solves #1077

## Motivation and Resolution
To increase the type safety, this PR introduces the `enum` replacement with the `companion pattern` approach.
To have a better understanding, here's a good thread on why it would be good to avoid enums in TS

[Don't use Enums in Typescript, they are very dangerous](https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh#:~:text=Excessive%20use%20of%20regular%20ENUMs,new%20values%20can%20be%20added.)

## Usage related changes
_n.a_

Here's the example of before vs after usage:

BEFORE
```ts
export enum NetworkName {
  SN_MAIN = 'SN_MAIN',
  SN_GOERLI = 'SN_GOERLI',
  SN_SEPOLIA = 'SN_SEPOLIA',
}
```
AFTER
```ts
export const NetworkName = {
  SN_MAIN: 'SN_MAIN',
  SN_GOERLI: 'SN_GOERLI',
  SN_SEPOLIA: 'SN_SEPOLIA',
} as const;

// eslint-disable-next-line @typescript-eslint/no-redeclare -- intentionally naming the variable the same as the type
export type NetworkName = (typeof NetworkName)[keyof typeof NetworkName];
```

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


src/constants.ts
```diff
@@ -24,46 +25,65 @@ export const MAX_STORAGE_ITEM_SIZE = 256n;
...
-  L1_BOUND_MAX_PRICE_PER_UNIT = 50,
-  MAX_FEE = 50,
-}
+export const BaseUrl = {
+  SN_MAIN: 'https://alpha-mainnet.starknet.io',
+  SN_GOERLI: 'https://alpha4.starknet.io',
+  SN_SEPOLIA: 'https://alpha-sepolia.starknet.io',
+} as const;
+
+// eslint-disable-next-line @typescript-eslint/no-redeclare -- intentionally naming the variable the same as the type
```
lukasaric, line 39: @ivpavici 
The eslint disable is here needed because of the limitations of `eslint no-redeclare rule`. They even recommended adding comments like this, but it can be painful since we have this occurrence in a lot of places now.
We have two options:
- Disable the `@typescript-eslint/no-redeclare` rule (which is not the deal breaker)
- Write comments where it's needed (technically the only valid case is when we use same name for type and const on purpose)

lukasaric: @ivpavici Moreover, it would be nice that we do `enum` replacement in the `starknet/types` repo.

lukasaric: Here's the PR on `types/js` :
https://github.com/starknet-io/types-js/pull/4

ivpavici: @tabaktoni please advise :)

tabaktoni (approved):
lgtm

ivpavici (approved):

PR #1095, opened 2024-04-23
fix: remove [warning] from typedoc for external usage

0xknwn:
running `npx typedoc --entryPoints src --out build` generates a number of warnings that are addressed here.  This PR replaces #1092. It is based on `next-version` and the changes are now split by commits so that they are easy to understand and if needed can be cherry-picked

## Motivation and Resolution

Some of the issues addressed that are small enhancements to use the starknet.js library. Others are more opinionated or improvements. For a better understanding, changes have been grouped per commits. This is the list of commits and fixes:

- fix: rename the @param to match the definitions
- fix: add a space before typedoc \<br/\> so that it catches the name
- fix: improve function declaration so that typedoc matches it
- fix: reintroduce the parameter name
- fix: remove unused type declaration and export
- ~~fix: export underlying types for RPC 0.6~~
- ~~fix: export the underlying types with the composite types~~

### RPC version (if applicable)

0.6 and 0.7

## Usage related changes

None

## Development related changes

None

## Checklist:

- [x] Performed a self-review of the code
- [x] Rebased to the last commit of the target branch (or merged it into my branch)
- [ ] Linked the issues which this PR resolves
- [x] Documented the changes in code (API docs will be generated automatically)
- [ ] Updated the tests
- [ ] All tests are passing

PR #1098, opened 2024-04-24
fix: cannot infer (TS2742) types from starknet-types@0.7

0xknwn:
the externalization of types in starknet-types that come with starknet.js 6.8 prevents typescript from inferring types that are not available in the project. To workaround this issue, this commit declares the SPEC type, helping the typescript inference.

## Motivation and Resolution

The typescript language server reports 4 errors like the one below in the project :

```
The inferred type of 'signMessage' cannot be named without a reference to
'../../node_modules/starknet-types-07/dist/types/api/components'. This is likely
not portable. A type annotation is necessary.ts(2742)
function signMessage(swo: StarknetWindowObject, typedData: TypedData): Promise<SIGNATURE>
```

> Note: this error does not prevent the build

0.7

## Usage related changes

None

None

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/account/default.ts
+++ b/src/account/default.ts
@@ -1,3 +1,5 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { SPEC } from 'starknet-types-07';
 import { UDC, ZERO } from '../constants';
 import { Provider, ProviderInterface } from '../provider';
 import { Signer, SignerInterface } from '../signer';
--- a/src/utils/stark.ts
+++ b/src/utils/stark.ts
@@ -1,3 +1,5 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { SPEC } from 'starknet-types-07';
 import { getStarkKey, utils } from '@scure/starknet';
 import { gzip, ungzip } from 'pako';
 
--- a/src/wallet/account.ts
+++ b/src/wallet/account.ts
@@ -3,6 +3,8 @@ import {
   type AddStarknetChainParameters,
   type NetworkChangeEventHandler,
   type WatchAssetParameters,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  type SPEC,
 } from 'starknet-types-07';
 
 import { Account, AccountInterface } from '../account';
--- a/src/wallet/connect.ts
+++ b/src/wallet/connect.ts
@@ -8,6 +8,8 @@ import {
   type ChainId,
   type StarknetWindowObject,
   type TypedData,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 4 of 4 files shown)


PhilippeR26 (commented):
I suppose that SPEC can't be imported by users. Maybe should be considered (I had the problem this morning in a DAPP).
I ask me if we should name it SPEC, or SPEC07, as it's specific for this rpc version.

0xknwn:
> I suppose that SPEC can't be imported by users. Maybe should be considered (I had the problem this morning in a DAPP). I ask me if we should name it SPEC, or SPEC07, as it's specific for this rpc version.

We cannot import the SPEC in our project to fix the issue that is, as I understand, on starknet.js: this happens when I use typescript inside starknet.js with the TS language server from vscode and with `npx typedoc`. The later is our actual issue. You you can check our project that extends the starknet.js account at [0xknwn/starknet-modular-account](https://github.com/0xknwn/starknet-modular-account).

Do you get the error yourself? Maybe I am trying to fix an issue that is not one? Run `npx typedoc --entryPoints src --out build` from the current develop branch of starknet.js and it should break on the fact that is cannot infer the types.

Long story short, I could declare the output of those functions that are failing by writing the return explicitly but I would have to import `SPEC` anyway. I would not have to use the eslint rule though. What is better is up to you. I also do not really understand why sometimes the project does return type declaration for the function and sometimes it does not.

The problem is actually related to the SPEC RPC version only because v07 is externalized now. I have the issue on the Account with `getUniversalSuggestedFee`. This function is protected so it is actually usable (and used 😜) by our code.

Thank you for your help. Huge thank you to all the people working on starknet.js That is a great project!

tabaktoni:
Legit, but will need to complete other name scoping also

0xknwn:
@PhilippeR26, @raizo07, @tabaktoni 

My initial issue was with typedoc but that makes sense that you do have the same issue with Docusaurus as shown in #1095 which is also my PR btw. Typescript cannot infer types that it does not know (i.e. are not imported in the project). That is exactly why I did the import.

If you do not like the eslint rule (which I do not like either!), you can resolve the type in the function spec by providing the returned type. A way or another, we need to help typescript with 6.8 so that it correctly handle the externalized types.

PhilippeR26:
I let @tabaktoni defines the strategy.

PhilippeR26 (commented):
I let @tabaktoni defines the strategy.

ivpavici (approved):

ivpavici:
@0xknwn please join OnlyDust so that I can send you the reward!
image

0xknwn:
> please join OnlyDust so that I can send you the reward!

done @ivpavici

PR #1119, opened 2024-05-02
feat: add getL1MessageHash RPC function

### LINKED ISSUE #702: Add provider function to get L1 message hash (where applicable)
It would be great if the SDK supports fetching and computing the L1 message hash of transactions resulting from L1->L2 interactions

A provider function `getL1MessageHash` which receives an L2 tx hash, and returns the L1 message hash, if it is applicable.

**Context**
L1 message hashes are computed in the following way [link to docs](https://docs.starknet.io/documentation/architecture_and_concepts/L1-L2_Communication/messaging-mechanism/#structure_and_hashing_l1-l2).

A snippet of solidity code to compute this:
```
pragma solidity >=0.8.2 <0.9.0;
contract Hasher {
  function getL1ToL2MsgHash(
        address fromAddress,
        uint256 toAddress,
        uint256 selector,
        uint256[] calldata payload,
        uint256 nonce
    ) external view returns (bytes32) {
        return
            keccak256(
                abi.encodePacked(
                    uint256(uint160(fromAddress)),
                    toAddress,
                    nonce,
                    selector,
                    payload.length,
                    payload
                )
            );
    }
}
```

The required information to compute this on L2 is available when querying `getTransactionByHash`
Example:
mainnet tx: [0x0075f4ad64b9eca93b4e1bb48fa3db4260e476fcf65e7ab97bf5f60a2a3022ce](https://voyager.online/tx/0x7fb25547e3be9aa8db2e97cc830b4b5d487110a4e42e65b45924b9a4e2fe5f7)

The result of `getTransactionbyHash`
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": {
    "calldata": [
      "0xae0ee0a63a2ce6baeeffe56e7714fb4efe48d419",
      "0x1d8ec38501c2fddf0b17fcd38df2029fb059b2016889e6962e5f19449735108",
      "0x354a6ba7a180000",
      "0x0"
    ],
    "contract_address": "0x73314940630fd6dcda0d772d4c972c4e0a9946bef9dabf4ef84eda8ef542b82",
    "entry_point_selector": "0x2d757788a8d8d6f21d1cd40bce38a8222d70654214e96ff95d8086e684fbee5",
    "nonce": "0x15b1a",
    "transaction_hash": "0x75f4ad64b9eca93b4e1bb48fa3db4260e476fcf65e7ab97bf5f60a2a3022ce",
    "type": "L1_HANDLER",
    "version": "0x0"
  }
}
```

The first param of `calldata` == `fromAddress`.
`contract_address` == `toAddress`
`nonce` == `nonce`
`entry_point_selector` == `selector`

`payload` and `payload.length` can be derived from the calldata field.

In case the L2 tx is not a result of an l1->l2 interaction, the function can return null (or error)

This can be a good first issue

princeibs:
## Motivation and Resolution
Closes #702
...

...

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


src/channel/rpc_0_6.ts
```diff
@@ -22,11 +22,12 @@ import { CallData } from '../utils/calldata';
 import { isSierra } from '../utils/contract';
 import { validateAndParseEthAddress } from '../utils/eth';
 import fetch from '../utils/fetchPonyfill';
-import { getSelector, getSelectorFromName } from '../utils/hash';
+import { computeHashOnElements, getSelector, getSelectorFromName } from '../utils/hash';
 import { stringify } from '../utils/json';
 import { getHexStringArray, toHex, toStorageKey } from '../utils/num';
 import { Block, getDefaultNodeUrl, isV3Tx, isVersion, wait } from '../utils/provider';
 import { decompressProgram, signatureToHexArray } from '../utils/stark';
+import { useEncoded } from '../utils/starknetId';
```
ivpavici, line 30: I don't think we used anything from StarknetId utils anywhere else in the code, doesn't look like we should do it here as well? cc: @tabaktoni

tabaktoni: Yep, not sure this is a correct encoder.

princeibs: That was the closest I found that could carry out the operation. You know of any I could use instead? @tabaktoni

princeibs:
> Please add a test for this feature.

image
I added a test but later removed it because tx hashes are only valid on their respective chains and the test cases are ran across various chains, which will make the test not to pass.
@tabaktoni

tabaktoni:
> > Please add a test for this feature.
> 
> image I added a test but later removed it because tx hashes are only valid on their respective chains and the test cases are ran across various chains, which will make the test not to pass. @tabaktoni

We run tests on devnet and Sepolia:
you can use describeIfDevnet & describeIfTestnet (this will be Sepolia only) to isolate tests based on env
https://github.com/starknet-io/starknet.js/blob/f92135235a69db3c190d506a9d950a54dc43e75c/__tests__/config/fixtures.ts#L115

src/channel/rpc_0_6.ts
tabaktoni: It is not only about using encode, the result of the getL1MessageHash should be **L1 message hash**.
Did you check that this produces the correct hash?

src/channel/rpc_0_6.ts
```diff
@@ -168,6 +169,22 @@ export class RpcChannel {
     return this.fetchEndpoint('starknet_getBlockWithTxs', { block_id });
   }
 
+  public async getL1MessageHash(txHash: BigNumberish) {
+    const txData = (await this.getTransactionByHash(txHash)) as any;
+    const { calldata, contract_address, entry_point_selector, nonce, version, type } = txData;
+    const hashedCallData = computeHashOnElements(calldata);
+
+    return computeHashOnElements([
+      useEncoded(type),
```
PhilippeR26, line 178: I think it's not the right encoding. Use the same encoding used for calculation of transaction hash : `shortString.encodeShortString()`

src/channel/rpc_0_7.ts
```diff
@@ -168,6 +169,22 @@ export class RpcChannel {
     return this.fetchEndpoint('starknet_getBlockWithTxs', { block_id });
   }
 
+  public async getL1MessageHash(txHash: BigNumberish) {
+    const txData = (await this.getTransactionByHash(txHash)) as any;
```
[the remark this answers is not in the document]
princeibs: @PhilippeR26 Done

src/channel/rpc_0_6.ts
princeibs: Done

src/channel/rpc_0_6.ts
princeibs: > I think it should work since the formula is implemented correctly 🙂. Can you help with a sample to verify with? @tabaktoni I am having a hard time getting one

https://sepolia.starkscan.co/tx/0x024f985090da17ccc9734efd2f516262fce50b37c702bf1e9d2a149796c05cfd
I am using this transaction to test

src/channel/rpc_0_7.ts
```diff
@@ -168,6 +169,22 @@ export class RpcChannel {
...
+      useEncoded(type),
+      version,
+      contract_address,
+      entry_point_selector,
+      hashedCallData,
+      '0x0',
+      nonce,
+    ]);
+  }
+
```
PhilippeR26: Are you sure this is algorithm?
With this transaction Hash : `https://sepolia.voyager.online/tx/0x28dfc05eb4f261b37ddad451ff22f1d08d4e3c24dc646af0ec69fa20e096819#messages`
The result has to be `0x55b3f8b6e607fffd9b4d843dfe8f9b5c05822cd94fcad8797deb01d77805532a`

PhilippeR26 (changes requested):
I fear you are not using the right algorithm...

princeibs:
What do you think could be wrong? @PhilippeR26 
I have reviewed the code and couldn't find what's wrong

PhilippeR26:
According to the issue, you have to code in the Starknet.js repo something that is equivalent to this Solidity code :
```solidity
pragma solidity >=0.8.2 <0.9.0;
contract Hasher {
  function getL1ToL2MsgHash(
        address fromAddress,
        uint256 toAddress,
        uint256 selector,
        uint256[] calldata payload,
        uint256 nonce
    ) external view returns (bytes32) {
        return
            keccak256(
                abi.encodePacked(
                    uint256(uint160(fromAddress)),
                    toAddress,
                    nonce,
                    selector,
                    payload.length,
                    payload
                )
            );
    }
}
```
And what you have coded is completely different.
Which source document have you used to create your code?

princeibs:
image
@PhilippeR26 I used this pseudocode which is gotten from the [linked resource](https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/messaging-mechanism/#structure_and_hashing_l1-l2) to create the function

PR #1122, opened 2024-05-07
Remove all code for Goerli testnet

PhilippeR26:
## Motivation and Resolution
After removal of any Goerli Testnet documentation in PR #1059 , and after discussion with Alex, this PR is removing any code related to Goerli Testnet. 

## Usage related changes
Goerli has been removed of all enums of `constants.ts`

## Development related changes
Official Testnet for the test suite is now Sepolia Testnet. Many modifications made these last months in `__tests__` directory where made to work only in Devnet-rs, without care of Goerli or Sepolia Testnets. Hereunder the necessary modifications performed to be able to pass the test suite with Sepolia Testnet :

IS_TESTNET environment variable was not set properly. Now it's set as true if an existing transaction hash in Sepolia Testnet is correctly read.

As declaration of Cairo 0 class is no more authorized in Sepolia Testnet,  some tests have to be launched only in Devnet-rs. I have modified the tests in accordance.

A Cairo 1.0.0 contract was tested in Goerli Testnet. As this contract is not available in Sepolia Testnet, and as it do not accept to declare/deploy anymore a such contract, the related code as been removed of `cairo1.test`.

After a deployment of contract, Sepolia Testnet do not accept to perform immediately the test `estimate message fee Cairo 1` in `rpcProvider.test`. We have to wait the validation of the pending block.
For EthSigner tests, the V2 & V3 tests are now only on Devnet-rs. The estimate fees are totally different and needs different correction values, depending of the network ; so, tuned to operate for Devnet-rs.

StarknetID tests adapted to Sepolia Testnet. Fees where set manually to 1 ETH for each transaction!!! Was a little expensive in Sepolia Testnet...

In Account.test, the sign EIP712 is not working properly for all accounts, because PR #989 is not yet implemented. So, moved temporarily to Devnet-rs only.
 
Also one small change in `cairo1v2.test` : a `declare` is replaced by a `declareIfNot` as this contract is already declared in Sepolia Testnet.

Test suite in Devnet-rs :
Capture d’écran du 2024-05-07 14-29-40

Test suite in Sepolia Testnet :
Capture d’écran du 2024-05-07 13-33-06
## Checklist:

- [x] Performed a self-review of the code
- [x] Rebased to the last commit of the target branch (or merged it into my branch)
- [x] Linked the issues which this PR resolves
- [x] Updated the tests
- [x] All tests are passing

PR #1123, opened 2024-05-08
feat: provider.getL1MessageHash

PhilippeR26:
## Motivation and Resolution
Solve issue #702.
Alex has asked to implement the calculation of a L1 message hash, from a Layer 2 transaction hash.
Here an example in Sepolia Testnet :
l1H

It's easy to perform with the Ethers library :
```typescript
const l1H = ethersUtils.solidityKeccak256(types, params);
```
This PR is implementing this calculation, without using the fat Ethers library.

## Usage related changes
Users will have access to this method : 
```typescript
const result = provider.getL1MessageHash('0x28dfc05eb4f261b37ddad451ff22f1d08d4e3c24dc646af0ec69fa20e096819');
```
If the L2 txH don't exists or isn't a `L1_Handler`, the function fails.

- The input data for the calculation is the response of `provider.getTransactionByHash(l2TxHash)`
- It was necessary to code a specific encoding of data, defined in
```typescript
ethersUtils.solidityPack(types, params);
```
- The hash is Keccak256. It's different to the Starknet Keccak, that is coded on 250 bits. This hash is available in the `@noble/hashes` library (already used by `@scure/starknet` lib).
- The test works only in Sepolia Testnet. To work properly, the fix included in PR #1122 has to be merged.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


ivpavici (commented):
lgtm but would leave to @tabaktoni if he has some suggestions.
I was mostly considering if it's important to clarify that this method is not part of the official RPC spec...

PhilippeR26:
Same as `provider.getL1GasPrice()` : not in any spec.

tabaktoni (approved):
looks delicious, lgtm

ivpavici (approved):

PR #1128, opened 2024-05-15
Verify message fix

tabaktoni:
## Motivation and Resolution
Based on PR https://github.com/starknet-io/starknet.js/pull/989 but removing pollution and containing hotfix in a single implementation

## Usage related changes
Added optional params in desc
```ts
  /**
   * Verify a signature of a JSON object
   *
   * @param typedData - JSON object to be verified
   * @param signature - signature of the JSON object
   * @param signatureVerificationFunctionName - optional account contract verification function name override
   * @param signatureVerificationResponse - optional response override { okResponse: string[]; nokResponse: string[]; error: string[] }
   * @returns true if the signature is valid, false otherwise
   * @throws {Error} if the JSON object is not a valid JSON or the signature is not a valid signature
   */
  public abstract verifyMessage(typedData: TypedData, signature: Signature): Promise<boolean>;
  /**
   * Verify a signature of a given hash
   * @warning This method is not recommended, use verifyMessage instead
   *
   * @param hash - hash to be verified
   * @param signature - signature of the hash
   * @param signatureVerificationFunctionName - optional account contract verification function name override
   * @param signatureVerificationResponse - optional response override { okResponse: string[]; nokResponse: string[]; error: string[] }
   * @returns true if the signature is valid, false otherwise
   * @throws {Error} if the signature is not a valid signature
   */
  public abstract verifyMessageHash(hash: BigNumberish, signature: Signature): Promise<boolean>;
```

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


src/account/default.ts
```diff
@@ -539,38 +539,90 @@ export class Account extends Provider implements AccountInterface {
...
+        // Response OK Signature
+        // Empty okResponse assume all non-nok responses are valid signatures
+        // OpenZeppelin 0.7.0 to 0.9.0, ArgentX 0.3.0 to 0.3.1 & Braavos Cairo 0.0.11 to 1.0.0 valid signature
+        if (
+          knownSignatureResponse.okResponse.length === 0 ||
+          knownSignatureResponse.okResponse.includes(resp[0].toString())
+        ) {
+          return true;
+        }
         return false;
```
tabaktoni, line 595: The only questionable is this one, i think it should be thrown here if not in nok nor ok set.

PhilippeR26 (approved):
To be honest, I didn't understood exactly what you made.
But if it works, it works! That the only thing important.

PR #1132, opened 2024-05-23
docs: utils/num docs

dsperac:
## Motivation and Resolution

...

...

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


src/utils/num.ts
```diff
@@ -206,27 +287,52 @@ export function getHexStringArray(value: Array<string>) {
  * // result2 = "0"
  * ```
  */
-export const toCairoBool = (value: boolean): string => (+value).toString();
+export function toCairoBool(value: boolean): string {
```
tabaktoni, line 290: Why replace arrow function ?

tabaktoni (changes requested):

src/utils/num.ts
```diff
@@ -28,21 +29,48 @@ export function isHex(hex: string): boolean {
...
+ * ```
  */
 export function toBigInt(value: BigNumberish): bigint {
   return BigInt(value);
 }
 
 /**
  * Test if value is bigint
+ *
+ * @param value value to test
```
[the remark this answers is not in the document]
dsperac: the first 'value' is the name of the param, and the second is from the description, it makes more sense with highlighting or when viewing the docs
image
image

src/utils/num.ts
dsperac: Consistency, everything else was in export function format. If this one was that way because of the size of the func, there are other 1-liner funcs in the file that weren't arrow.

tabaktoni (approved):

src/utils/num.ts
```diff
@@ -122,79 +161,121 @@ export function assertInRange(
...
+ * @returns converted hex-string
+ * @throws str needs to be a number string in hex or whole number format
  * @example
  * ```typescript
  * const result = getHexString("123");
  * // result = "0x7b"
  *
  * const result2 = getHexString("Hello");
- * // Throws Error: Hello need to be hex-string or whole-number-string
+ * // throws Error: Hello need to be hex-string or whole-number-string
```
ivpavici, line 247:
```suggestion
 * // throws Error: Hello needs to be a hex-string or whole-number-string
```

here as well or?

Issue #1090, opened 2024-04-19
Improve JSDocs by adding examples of usages


ivpavici:
Inspired by Viem docs, example:

https://viem.sh/docs/accounts/privateKey#usage

This is a big endeavor so it can be completed in multiple smaller PR-s (and more users can take it in parallel!)

IMPORTANT: add `@example` only to **exported** functions 

TODO utils:
- [ ] src/utils/calldata/validate 
- [ ] src/utils/calldata/formatter
- [ ] src/utils/calldata/propertyOrder
- [ ] src/utils/calldata/requestParser
- [ ] src/utils/calldata/responseParser
- [ ] src/utils/calldata/tuple
- [ ] src/utils/calldata/parser/index.ts
- [ ] src/utils/events/index.ts
- [ ] src/utils/hash/classHash.ts
- [ ] src/utils/hash/transactionHash/v3.ts

ivpavici:
Users please pick one part of the docs which you want to work on and write here for everyone else to know!

elielnfinic:
Hey @ivpavici, could you please specify on which elements examples should be added ?

RolandoDrRobot:
Hello @ivpavici I would love to be part of this effort! is this the docs you guys want to enhance? https://www.starknetjs.com/docs/guides/create_account I can take the accounts part, thanks (I am in OD Hack telegram, ready to start on Monday 🫡)

Ayoazeez26:
I would love to take the [signature](https://www.starknetjs.com/docs/guides/signature) part of the docs

Jemiiah:
if this work isn't assigned to anybody, can i get it, @ivpavici ?

ivpavici:
hi all!

To explain a bit more, we are not talking about rewriting Guides (like https://www.starknetjs.com/docs/guides/create_account) but adding examples to JSdocs in the code, how they can be used, for example on utils:

![image](https://github.com/starknet-io/starknet.js/assets/2848732/3282dffb-ce8d-4b14-b54d-315bc7cdf943)

ivpavici:
@Jemiiah no need to assign since multiple people can work on it... pick a part of the code (please write here which) and send a PR with this issue linked

Ugo-X:
Hello @ivpavici would like to handle the Estimate fees segment, if there's no need for an assignment won't there be a clash at some point?

ivpavici:
@Ugo-X I will consider PR-s from contributors who announce the parts up front.
[...]

PR #1133, opened 2024-05-23
docs: jsdoc adds

PhilippeR26:
## Motivation and Resolution
Participates to #1090
As PRs 1112, 1114, 1115 & 1118 are stalled, this PR is finalizing all the job initiated in these PRs

## Usage related changes
- To solve PR 114, added JSDOC for utils/classHash and utils/provider.
  Also few warnings from Typedoc are solved.
  Also some cosmetics in guides.

N/A

## Checklist:

- [x] All tests are passing

### DIFF
--- a/src/utils/hash/classHash.ts
+++ b/src/utils/hash/classHash.ts
@@ -33,8 +33,16 @@ export function computePoseidonHash(a: BigNumberish, b: BigNumberish): string {
 }
 
 /**
- * Compute pedersen hash from data
- * @returns format: hex-string - pedersen hash
+ * Compute Pedersen hash from data
+ *
+ * @param {BigNumberish[]} data Array of data to compute Pedersen hash on
+ * @returns {string} hex-string of Pedersen hash
+ *
+ * @example
+ * ```typescript
+ * const result = hash.computeHashOnElements(['0xabc', '0x123', '0xabc123'])
+ * // result = 0x148141e8f7db29d005a0187669a56f0790d7e8c2c5b2d780e4d8b9e436a5521
+ * ```
  */
 export function computeHashOnElements(data: BigNumberish[]): string {
   return [...data, data.length]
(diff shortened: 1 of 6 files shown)


src/utils/hash/classHash.ts
```diff
@@ -110,7 +133,9 @@ export function formatSpaces(json: string) {
 
 /**
  * Compute hinted class hash for legacy compiled contract (Cairo 0)
- * @returns format: hex-string
+ * @param {LegacyCompiledContract} compiledContract
+ * @returns {string} hex-string
+ * No example, as this function is not exported to user
  */
 export default function computeHintedClassHash(compiledContract: LegacyCompiledContract) {
```
dsperac, line 140: Can we remove default from here? Would be more clean with explicit imports, this func isn't immediately obvious to be the default one, especially since we have other ones like computeLegacyContractClassHash

src/utils/provider.ts
```diff
@@ -143,11 +224,26 @@ export class Block {
...
   // TODO: fix any
+  /**
+   * Return the identifier as a string
+   * @example
+   * ```typescript
+   * const result = new provider.Block(123456n).queryIdentifier;
+   * // result = "blockHash=0x1e240"
+   * ```
+   */
   get queryIdentifier(): any {
```
dsperac, line 247: Do we need @returns on these?

PhilippeR26: Not really a documented case...
@type is rarely used with Typescript. Mainly for JS when unclear type response ; not applicable for this case.
Seems that most of projects are just implementing a comment, and sometime an example, but nothing else: no @returns.
Well, to be more royalist than the king, I will add @returns...

src/utils/hash/classHash.ts
PhilippeR26: Traced in #1135

dsperac (approved):

ivpavici (approved):

Issue #1134, opened 2024-05-23
Bug parsing events when nested events involved


NikitaMishin:
**Location:** https://github.com/starknet-io/starknet.js/blob/66a5c0341eccfef0dcdf1312c15627b7d4f6b675/src/utils/events/index.ts#L52C21-L52C28

Encounter issue with parsing events emitted by components:

**Sample of code** (just take any events that emitted in nested components):
```
    const eventsC = provider.getEvents(....)
    const abiEvents = STARKNETJS.events.getAbiEvents(abi);
    const abiStructs =  STARKNETJS.CallData.getAbiStruct(abi);
    const abiEnums = STARKNETJS.CallData.getAbiEnum(abi);
    let parsedEvents = STARKNETJS.events.parseEvents(eventsC.events, abiEvents, abiStructs, abiEnums)
    console.log(parsedEvents)

```
produces 0 parsed events

**Reason:** AbiEvents not handle correctly case for nested events when they defined by extra key that associated with component. In my particular case second key was the correct one to do lookup to abiEvents
**Potential solution**: 

```
let abiEvent: EventAbi | undefined;

// Loop through all keys and find the first matching event
for (const key of recEvent.keys) {
    abiEvent = abiEvents[key];
    if (abiEvent) {
        break; // Exit the loop once a match is found
    }
}
```
iterate over all keys/or first N instead of just first one. I presume there should be no collisions

MartinXV111:
can i hop on this?

martinvibes:
hi @NikitaMishin can i hop on this

NikitaMishin:
also  if several components have same name for event  getAbiEvents would just shallow rewrite it,
 not taking into account component hash

od-hunter:
Hello @NikitaMishin please assign me to this project. This is my first time contributing to Starknet ecosystem and would love to proof my acquired skills, thank you.

od-hunter:
Hello @NikitaMishin im ready for work assign me please.

od-hunter:
@NikitaMishin still waiting for your go ahead.

Timilehin-bello:
@NikitaMishin please can you kindly assign this task to me. I'll start working on it and revert

PhilippeR26:
Hello @NikitaMishin ,
I would like to repeat completely your problem. I need the network used, the parameters in getEvents, the address of the contract of these events.

PhilippeR26:
What's the content of `restKeys`?

[...]

PR #1140, opened 2024-05-31
BUG: Nested events not handled

PhilippeR26:
## Motivation and Resolution
Solves #1134 .
Cairo is now using a new concept : the components. Events can be nested in these components. To differentiate them, an event name can now be defined with more than one hash number ; so the number of felts to represent the hashed name in `key[]` can be greater than one (unless usage of the Cairo `#[flat]` attribute). See https://book.cairo-lang.org/ch14-03-contract-events.html?highlight=events#defining-events .
As the current code is not taking into account this case of nested events, the parse of a such event is returning an empty array.
This PR is solving this case.

- The abi includes a complete tree of the events of all components ; `getAbiEvents()` creates recursively a tree of events hashs. It will be used for the event parse.
- `events.parseEvents()` is reading the event name hash in `key[]` ; if necessary, several numbers are read.
- This PR is also the opportunity to fix a problem in the documentation site : the invitation link for Discord is valid only 7 days. So, it has been replaced by a link to the space in Starknet Discord.
- tests : A big abi from an existing project deployed in Sepolia is used, with an add at the end from Openzeppelin contracts including a `Flat` Event. It includes normal events and nested events, with 2 events with the same name (`Deposit`). So there is no use of Devnet-rs ; inputs and outputs have been recovered from Sepolia real examples.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


src/types/lib/contract/abi.ts
```diff
@@ -48,7 +48,9 @@ export type EnumAbi = {
 
 export type AbiEvents = { [hash: string]: EventAbi };
 
-export type EventAbi = Cairo1Event | LegacyEvent;
+export type EventAbi = Cairo1Events | LegacyEvent;
+
+export type Cairo1Events = Cairo1Event | AbiEvents;
```
tabaktoni, line 53: PR is LGTM, just some polishing.

- Rename all Cairo1 to Cairo. (we can consider old Cairo Cairo0 or Legacy and all new one's just Cairo)

tabaktoni: One more thing:
```ts
export type AbiEvents = { [hash: string]: EventAbi }; // Expect from name this are definitions of events in ABI
export type EventAbi = Cairo1Events | LegacyEvent; // Except from name this is definition of one event in abi with swaped words, but it is multiple onchain events for Cairo1 and singular onchain event for Cairo0
export type Cairo1Events = Cairo1Event | AbiEvents; // Then here we have onchain actual events that are > Cairo1 singular event or multiple ABI definitions of events ?


export type AbiStructs = { [name: string]: StructAbi }; // This name swaping for singular is confusing as hell 
export type AbiEnums = { [name: string]: EnumAbi }; // This name swaping for singular is confusing as hell

```

Can we resolve this namings its super confusingn hard to figure out what is what.
We already have a breaking changes so let batch tham

tabaktoni (changes requested):

src/types/lib/contract/abi.ts
tabaktoni: One more thing
```ts
export type AbiEvents = { [hash: string]: EventAbi }; // Expect from name this are definitions of events in ABI, This name swaping for singular is confusing as hell 
export type EventAbi = Cairo1Events | LegacyEvent; // Except from name this is definition of one event in abi with swaped words, but it is multiple onchain events for Cairo1 and singular onchain event for Cairo0
export type Cairo1Events = Cairo1Event | AbiEvents; // Then here we have onchain actual events that are > Cairo1 singular event or multiple ABI definitions of events ?
export type AbiStructs = { [name: string]: StructAbi }; // This name swaping for singular is confusing as hell 
export type AbiEnums = { [name: string]: EnumAbi }; // This name swaping for singular is confusing as hell
```

All this is super confusing let batch breaking changes and resolve this also.

src/utils/events/index.ts
```diff
@@ -3,63 +3,173 @@ import {
...
+        const abiEntryMod = { ...abiEntry };
+        abiEntryMod.name = entryName;
+        return {
+          ...acc,
+          [addHexPrefix(starkCurve.keccak(utf8ToArray(entryName)).toString(16))]: abiEntryMod,
+        };
+      }, {});
+  } else {
+    // Cairo 1 abi
+    const abiEventsStructs = abi.filter((obj) => obj.type === 'event' && obj.kind === 'struct');
```
ivpavici, line 61: ` obj.type === 'event'`

could be extracted as a helper, similar to isString or isObject logic ?

src/types/lib/contract/abi.ts
tabaktoni: Also, this ABI Spec stuff should be used from API/spec instead of redefining locally

src/utils/events/index.ts
```diff
@@ -3,63 +3,173 @@ import {
...
-      const entryName = abiEntry.name.slice(abiEntry.name.lastIndexOf(':') + 1);
-      const abiEntryMod = { ...abiEntry };
-      abiEntryMod.name = entryName;
-      return {
-        ...acc,
-        [addHexPrefix(starkCurve.keccak(utf8ToArray(entryName)).toString(16))]: abiEntryMod,
-      };
-    }, {});
+  let finalResult: AbiEvents;
+  if (!isCairo1Abi(abi)) {
```
ivpavici, line 46: are we testing now both of these cases in our tests, Cairo 0 and Cairo 1? 
We could even extract this logic in `if` and `else` Cairo 1 abi logic into 2 functions and do independent test for both of those new functions. Also, this function would become smaller if we do it this way

src/utils/events/index.ts
```diff
@@ -3,63 +3,173 @@ import {
...
+ * @param {any} item the tested item
+ * @returns {boolean}
+ * @example
+ * ```typescript
+ * const result = events.isObject({event: "pending"});
+ * // result = true
+ * ```
+ */
+export function isObject(item: any): boolean {
+  return item && typeof item === 'object' && !Array.isArray(item);
```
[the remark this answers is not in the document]
PhilippeR26: I had the same toughts, with the same conclusion. So I leave it there.

src/types/lib/contract/abi.ts
PhilippeR26: AbiStructs & AbiEnums are coded this way for a long time.
I can change all this, but I am not sure at all to have understood exactly what you are expecting.
I will try something, but you will have probably to finalize, because these things are so personal, and I have not access to your brain (probably a good thing for everybody...).

PhilippeR26:
```
export type Cairo1Events = Cairo1Event | AbiEvents
```

A Cairo 1  Abi includes 2 types of events definition :
- The detailed definition of each event. includes`members` key.
- The list of the events included in a component or in the main code. Descriptions can be nested. includes `variants` key.

PhilippeR26:
I think I will first commit something to answer to @tabaktoni, and when it will be accepted, I will focus on @ivpavici  requests.

src/utils/events/index.ts
PhilippeR26: Cairo 0 events have never been tested in the test suite. I have tested locally (with success) with an old customized Cairo 0 account that emits Cairo 0 events. If you want, I can add it in the test suite, but it will take some time (declare, deploy, invoke). What do you prefer?

src/utils/events/index.ts
PhilippeR26: OK. Where to locate this function?

src/utils/events/index.ts
tabaktoni: As cairo0 is in support mode for existing deployed contracts as so we can switch also to mode kinda supporting it but should focus on new cairo only

ivpavici: ok then, no need for a test here!

src/utils/events/index.ts
ivpavici: I thought of maybe the same file, but this is nitpicking, since we also have `obj.kind === 'struct'` etc... so ok, let's leave it like this unless in the future there is a need to extract to other methods :)

PhilippeR26:
@tabaktoni 
I have added the types from rpc 0.6 (as rpc 0.7 event types can't be imported with current libs used by SNJS) :
abi.ts :
```typescript
import type { ENUM_EVENT, EVENT_FIELD, STRUCT_EVENT } from '../../api/rpcspec_0_6/contract';
```
I will move to 0.7 when exports available.

NOTA : I have now a huge quantity of commits listed in this PR, that appeared after merge to next-version branch. I don't know how to solve this. If it's a problem, if somebody can solve this ...

tabaktoni (approved):
lgtm

penovicp (approved):

PR #1145, opened 2024-06-05
fix: export abi types

tabaktoni:
## Motivation and Resolution

...

...

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


PhilippeR26 (commented):
Seems that rpc spec 0.6 is identical to rpc spec 0.7 about contract types.
In this case, for any src/utils code , is it better to use the internal rpcspec_06, or to use the rpcspec 0.7 of the types-js library?

tabaktoni:
> Seems that rpc spec 0.6 is identical to rpc spec 0.7 about contract types. In this case, for any src/utils code , is it better to use the internal rpcspec_06, or to use the rpcspec 0.7 of the types-js library?

Yea, abi didn't change, I would recommend using rpcspec 0.7  as it is the latest one, and 0.6 will be deprecated at some point.

PhilippeR26:
So, you should also add exports to these functions in the types-js lib.

tabaktoni:
> So, you should also add exports to these functions in the types-js lib.

Yes it is added

Issue #1144, opened 2024-06-04
Cairo Zeroable type not supported


PhilippeR26:
**Describe the bug**
Try to interact with a Cairo function that use the type `NonZero` generates an Error.

**To Reproduce**
The contract `0x07be990e59999c2e4210d2e42f99c5701bb641da25ccccad14b206c2f0acb2c2` deployed in Sepolia Testnet has a `pause` function using `NonZero` :
```json
{
    "name": "pause",
    "type": "function",
    "inputs": [
      {
        "name": "some_param",
        "type": "core::zeroable::NonZero::<core::integer::u128>"
      }
    ],
    "outputs": [],
    "state_mutability": "external"
  }
```
code :
```typescript
const testAddress= "0x07be990e59999c2e4210d2e42f99c5701bb641da25ccccad14b206c2f0acb2c2";
const testContract = await myProvider.getClassAt(testAddress);
const myContract=new Contract( testContract.abi,testAddress,account0);
const res=await myContract.invoke("pause",[200]);
```
Error :
```
Error: Validate Unhandled: argument some_param, type core::zeroable::NonZero::<core::integer::u128>, value 200
    at /D/starknetFork/starknet.js/src/utils/calldata/validate.ts:320:15
```

**Expected behavior**
Full compatibility with `core::zerable`.

**Screenshots**
N/A
**Desktop (please complete the following information):**
- Browser & version [e.g. chrome, safari, webworker] N/A
- Node version [e.g. 16.0.1]
- Starknet.js version v6.9.0
- Network [devnet, testnet]

**Additional context**
I will investigate is there are other types in `core::zeroable` that fails also.

Jemiiah:
_i_

PR #1150, opened 2024-06-07
feat: handling of Cairo Zeroable type

PhilippeR26:
## Motivation and Resolution
Solves issue #1144
Cairo is using now the `NonZero` type in its Abi.
I didn't found any documentation on this type. So I read the Cairo code, and used this message :
Telegram : https://t.me/sncorestars/11902/45433
```
   Author : Ori Ziv (08/apr/2024)
   "NonZero is only supported for purely numeric types (u*, i* and felt252) and EcPoint."
```
As EcPoint do not includes trait Serde, it can't be seen in an ABI.
 u512 is not compatible.
i* are not currently handled by Starknet.js (and core::zeroable::NonZero::<i*> seems not to compile in Cairo 2.6.3).
So, are authorized here : u8, u16, u32, u64, u128, u256 and felt252.

## Usage related changes
Contracts with an ABI including this kind of things is now handled, both for requests and for responses :
`core::zeroable::NonZero::<core::integer::u256>`

- added an helper : `isTypeNonZero`
- This new type is not something _rigid_ as uint256. It more similar to `array`, and is just listing real types. That's why I do not included the code in the CairoDataTypes directory.
- added handling in parsers.
- tests : As `cairo1v2.test.ts `is now huge, I have create a new `cairov240onwards.test.ts`, moved all recent tests due to Cairo evolution, and added new tests for NonZero. A new test contract has been created to test various cases.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


tabaktoni (approved):
LGTM, 
This seems like a flag on a numeric types.
For now a valid solution, later we can dd how to add it on CairoType* class as non-zero flag.

PR #1154, opened 2024-06-10
Fix: public responseParser

tabaktoni:
## Motivation and Resolution

> YoavG
> RpcProvider's CTOR expects both channel and responseParser if channel is provided in RPCProviderOptions
> Since I wanted to create a provider with a custom RpcChannel I had to provide the response parser as well. Since I didn't want to touch the response parser I expected that the default response parser will be available via new RPCResponseParser but it isn't since it is not exported.
> My workaround was to instantiate the provider as usual and then override provider.channel = customChannel (edited) 

set responseParser to be public should add an additional option to instantiate from default and change any/both channels or parsers.

mentioned above

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


ivpavici (approved):

__tests__/rpcProvider.test.ts
```diff
@@ -50,6 +50,17 @@ describeIfRpc('RPCProvider', () => {
...
 
+  test('instantiate from rpcProvider', () => {
+    // instantiate rpc provider template
+    const newInsRPCProvider = new RpcProvider();
+    // like* change channel and parser
+    newInsRPCProvider.channel = rpcProvider.channel;
+    newInsRPCProvider.responseParser = rpcProvider.responseParser;
+    // instantiate from modified instance
+    const FinalInsRPCProvider = new RpcProvider(newInsRPCProvider);
+    expect(FinalInsRPCProvider).toBeInstanceOf(RpcProvider);
```
penovicp, line 61: I would suggest changing the test to something like this so both branches of the parser conditional are covered:

```suggestion
    const newInsRPCProvider = new RpcProvider();
    
    let FinalInsRPCProvider = new RpcProvider(newInsRPCProvider);
    expect(FinalInsRPCProvider.channel).toBe(newInsRPCProvider.channel);
    expect(FinalInsRPCProvider.responseParser).toBe(newInsRPCProvider.responseParser);
    
    delete (newInsRPCProvider as any).responseParser;
    FinalInsRPCProvider = new RpcProvider(newInsRPCProvider);
    expect(FinalInsRPCProvider.channel).toBe(newInsRPCProvider.channel);
    expect(FinalInsRPCProvider.responseParser).toBeInstanceOf(RPCResponseParser);
```

src/provider/rpc.ts
```diff
@@ -29,27 +30,30 @@ import {
...
+import { hexToBytes, toHex } from '../utils/num';
 import { wait } from '../utils/provider';
+import { RPCResponseParser } from '../utils/responseParser/rpc';
+import { GetTransactionReceiptResponse, ReceiptTx } from '../utils/transactionReceipt';
 import { LibraryError } from './errors';
 import { ProviderInterface } from './interface';
 
 export class RpcProvider implements ProviderInterface {
-  private responseParser: RPCResponseParser;
+  public responseParser: RPCResponseParser;
```
penovicp, line 46: I would suggest making this `private` again or adding an export for the `RPCResponseParser` class somewhere so it is exposed by the library and can be picked up by the automated documentation generation.

penovicp (approved):
Left two suggestions, in general lgtm

src/provider/rpc.ts
tabaktoni: c2887ed250c0f7331986738f6d3ef563eb0ea0ce

Issue #1014, opened 2024-03-14
Add test and support for negative Signed integers smaller than 252 bits


tabaktoni:
**Is your feature request related to a problem? Please describe.**
https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/serialization_of_Cairo_types/#data_types_of_252_bits_or_less

**Describe the solution you'd like**
Test current behavior and support required field el calc
i8, i16, i32, i64, and i128

Negativ value should be 2^{251} + 17*2^{192} + 1 + (negative value)
 ex -5 should serialise to 2^{251} + 17*2^{192} + 1 - 5

HumbertoBrenes-Labot:
Hi @tabaktoni I can handle this one!

AryanGodara:
@tabaktoni Can I work on this one after completing #1018 ? :)

tabaktoni:
Yes ofc, maybe the task will need to be expanded to multiple sub tasks we will need to check how to implement all Cairo int types in the manner of CairoInt. 
But the initial step (Test current behavior ) can be immediately tackled.

AryanGodara:
@ivpavici I'm still working on the other issue. So please don't unassign there 😅 
I had left a doubt regarding testing earlier. And I'd like to complete it during this week

ivpavici:
ok when you are done there we can assign you here if no one else wants to pick it up

lucilapastore:
Hey! I am from Argentina. I am at the mu. I want to contribute and make this my first issue. Regards!

ivpavici:
@lucilapastore good luck!

fmmesen:
Could I work on this? @ivpavici

ivpavici:
good luck!

fmmesen:
Hey @ivpavici , sorry for the delay! I just have a doubt

Doing the serialization manually:

2^{251} + 172^{192} + 1 - 5

I'll receive:

3618502788666131113263695016908177884250476444008934042335404944711319814140
result that will be returned as a felt252

So the idea is to create a unit test that will expect(result).ToBe(361850278866613111......) for i8, i16, i32, i64 and i128?

tabaktoni:
Implementation needs to be the similar  to  [CairoUint256](https://github.com/starknet-io/starknet.js/blob/66a5c0341eccfef0dcdf1312c15627b7d4f6b675/src/utils/cairoDataTypes/uint256.ts#L18)

In addition to this you need to define and store: 

[...]

PR #1177, opened 2024-07-05
Add test and support for negative Signed integers smaller than 252 bits

fmmesen:
feat: added new data types
test: added new tests for data types
#1014  

## Motivation and Resolution

Created new Cairo Data Types integer classes and their unit tests for every range of bits below 252, similar to the `uint256.ts` file.

- Added cairoInt, int8, int16, int32, int64, int128 clases.
- Added tests for every new data type.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


fmmesen:
Hello @tabaktoni @ivpavici !
New feature is done and ready to be reviewed.

penovicp (changes requested):
The purpose of the linked issue isn't to just create the corresponding utility classes but to actually use them for request/response handling.

Also, `CairoInt` looks like it should be an abstract class and the type implementation classes look like they could have improved code reuse.

fmmesen:
Hello @penovicp, I have a doubt about where exactly I should use those classes for request/response handling. Could you provide some guidance?

penovicp:
This [comment](https://github.com/starknet-io/starknet.js/issues/1014#issuecomment-2176111160) describes what the required behaviour should be.

The existing data type utility classes such as [CairoUint256](https://github.com/starknet-io/starknet.js/blob/v6.11.0/src/utils/cairoDataTypes/uint256.ts#L18) and [CairoUint512](https://github.com/starknet-io/starknet.js/blob/v6.11.0/src/utils/cairoDataTypes/uint512.ts#L15) can be checked to see how and where the signed integer classes should be used.

fmmesen:
@penovicp Already used the utility classes, please let me know if you see that I'm missing something

Issue #1168, opened 2024-07-02
Adding Ledger Signer support


amanusk:
**Is your feature request related to a problem? Please describe.**
Ledger now has a Starknet App, supporting singing transactions with Stark Curve on Ledger. Any existing Starknet account can now be used with Ledger as a signer. 

**Describe the solution you'd like**
Add a custom `LedgerSigner` to starknet.js

This can be part of a broader generic signer implementation, but provided as part of the SDK for simpler integration 

**Additional context**
Reference implementation on starknet-rs:

https://github.com/xJonathanLEI/starknet-rs/blob/db1fa598232f0698d942cc974f481b5d888ac080/starknet-signers/src/ledger.rs

ivpavici:
relevant:
https://github.com/starknet-io/starknet.js/pull/748

PhilippeR26:
@amanusk Is it a JS/TS equivalence to the rust library `coins-ledger`?

PhilippeR26:
My (very) old Nano S is not compatible with the Starknet ledger APP.
So, I ordered a new Nano X, to test it.
![ledgerNanoX](https://github.com/starknet-io/starknet.js/assets/81040730/49ec600a-10e7-4ebc-b6dc-0d8d2d89209c)
Nice!

PhilippeR26:
I can communicate with it :
```typescript
TransportNodeHid {
  exchangeTimeout: 30000,
  unresponsiveTimeout: 15000,
  deviceModel: {
    id: 'nanoX',
    productName: 'Ledger Nano X',
    productIdMM: 64,
    legacyUsbProductId: 4,
    usbOnly: false,
    memorySize: 2097152,
    masks: [ 855638016 ],
    getBlockSize: [Function: getBlockSize],
    bluetoothSpec: [ [Object] ]
  },
  ```
  I have not to dig to 
  - recover the pubkey
  - sign a hash (as the ledger APP is not able to build a transaction)

PhilippeR26:
The job to perform is mainly to create code for these 2 red parts :
![LedgerArchitecture](https://github.com/starknet-io/starknet.js/assets/81040730/b5a5608b-5b09-417b-8752-ade05ccb221f)
for 3 methods :
- getVersion
- getPubKey
- sign

PhilippeR26:
It was difficult, but it works.
The documentation is full of errors or missing data : https://github.com/LedgerHQ/app-starknet/blob/develop/docs/apdu.md
[...]

PR #1185, opened 2024-07-24
Ledger Signer

PhilippeR26:
## Motivation and Resolution
Solves #1168 
LedgerTitle
As Starknet is now handled by the Ledger Nano S+ / X, we can create a signer that handle the private keys of this hardware wallet.

## Usage related changes
After installing the Starknet Ledger DAPP in your Ledger, you can use the Ledger as signer.
The way to proceed depends of the type of connection, and of the framework used (see the guide for more details).
For example, in a Node script, using a Ledger Nano + :
```typescript
import TransportNodeHid from '@ledgerhq/hw-transport-node-hid';
const myLedgerTransport: Transport = await TransportNodeHid.create();
const myLedgerSigner = new LedgerSigner(myLedgerTransport, 0);
const pubK = await myLedgerSigner.getPubKey();
const fullPubK = await myLedgerSigner.getFullPubKey();
// ...
// deploy here an account related to this public key
// ...
const ledgerAccount = new Account(myProvider, ledger0addr, myLedgerSigner);
```
Then you can use all usual methods of the Account class. When a signature is necessary, a message is displayed on the Ledger screen, and a manual validation is requested.
By this way, the private key of the account is never exposed.

The signature guide has been amended to include the Ledger Signer.

- A new class has been added in the signer directory.
- As the transport library to use depends of the framework, its creation has to be outside of Starknet.js (impossible to build using the native libs or the node libs).
- tests : of course automated tests are impossible. Just the new utils functions are tested. Nevertheless, the Nodes and Web USB connections have been tested in separated repos :
Nodes USB : https://github.com/PhilippeR26/starknet.js-workshop-typescript/blob/main/src/scripts/ledgerNano/5.testLedgerAccount.ts
Web USB : https://github.com/PhilippeR26/Starknet-Ledger-Wallet (the 3 types of signatures are tested)

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


tabaktoni (approved):
🥇 LFG
LGTM!

penovicp (approved):

Issue #1137, opened 2024-05-28
Problems to use RPCSPEC06 and RPCSPEC07 namespaces


PhilippeR26:
**Describe the bug**
2 problems encountered :
### 1. No Vscode help in RPC spacename
Reading the Starknet.js code, I understand that the paths to use
the enums located in src/types/api/rpcspec_0_6/nonspec.ts or in  spec 07  are :
```typescript
import {  types,  RPC } from "starknet";
const lastBlockTag=types.RPC.RPCSPEC06.EBlockTag.LATEST;
const lastBlockTag2=RPC.RPCSPEC06.EBlockTag.LATEST;
const lastBlockTag3=types.RPC.RPCSPEC07.EBlockTag.LATEST;
const lastBlockTag4=RPC.RPCSPEC07.EBlockTag.LATEST;
const lastBlockTag5=types.RPC.EBlockTag.LATEST;
const lastBlockTag6=RPC.EBlockTag.LATEST;
```
Long and complicated path.... but it works.
After typing `types.`, vsCode is helping to reach `RPC` :
![image](https://github.com/starknet-io/starknet.js/assets/81040730/1af9903a-d6d1-4f43-ada5-032d510f2974)
But after selecting `RPC`, there is no more help from vsCode :
![image](https://github.com/starknet-io/starknet.js/assets/81040730/98467a0a-3a03-445a-a7d8-f3dca3ca33a9)
There it becomes complicated. Only digging in a very nested source code can help you.

### 2. use of RPC types 
If you add to the previous code :
```typescript
const callOpt:types.RPC.RPCSPEC06.SPEC.SIERRA_ENTRY_POINT={selector: "0x123",function_idx:2};
```
The compilation fails :
```
TSError: ⨯ Unable to compile TypeScript:
src/tmp/26.testSepolia.ts:38:25 - error TS2694: Namespace 'index$2' has no exported member 'RPC'.

38     const callOpt:types.RPC.RPCSPEC06.SPEC.SIERRA_ENTRY_POINT={selector: "0x123",function_idx:2};
                           ~~~
```
So, how to use these types?

**To Reproduce**
See above

**Expected behavior**
Good user experience to use RPC enums.
Have a way to use types of RPC.

**Screenshots**
See above.

**Desktop (please complete the following information):**

-  Browser & version [e.g. chrome, safari, webworker] : any
- Node version [e.g. 16.0.1] ts-node": "^10.9.2
- Starknet.js version : v6.9.0
[...]

PR #1193, opened 2024-08-02
chore: revert companion pattern except constants

lukasaric:
## Motivation and Resolution

Use companion pattern instead because of the type safety except constants to avoid build issues.
Tested by linking the newly built version within [Starknet Wallet Account](https://github.com/PhilippeR26/Starknet-WalletAccount) (linked it locally, also tested packed variant of project).

Next steps for new PR’s:

- Fix/improve the whole build process so that `constants.ts` works properly without using enums
- Fix [namespaces issues](https://github.com/starknet-io/starknet.js/issues/1137) #1137 

## Usage related changes

n.a

 n.a

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


ivpavici:
@PhilippeR26 when you catch some time, could you test this solution please 🙏

PhilippeR26 (commented):
I tested the cases of #1137 .
Still the same problem.
I see zero improvement.

lukasaric:
> I tested the cases of #1137 . Still the same problem. I see zero improvement.

As I said in the description, this issue is not covered via this PR (that’s a next step for a new follow up PR). I’ll update description to be more clear

PhilippeR26:
> > I tested the cases of #1137 . Still the same problem. I see zero improvement.
> 
> As I said in the description, this issue is not covered via this PR (that’s a next step for a new follow up PR). I’ll update description to be more clear
🤔
So, what can be tested to verify this PR?

lukasaric:
> > > I tested the cases of #1137 . Still the same problem. I see zero improvement.
> > 
> > 
> > As I said in the description, this issue is not covered via this PR (that’s a next step for a new follow up PR). I’ll update description to be more clear
> > 🤔
> > So, what can be tested to verify this PR?

Build should work properly in projects that are using starknet. I tested with your starknet wallet account repo and it’s working fine, but I wanted to double check with the author :)

PhilippeR26:
Understood.
I have tested with :
- a node using contants : ✅
- my most advanced DAPP : the Airdrop demo https://github.com/PhilippeR26/Airdrop-for-Starknet : ✅

src/types/api/rpcspec_0_6/contract.ts
```diff
@@ -11,7 +11,7 @@ export type ABI = Array<
   FUNCTION | CONSTRUCTOR | L1_HANDLER | EVENT | STRUCT | ENUM | INTERFACE | IMPL
 >;
 
-export type FUNCTION = {
+type FUNCTION = {
```
penovicp, line 14: I believe removing exports for types makes the affected types inaccessible for the API documentation generation.

I doubt anyone actually uses/reads it down to this level and the existing docs probably have more dead-end references, but I think having the reference available is better than not having it.

penovicp (commented):
Mentioned a possible documentation issue. In general, double-checking how the documentation is affected might be worthwhile.

Aside from that lgtm

src/types/api/rpcspec_0_6/contract.ts
lukasaric: Hmm, these types exports were never accessible because in the `index` file they were never re-exported (`Methods` and `ABI` were exported only 👉  [link](https://github.com/starknet-io/starknet.js/pull/1193/files#diff-149d2fb59cd50a587db52e2e50bc00bf56fa5d3eed47c05b754aa3fc636e91edL5)).
And ofc these exports were never used within the pkg itself
cc @penovicp

penovicp: Good point, these RPC 0.6 types are already affected with the issue I mentioned because of the missing re-export. Just to show what I meant, this is the [ABI](https://starknetjs.com/docs/API/namespaces/types.RPC.RPCSPEC06#abi) doc that references them, and this is the [ABI](https://starknetjs.com/docs/API/namespaces/types.RPC.RPCSPEC07.API#abi) doc for 0.7 where the references exist.

Let's leave them unexported, 0.6 shouldn't be around much longer anyway.

penovicp (approved):

PR #1211, opened 2024-08-30
feat: isClassDeclared  createBulkInvocations, fix cairo0 test

### LINKED ISSUE #1209: Failing tests on devnet v0.2.0-rc.x
Due to changes in the blockifier ("A change was made so that re-declaring cairo0 classes also fails with a "already declared" error") integrated into devnet v0.2.0-rc.x  we must go through our tests and apply the logic to circumvent this error 

Job with failing tests:
https://github.com/starknet-io/starknet.js/actions/runs/10507187425/job/29108393106

tests:
-  deploy and test Wallet › simulate transaction - single transaction S0.11.2 › declare tests only on devnet > simulate DECLARE - Cairo 0 Contract
- deploy and test Wallet › Estimate fee bulk & estimate fee › declare tests only on devnet › declare, deploy & multi invoke functions

Error:
` 40: Contract error: {"revert_error":"Class with hash 0x054328a1075b8820eb43caf0caa233923148c983742402dcfc38541dd843d01a is already declared."}`

tabaktoni:
## Motivation and Resolution
Fix #1209

- Implement TransactionType Fix from #1193 to be able to use TS Invocations properly

## Usage related changes

New
 Test if class is already declared from `ContractClassIdentifier`
```ts
provider.isClassDeclared(contractClassIdentifier: ContractClassIdentifier, blockIdentifier?: BlockIdentifier)
```

Build _bulk_ invocations with auto-detect declared class and declare order sort
~~provider.createBulkInvocations(invocations: Invocations)~~
```ts
provider.prepareInvocations(invocations: Invocations)
```

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/provider/rpc.ts
+++ b/src/provider/rpc.ts
@@ -1,4 +1,5 @@
 import type { SPEC } from 'starknet-types-07';
+
 import { RPC06, RPC07, RpcChannel } from '../channel';
 import {
   AccountInvocations,
@@ -8,12 +9,14 @@ import {
   BlockTag,
   Call,
   ContractClassResponse,
+  ContractIdentifier,
   ContractVersion,
   DeclareContractTransaction,
   DeployAccountContractTransaction,
   GetBlockResponse,
   GetTxReceiptResponseWithoutHelper,
   Invocation,
+  Invocations,
   InvocationsDetailsWithNonce,
   PendingBlock,
   PendingStateUpdate,
@@ -25,24 +28,24 @@ import {
   getContractVersionOptions,
   getEstimateFeeBulkOptions,
   getSimulateTransactionOptions,
-  waitForTransactionOptions,
   type Signature,
   type TypedData,
+  waitForTransactionOptions,
 } from '../types';
 import type { TransactionWithHash } from '../types/provider/spec';
 import assert from '../utils/assert';
+import { CallData } from '../utils/calldata';
 import { getAbiContractVersion } from '../utils/calldata/cairo';
-import { isSierra } from '../utils/contract';
+import { extractContractHashes, isSierra } from '../utils/contract';
+import { solidityUint256PackedKeccak256 } from '../utils/hash';
 import { isBigNumberish, toBigInt, toHex } from '../utils/num';
 import { wait } from '../utils/provider';
 import { RPCResponseParser } from '../utils/responseParser/rpc';
+import { formatSignature } from '../utils/stark';
 import { GetTransactionReceiptResponse, ReceiptTx } from '../utils/transactionReceipt';
+import { getMessageHash, validateTypedData } from '../utils/typedData';
 import { LibraryError } from './errors';
 import { ProviderInterface } from './interface';
-import { solidityUint256PackedKeccak256 } from '../utils/hash';
-import { CallData } from '../utils/calldata';
-import { formatSignature } from '../utils/stark';
-import { getMessageHash, validateTypedData } from '../utils/typedData';
 
 export class RpcProvider implements ProviderInterface {
   public responseParser: RPCResponseParser;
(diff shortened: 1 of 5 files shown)


src/provider/rpc.ts
```diff
@@ -570,4 +573,59 @@ export class RpcProvider implements ProviderInterface {
...
+    }
+  }
+
+  /**
+   * Build bulk invocations with auto-detect declared class
+   * 1. Test if class is declared if not declare it preventing already declared class error and not declared class errors
+   * 2. Order declarations first
+   * @param invocations
+   */
+  public async createBulkInvocations(invocations: Invocations) {
```
penovicp, line 611: I would suggest renaming the method, I feel like users might make a mistake and think it's similar to `accountInvocationsFactory()`. Admittedly I haven't come up with a surefire alternative, maybe something like `sanitizeInvocations()`.

src/provider/rpc.ts
```diff
@@ -570,4 +573,59 @@ export class RpcProvider implements ProviderInterface {
...
+      classHash = contractClassIdentifier.classHash;
+    } else {
+      throw Error('contractClassIdentifier type not satisfied');
+    }
+
+    try {
+      const result = await this.getClass(classHash, blockIdentifier);
+      return result instanceof Object;
+    } catch (error) {
+      return false;
```
penovicp, line 601: Could be filtered for instances of `LibraryError` and re-throw otherwise.

penovicp (approved):
Two minor nitpicks, in general lgtm

src/types/lib/index.ts
```diff
@@ -99,6 +100,11 @@ export type DeclareContractPayload = {
   compiledClassHash?: string;
 };
 
+/**
+ * DeclareContractPayload with classHash or contract defined
+ */
+export type ContractClassIdentifier = DeclareContractPayload | { classHash: string };
```
PhilippeR26, line 106: Why an object containing the classHash? Seems complicated.
I think that the easiest way for the user is something like this :
```typescript
export type ContractClassIdentifier = DeclareContractPayload | BigNumberish;
```

src/provider/rpc.ts
PhilippeR26, line 611: I agree with @penovicp ; all these public functions with `bulk` in their names are not understood by the users.
Such names are OK for internal functions, but are problematic for basic users usage.
We have already this problem with `getEstimateFeeBulk` & `parseFeeEstimateBulkResponse`.
My proposal : `buildInvocations`

PhilippeR26 (commented):
Just little naming concerns.

src/types/lib/index.ts
tabaktoni: Because it would be ambiguous.
Class can be identified by classHash or by contract, in case of BigNumberish it can be both and non-specific.

PhilippeR26: For me, a class is only identified by its hash, and nothing else.

PhilippeR26: And if you have the address of an instance of a class, you have already the answer of question `is classDeclared`? 
It's more than weird to ask if a parent is existing if you have already one of its children....

tabaktoni: > And if you have the address of an instance of a class, you have already the answer of the question `is classDeclared`. It's more than weird to ask if a parent is existing if you have already one of its children....

Agree, and this does not contain an address. What address are you referring to?

tabaktoni: It does not deal with **Contract instance** nor addresses, just with **Contract Class** aka. definition classHash

**ContractClassIdentifier is** 

```ts
{
  contract: CompiledContract | string;
  classHash?: string;
  casm?: CompiledSierraCasm;
  compiledClassHash?: string;
};
```
In this case it covers cases for:
1. identifying class from its source code by computing classHash from contract
2. optional zero computing and classHash is already provided
3. less computing by providing optional params for classHash

```ts
or 
{ classHash: string }
```
This case cover providing non-optional classHash

So that in the end, you can get classHash which is in essence ContractClassIdentifier. 

Changing it to 
```ts
export type ContractClassIdentifier = DeclareContractPayload | BigNumberish;
```
In this case, BigNumberish represents classHash but in the source, we used mixed cases sometimes string sometimes BigNumberish.

Solutions:
1.
Another case is that we set ` | { classHash: BigNumberish }`.
This case is less ambiguous and I think we can do it.

2. create `type ClassHashIdentifier: BigNumberish`
and than do:
```ts
export type ContractClassIdentifier = DeclareContractPayload | ClassHashIdentifier;
```

src/provider/rpc.ts
tabaktoni: prepareBulkInvocations ?
prepareInvocations ?

src/types/lib/index.ts
tabaktoni: > For me, a class is only identified by its hash, and nothing else.

Yes but you can provide that classHash or you can provide source code and we can compute classHash from it.
So both source code and classHash are identifiers

PhilippeR26 (approved):
OK for me.

PR #1212, opened 2024-09-02
Chore: Types & utils improvements

lukasaric:
## Motivation and Resolution

- introduce `typed` util, which groups `typeof` like values checks (docs needs to be updated)
- improve types overall (return types, interfaces etc)
- remove `url` util
- add `assert.test.ts`

It resolves: #1214

## Usage related changes

_n.a_

_n.a_

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


src/utils/typed.ts
```diff
@@ -0,0 +1,3 @@
+export const isUndefined = (value: unknown): value is undefined => {
```
penovicp, line 1: I've been meaning to do a refactoring where a similar file would be introduced so the basic `typeof` wrapper functions could be aggregated from `utils/num` and `utils/shortString`. Seems like this is a convenient PR to include that change.

lukasaric: yeap, something that I also had in my mind, that's why this one is still in draft and I did not updated the description because of the potential TODO's :)

lukasaric: @penovicp, what's the process for introducing new versioned docs? I'm asking since we'll need to update the versioned docs since utilities are changed now

penovicp: The active pre-release documentation is available as `Next` in the dropdown, the `guides` are written manually and tracked by version control and the `API` segment is generated automatically. The versioned docs are essentially snapshots of the `Next` documentation when a starknet.js version is promoted to a full release (merged to `main`), usually there is no need change them. So the changes in this PR will be automatically available as `Next` once it is merged to develop.

And just to provide the full context, the customary approach for generating new versioned docs is for one of the maintainers to run the `[Manual] Documentation Version PR` workflow on the core repo that generates the docs and opens a corresponding PR. This can be tested on any GitHub fork or the workflow script steps can be reproduced to check the behaviour locally.

src/utils/encode.ts
```diff
@@ -262,12 +269,9 @@ export function sanitizeBytes(str: string, byteSize = 8, padding = STRING_ZERO):
  * ```
  */
 export function sanitizeHex(hex: string): string {
-  hex = removeHexPrefix(hex);
-  hex = sanitizeBytes(hex, 2);
-  if (hex) {
-    hex = addHexPrefix(hex);
-  }
-  return hex;
+  const hexWihtoutPrefix = removeHexPrefix(hex);
```
penovicp, line 272: "without" typo

src/utils/stark.ts
```diff
@@ -280,15 +289,15 @@ export function toTransactionVersion(
 /**
  * Convert Transaction version to Fee version or throw an error
  * @param {BigNumberish} [providedVersion] 0..3 number representing the transaction version
- * @returns {ETransactionVersion} the fee estimation version corresponding to the transaction version provided
+ * @returns {TransactionVersion} the fee estimation version corresponding to the transaction version provided
```
penovicp, line 292: `{TransactionVersion}` doesn't match the `ETransactionVersion | undefined` return type from the function signature (although I'm pretty sure it doesn't matter and the `@returns` types are mostly unused).

src/utils/url.ts
```diff
@@ -1,79 +0,0 @@
-import urljoin from 'url-join';
```
penovicp, line 1: It should be possible to remove this dependency now.

lukasaric: nice catch 🎯

penovicp (approved):

src/utils/fetchPonyfill.ts
```diff
@@ -2,7 +2,9 @@
 // @ts-ignore
 import makeFetchCookie from 'fetch-cookie';
 import isomorphicFetch from 'isomorphic-fetch';
+import { IS_BROWSER } from './encode';
+import { isUndefined } from './typed';
 
-export default (typeof window !== 'undefined' && window.fetch) || // use buildin fetch in browser if available
-  (typeof global !== 'undefined' && makeFetchCookie(global.fetch)) || // use buildin fetch in node, react-native and service worker if available
+export default (IS_BROWSER && window.fetch) || // use buildin fetch in browser if available
+  (!isUndefined(global) && makeFetchCookie(global.fetch)) || // use buildin fetch in node, react-native and service worker if available
```
ivpavici, line 9:
```suggestion
  (!isUndefined(global) && makeFetchCookie(global.fetch)) || // use builtin fetch in node, react-native and service worker if available
```

ivpavici, line 9:
```suggestion
  (!isUndefined(global) && makeFetchCookie(global.fetch)) || // use builtin fetch in node, react-native and service worker if available
```

src/utils/encode.ts
```diff
@@ -183,15 +187,14 @@ function padString(str: string, length: number, left: boolean, padding = STRING_
  * @param {number} length The target length for the padded string.
  * @param {string} [padding='0'] The string to use for padding. Defaults to '0'.
  * @returns {string} The padded string.
- *
- * @example
+ * * @example
```
ivpavici, line 190:
```suggestion
 * @example
```

I think examples are without this extra *

ivpavici (approved):

PR #1213, opened 2024-09-02
feat: increase rpc error information propagation

penovicp:
## Motivation and Resolution

The aim of this PR is to make the structured RPC error information more accessible which should enable more nuanced error handling.

## Usage and development related changes

A new `RpcError` class is introduced which exposes the RPC error information through its properties. It also contains an `isType()` convenience method which can be used as follows:

```typescript
try {
  ...
} catch (error) {
  if (error instanceof RpcError && error.isType('CONTRACT_NOT_FOUND')) { ... }
}
```

`RpcError` is a subclass of `LibraryError` that uses the same error message format as the current implementation so the changes are backward compatible.

## TODO:
- types
- docs

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


scripts/generateRpcErrorMap.js
```diff
@@ -0,0 +1,19 @@
+const starknet_api_openrpc = require('starknet_specs/api/starknet_api_openrpc.json');
```
ivpavici, line 1: should this script be periodically ran to check if there are changes in the rpc specs - and as a consequence the utils/errors files should be updated?

If so I would make a note in this file, about it's purpose and why it exists?

penovicp: I added a description. The script was referenced in the locations where its results are used, but more information can't hurt.

It's fairly rudimentary so it doesn't "need" to be run, it just simplifies the corresponding manual task. It could be argued that the code mapping object output is unnecessary since its typing enforces the exact entries, I left it in since the output ensures a descending code order. The script won't be used often, just when the RPC specification support is updated.

I also toyed with the idea of directly manipulating the target files but that made the code too convoluted for the current functionality. If other scripts are added I would revisit that idea and move them to the `types-js` module/repo. For some wider context, the script is related to a remnant idea when the automation of the TS type generation from the RPC specification was being explored. My idea was to have two tiers, a one-to-one representation created by some tool that conforms to the JSON-RPC standard and an added set of utility types created by custom mappers.

ivpavici (approved):

Issue #1217, opened 2024-09-03
Utils events tests


lukasaric:
**Is your feature request related to a problem? Please describe.**
Write tests for `./src/utils/events/index.ts`

**Describe the solution you'd like**
_n.a_

**Describe alternatives you've considered**
_n.a_

**Additional context**
_n.a_

Jemiiah:
@lukasaric can I work on this issue

PR #1218, opened 2024-09-03
chore: add events util tests

lukasaric:
## Motivation and Resolution
Write `events` util tests to improve tests coverage. That includes tests for these utiils:

- [x] - `isAbiEvent`
- [x] - `getAbiEvents`
- [x] - `parseEvents`
- [x] - `parseUDCEvent`

It resolves: #1217

_n.a_

## Usage related changes

_n.a_

`isObject` will not be covered here since it is moved to a different util in #1212 

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


src/utils/events/index.ts
```diff
@@ -265,7 +266,7 @@ export function parseEvents(
  * @returns DeployContractResponse | UDC Event Response data
  */
 export function parseUDCEvent(txReceipt: InvokeTransactionReceiptResponse) {
-  if (!txReceipt.events) {
+  if (!txReceipt.events?.length) {
```
lukasaric, line 269: The old implementation will never happen (`events` are mandatory by the TS). Either we can check the `length` as I did in the PR and throw an error if there are no events, or we can remove the error handling part, and it will just return the response with the undefined values (it worked like that in the prev. implementation). In both ways, all tests pass. @penovicp thoughts? 🤔

penovicp (approved):

ivpavici (approved):

PR #1227, opened 2024-09-13
feat: add support for parsing emitted events

3alpha:
## Motivation and Resolution

`provider.getEvents` method returns list of `EmittedEvents` which contains fields for block number, block hash and transaction hash. Parsing list of those events with `parseEvents` would drop those fields. This PR adds them back to the returned array if they are available.

...

## Usage related changes

Users now use additional data from parsed events structure.

N/A

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/src/utils/events/index.ts
+++ b/src/utils/events/index.ts
@@ -194,7 +194,7 @@ function mergeAbiEvents(target: any, source: any): Object {
  * ```
  */
 export function parseEvents(
-  providerReceivedEvents: RPC.EmittedEvent[] | RPC.Event[],
+  providerReceivedEvents: RPC.EmittedEvent[],
   abiEvents: AbiEvents,
   abiStructs: AbiStructs,
   abiEnums: AbiEnums
--- a/src/utils/events/index.ts
+++ b/src/utils/events/index.ts
@@ -194,57 +194,62 @@ function mergeAbiEvents(target: any, source: any): Object {
  * ```
  */
 export function parseEvents(
-  providerReceivedEvents: RPC.Event[],
+  providerReceivedEvents: RPC.EmittedEvent[] | RPC.Event[],
   abiEvents: AbiEvents,
   abiStructs: AbiStructs,
   abiEnums: AbiEnums
 ): ParsedEvents {
-  const ret = providerReceivedEvents.flat().reduce((acc, recEvent: RPC.Event) => {
-    let abiEvent: AbiEvent | AbiEvents = abiEvents[recEvent.keys.shift() ?? 0];
-    if (!abiEvent) {
+  const ret = providerReceivedEvents
+    .flat()
+    .reduce((acc, recEvent: RPC.EmittedEvent | RPC.Event) => {
+      let abiEvent: AbiEvent | AbiEvents = abiEvents[recEvent.keys.shift() ?? 0];
+      if (!abiEvent) {
+        return acc;
+      }
+      while (!abiEvent.name) {
+        const hashName = recEvent.keys.shift();
+        assert(!!hashName, 'Not enough data in "keys" property of this event.');
+        abiEvent = (abiEvent as AbiEvents)[hashName];
+      }
+      // Create our final event object
+      const parsedEvent: ParsedEvent = {};
+      parsedEvent[abiEvent.name as string] = {};
+      // Remove the event's name hashed from the keys array
+      const keysIter = recEvent.keys[Symbol.iterator]();
+      const dataIter = recEvent.data[Symbol.iterator]();
+
+      const abiEventKeys =
+        (abiEvent as CairoEventDefinition).members?.filter((it) => it.kind === 'key') ||
+        (abiEvent as LegacyEvent).keys;
+      const abiEventData =
+        (abiEvent as CairoEventDefinition).members?.filter((it) => it.kind === 'data') ||
+        (abiEvent as LegacyEvent).data;
+
+      abiEventKeys.forEach((key) => {
+        parsedEvent[abiEvent.name as string][key.name] = responseParser(
+          keysIter,
+          key,
+          abiStructs,
+          abiEnums,
+          parsedEvent[abiEvent.name as string]
+        );
+      });
+
+      abiEventData.forEach((data) => {
+        parsedEvent[abiEvent.name as string][data.name] = responseParser(
+          dataIter,
+          data,
+          abiStructs,
+          abiEnums,
+          parsedEvent[abiEvent.name as string]
+        );
+      });
+      if ('block_hash' in recEvent) parsedEvent.block_hash = recEvent.block_hash;
+      if ('block_number' in recEvent) parsedEvent.block_number = recEvent.block_number;
+      if ('transaction_hash' in recEvent) parsedEvent.transaction_hash = recEvent.transaction_hash;
+      acc.push(parsedEvent);
       return acc;
-    }
-    while (!abiEvent.name) {
-      const hashName = recEvent.keys.shift();
-      assert(!!hashName, 'Not enough data in "keys" property of this event.');
-      abiEvent = (abiEvent as AbiEvents)[hashName];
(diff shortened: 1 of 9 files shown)


PhilippeR26:
Hello,
Seems you didn't created any issue for this subject.
What's your motivation? Bug resolution? improvement?
In which case will this modification be useful?

tabaktoni (approved):
lgtm!

tabaktoni:
> Hello, Seems you didn't created any issue for this subject. What's your motivation? Bug resolution? improvement? In which case will this modification be useful?

It looks like we lost part of the information about events's block and tx after parsing the event.
As this part is added at the end of obj it should be backward compatible?

3alpha:
> Hello, Seems you didn't created any issue for this subject. What's your motivation? Bug resolution? improvement? In which case will this modification be useful?

Hey, sorry, I can create issue if necessary. Issue this solves is situation where you get events via `provider.getEvents` method, you parse them and want to know, for example, who triggered transaction. Currently, data that connects events with "upsteam" structures like transactions and blocks gets dropped and this fixes that. With this fix, you can easily fetch transaction hash and from there fetch transaction and from there caller.

Only issue is how are these new keys ordered so backwards compatibility is not accidentally broken in code that uses `Object.keys(event)[0]` to fetch event names. I am pretty sure its by order of insertion, but I am not 100% sure.

penovicp (approved):

penovicp:
> > Hello, Seems you didn't created any issue for this subject. What's your motivation? Bug resolution? improvement? In which case will this modification be useful?
> 
> Hey, sorry, I can create issue if necessary. Issue this solves is situation where you get events via `provider.getEvents` method, you parse them and want to know, for example, who triggered transaction. Currently, data that connects events with "upsteam" structures like transactions and blocks gets dropped and this fixes that. With this fix, you can easily fetch transaction hash and from there fetch transaction and from there caller.
> 
> Only issue is how are these new keys ordered so backwards compatibility is not accidentally broken in code that uses `Object.keys(event)[0]` to fetch event names. I am pretty sure its by order of insertion, but I am not 100% sure.

The object key order used to undefined, but when people kept relying on the JS engine behaviour that preserved the order it was eventually codified in the language specification. So non-numeric string keys, like the ones from this PR, should be ordered by their insertion order.

src/utils/events/index.ts
```diff
@@ -194,57 +194,62 @@ function mergeAbiEvents(target: any, source: any): Object {
  * ```
  */
 export function parseEvents(
-  providerReceivedEvents: RPC.Event[],
+  providerReceivedEvents: RPC.EmittedEvent[] | RPC.Event[],
```
PhilippeR26, line 197: Why `RPC.EmittedEvent[] | RPC.Event[]`?
Isn't it only `RPC.EmittedEvent[]`?
In which case are  block_hash, block_number & transaction_hash not included?

PhilippeR26 (commented):
Just a question.

src/utils/events/index.ts
3alpha: Block/tx info is not included in a case where you first fetch transaction receipt with `starknet_getTransactionReceipt`. In those kinds of reponses `RPC.Event` struct is returned (see: [this](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L2661)). On the other hand when fetching data with `starknet_fetchEvents` returned struct is `RPC.EmittedEvents`.

So I thought it would be nice to leave that as is in case someone had code like this
```js
events.parseEvents(receipts.map(t => t.events),...)
```

PhilippeR26: I think that the root cause of this wrong input type of `events.parseEvents()` is due to `Contract.parseEvent()`. This method is calling `events.parseEvents()` with `RPC.event `format ; here is the root cause of all this mess.
I think that you have to modify `Contract.parseEvent()` : 
- if the event is included in a block, this type has to be sent to `events.parseEvents()` :
```
type EMITTED_EVENT = EVENT & {
  block_hash: BLOCK_HASH;
  block_number: BLOCK_NUMBER;
  transaction_hash: TXN_HASH;
};
```
- if the event is in a PENDING transaction, this type has to be sent to `events.parseEvents()` :
```
type EMITTED_EVENT = EVENT & {
  transaction_hash: TXN_HASH;
};
```
`events.parseEvents()` has to accept both types as input.

`RPC.Event[]` type was here only to handle `Contract.parseEvent()` ; it's useless for users. They uses `EVENTS_CHUNK`type (result of `RpcProvider.getEvents()`), that includes `EmittedEvent` type. Never `RPC.Events`.

3alpha: Not sure how pending transaction have anything to do with this. AFAIK events are created and emitted during transaction execution. So there is always tx hash, block hash and block number corresponding to the event. They are missing from events array of the recipiet because that is always available in parent (receipt) object.

PhilippeR26: If PENDING is not appropriate, then we have still a problem in Contract.parseEvents() : 
https://github.com/starknet-io/starknet.js/blob/47e52cf39a71bf99188edc4991b002018e296504/src/contract/default.ts#L333
`InvokeTransactionReceiptResponse` type is not the good one, because it's made of `INVOKE_TXN_RECEIPT | PENDING_INVOKE_TXN_RECEIPT`
And in this case, only `EMITTED_EVENT`  should be transfered from `Contract.parseEvent()` to `events.parseEvents()` (not `RPC.event`).
Comment about uselessness of `RPC.event` as input of `events.parseEvents()` remains valid.

3alpha: So if I understood correctly, you would like `Contract.parseEvent()` to change and look like this?
```ts
  public parseEvents(receipt: GetTransactionReceiptResponse): ParsedEvents {
    return parseRawEvents(
      (receipt as InvokeTransactionReceiptResponse).events?.
      map(event => { return {
        block_hash: (receipt as INVOKE_TXN_RECEIPT).block_hash,
        block_number: (receipt as INVOKE_TXN_RECEIPT).block_number,
        transaction_hash: (receipt as INVOKE_TXN_RECEIPT).transaction_hash,
        ...event
      }})
      .filter(
        (event) => cleanHex(event.from_address) === cleanHex(this.address),
        []
      ) || [],
      this.events,
      this.structs,
      CallData.getAbiEnum(this.abi)
    );
  }
  ```

Comment (PhilippeR26) [reply src/utils/events/index.ts:197]: Yes, this way.
I tested successfully this :
**account/default.ts**
```typescript
public parseEvents(receipt: GetTransactionReceiptResponse): ParsedEvents {
let parsed: ParsedEvents;
receipt.match({
  success: (txR: SuccessfulTransactionReceiptResponse) => {
    const emittedEvents =
      (txR as InvokeTransactionReceiptResponse).events
        ?.map((event) => {
          return {
            block_hash: (txR as INVOKE_TXN_RECEIPT).block_hash,
            block_number: (txR as INVOKE_TXN_RECEIPT).block_number,
            transaction_hash: (txR as INVOKE_TXN_RECEIPT).transaction_hash,
            ...event,
          };
        })
        .filter((event) => cleanHex(event.from_address) === cleanHex(this.address), []) || [];
    parsed = parseRawEvents(
      emittedEvents,
      this.events,
      this.structs,
      CallData.getAbiEnum(this.abi)
    );
  },
  _: () => {
    throw Error('This transaction was not successful.');
  },
});
return parsed!;
}
```

**events/index.ts**
```typescript
export function parseEvents(
  providerReceivedEvents: RPC.EmittedEvent[],
  abiEvents: AbiEvents,
  abiStructs: AbiStructs,
  abiEnums: AbiEnums
): ParsedEvents {
...
```

PhilippeR26:
I have implemented my last message, to finalize this PR.

tabaktoni (approved):
Updated test remove deletes with custom matchers

PR #1253, opened 2024-10-31
fix: contract class splitArgsAndOptions

### LINKED ISSUE #1252: `splitArgsAndOptions` with single argument containing `options`
**Describe the bug**

```
export const splitArgsAndOptions = (args: ArgsOrCalldataWithOptions) => {
  const options = [
    'blockIdentifier',
    'parseRequest',
    'parseResponse',
    'formatResponse',
    'maxFee',
    'nonce',
    'signature',
    'addressSalt',
  ];
  const lastArg = args[args.length - 1];
  if (typeof lastArg === 'object' && options.some((x) => x in lastArg)) {
    return { args: args as ArgsOrCalldata, options: args.pop() as ContractOptions };
  }
  return { args: args as ArgsOrCalldata };
};
```

When it's called during [buildCall](https://github.com/starknet-io/starknet.js/blob/ad70b455ae108970cd739e107d9d1367372b8d52/src/contract/default.ts#L58-L67), and the contract expects a single argument which is struct with a property that overlaps this list of hardcoded options. Resulting in this:

```
Error: Invalid number of arguments, expected 1 arguments, but got 0
    at _CallData.validate (index.ts:95:13)
    at default.ts:236:23
    at getCalldata (default.ts:115:10)
    at Contract.call (default.ts:234:22)
    at Contract.get_proposal_hash (default.ts:61:21)
```

**To Reproduce**

I exported ABIs using abi-wan-kanabi for [this contracts](https://sepolia.voyager.online/contract/0x050c97a3d6ead0f7998ef29e1bbd85d3662abd152016406ee1cb2d37d156fdb5#readContract) and tried to call the `get_proposal_hash`. I spotted the error because the similar call worked for the function `encode_proposal_data` that accepts two arguments and works just fine. 

**Expected behavior**
- 

- Browser & version Chrome Version 124.0.6367.208
- Node version v20.14.0
- Starknet.js version 6.17.0
- Network devnet

tabaktoni:
## Motivation and Resolution
fix #1252

```
export type ArgsOrCalldata = RawArgsArray | [Calldata] | Calldata;
export type ArgsOrCalldataWithOptions = ArgsOrCalldata & ContractOptions;
```
If we thrust types, It seems that ContractOptions is the only object and every other possible input is an Array.

...

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/contract/default.ts
+++ b/src/contract/default.ts
@@ -5,6 +5,7 @@ import { ProviderInterface, defaultProvider } from '../provider';
 import {
   Abi,
   AbiEvents,
+  AbiStruct,
   ArgsOrCalldata,
   ArgsOrCalldataWithOptions,
   AsyncContractFunction,
@@ -21,7 +22,6 @@ import {
   ParsedEvents,
   RawArgs,
   Result,
-  AbiStruct,
   ValidateType,
 } from '../types';
 import assert from '../utils/assert';
@@ -29,8 +29,8 @@ import { CallData, cairo } from '../utils/calldata';
 import { createAbiParser } from '../utils/calldata/parser';
 import { getAbiEvents, parseEvents as parseRawEvents } from '../utils/events/index';
 import { cleanHex } from '../utils/num';
-import { ContractInterface } from './interface';
 import type { GetTransactionReceiptResponse } from '../utils/transactionReceipt';
+import { ContractInterface } from './interface';
 
 export type TypedContractV2<TAbi extends AbiKanabi> = AbiWanTypedContract<TAbi> & Contract;
 
@@ -46,7 +46,7 @@ export const splitArgsAndOptions = (args: ArgsOrCalldataWithOptions) => {
     'addressSalt',
   ];
   const lastArg = args[args.length - 1];
-  if (typeof lastArg === 'object' && options.some((x) => x in lastArg)) {
+  if (typeof lastArg === 'object' && !Array.isArray(lastArg) && options.some((x) => x in lastArg)) {
     return { args: args as ArgsOrCalldata, options: args.pop() as ContractOptions };
   }
   return { args: args as ArgsOrCalldata };


penovicp (commented):
I don't think the `ArgsOrCalldataWithOptions` type is accurate, the `ContractOptions` object is used as a concatenation to the regular parameters of the encapsulated method and these get passed to the `splitArgsAndOptions` utility. Example from the tests: [link](https://github.com/starknet-io/starknet.js/blob/v6.17.0/__tests__/contract.test.ts#L572). So the type would be closer to something like this: `[...ArgsOrCalldata, ContractOptions?]`

The issue boils down to differentiating whether the last element is a `ContractOptions` or a `RawArgsObject` or `MultiType` object.

tabaktoni:
OK i belave at least now i fixed type

penovicp:
Yes, that is a more accurate representation.

I believe the spread `Calldata` branch can also be removed (`[...Calldata] | [...Calldata, ContractOptions]` from the reworked `ArgsOrCalldataWithOptions` type). It is supported by `splitArgsAndOptions` itself, however, `splitArgsAndOptions` is in all its usages followed by the `getCalldata()` utility that filters the argument based on the `__compiled__` marker which isn't propagated if the `Calldata` is spread. This means that the only spread input that should be able to work is `method(...Calldata, { parseRequest: false })`.

I don't see any utility in restoring the full support, so refactoring the code to exclude the supported scenario seems preferable, it should also simplify detecting whether a `ContractOptions` object is used.

tabaktoni:
I agree focused on this pr to fix type and new one to propose solution

penovicp (approved):

src/types/contract.ts
```diff
@@ -21,8 +21,37 @@ export type Result =
   | boolean
   | CairoEnum;
 
-export type ArgsOrCalldata = RawArgsArray | [Calldata] | Calldata;
-export type ArgsOrCalldataWithOptions = ArgsOrCalldata & ContractOptions;
+// export type ArgsOrCalldata = RawArgsArray | [Calldata] | Calldata;
```
ivpavici, line 24: only q I have is if these are left intentionally here commented out?

PhilippeR26 (approved):
No remarks.

PR #1245, opened 2024-10-10
[V7 Breaking] feat: improve WalletAccount class instantiation

dhruvkelawala:
## Motivation and Resolution

Currently, the `WalletAccount` class in the `account.ts` file has the connection logic and account address retrieval mixed with the object construction process in the constructor. This violates the principle of separation of concerns and can lead to potential issues such as race conditions and difficulty in testing.

To address this concern, the connection logic and account address retrieval have been separated from the constructor and moved into static `connect` and `connectSilent` methods. This change ensures that the object construction process remains synchronous and focused on initializing the object's state, while the connection logic is handled separately.

### RPC version (if applicable)

N/A

## Usage related changes

No changes to the usage of the WalletAccount class from the user's perspective.

## Development related changes

- The `WalletAccount` class constructor no longer includes any asynchronous operations or connection logic. It now focuses solely on initializing the object's state and setting up event listeners.

- The `connect` and `connectSilent` static methods have been introduced to handle the connection logic and account address retrieval from the wallet provider.

- Developers should now use the `connect` or `connectSilent` methods to create instances of `WalletAccount` with the appropriate account address retrieved from the wallet provider. 

- This approach facilitates better integration with libraries like `starknetkit` and `starknet-react`, which handle the connection logic internally. After establishing a connection, these libraries can return a `WalletAccount` instance without attempting to connect again, as the connection logic is now separated from the object construction process.

## Checklist:

- [x] Performed a self-review of the code
- [x] Rebased to the last commit of the target branch (or merged it into my branch)
- [x] Linked the issues which this PR resolves
- [x] Documented the changes in code (API docs will be generated automatically)
- [x] Updated the tests
- [x] All tests are passing

PR #1259, opened 2024-11-09
feat: WalletAccount non-breaking temp solution

tabaktoni:
## Motivation and Resolution

Based on #1245
This should be a temporary fix that can be merged as it should not contain breaking changes.

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


PhilippeR26:
What are the differences with PR1245?

tabaktoni:
> What are the differences with PR1245?

Please check now, I started PR but got pulled on HH

tabaktoni:
In general, the idea is to have a non-breaking constructor solution. So for existing users, this should continue to work as it was before, but now they can also provide an address or use static methods to avoid wallet popup.

PhilippeR26:
But we have still something async in the constructor, that is the root cause of PR 1245 creation.

tabaktoni:
The imminent Issue is the wallet popup opening without the option to stop it. 
This resolves it without breaking changes so it could be merged asap.

The more proper solution could go into the next mayor versions with breaking changes.

penovicp (approved):
Left two suggestions that truncate the new code. Aside from that lgtm.

PhilippeR26 (commented):
I think that this solution is the best compromise, even if we don't solve the initial problem without braking changes.

For Information, OKX wallet team is reporting that they have problems with this subject.
They are waiting an new version with `connect`.
So, we should go forward on this subject.

src/wallet/account.ts
```diff
@@ -64,17 +66,11 @@ export class WalletAccount extends Account implements AccountInterface {
...
-          silent_mode: false,
-        },
-      })
-      .then((res) => {
-        this.address = res[0].toLowerCase();
+    if (!address.length) {
+      requestAccounts(this.walletProvider).then(([accountAddress]) => {
+        this.address = accountAddress.toLowerCase();
       });
+    }
```
dhruvkelawala, line 73: Let's add a deprecation warning here saying that connect should be use instead.

src/wallet/account.ts
```diff
@@ -170,5 +166,15 @@ export class WalletAccount extends Account implements AccountInterface {
     return signMessage(this.walletProvider, typedData);
   }
 
+  static async connect(
+    provider: ProviderInterface,
+    walletProvider: StarknetWalletProvider,
+    cairoVersion?: CairoVersion,
+    silentMode: boolean = false
```
dhruvkelawala, line 173: I like my implementation better. Separating "connect" and "connectSilent" is a better approach imo, then passing a boolean here

tabaktoni: OK will add connectSilent as alias on fixed connect, I think these best compromise on this.

dhruvkelawala:
If it's based on my PR, I would also suggest changing the base branch to my PR branch and then merge this PR as "Backwards compatibility" or something.

src/wallet/account.ts
tabaktoni: c83e304daa58d97b0e4abac6c06f03fb3ba1652e

src/wallet/account.ts
tabaktoni: c83e304daa58d97b0e4abac6c06f03fb3ba1652e

PR #1264, opened 2024-11-15
docs: walletAccount - add details for chainId

PhilippeR26:
## Motivation and Resolution
Following a suggestion of @tabaktoni , the guide of `WalletAccount` has been updated, to add explanation about `chainId`.

## Usage related changes
N/A

N/A

## Checklist:
- [x] All tests are passing

### DIFF
--- a/www/docs/guides/walletAccount.md
+++ b/www/docs/guides/walletAccount.md
@@ -35,7 +35,7 @@ Instantiating a new `WalletAccount`:
 
 ```typescript
 import { connect } from '@starknet-io/get-starknet'; // v4.0.3 min
-import { WalletAccount } from 'starknet'; // v6.18.0 min
+import { WalletAccount, wallet } from 'starknet'; // v6.18.0 min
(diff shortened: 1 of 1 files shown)


tabaktoni:
Should we add `requestChainId` to WalletAccount class?
Also `deploymentData` and `supportedSpecs` are not there I can't remember why.

PhilippeR26:
We have also to solve item 2 of https://github.com/starknet-io/types-js/issues/20

PhilippeR26:
> Should we add `requestChainId` to WalletAccount class? 

Seems not mandatory, but if we do it, it will be necessary to use a method name that explains that the response is about the wallet (not the account).

tabaktoni:
> We have also to solve item 2 of [starknet-io/types-js#20](https://github.com/starknet-io/types-js/issues/20)

Resolved

PhilippeR26:
Nice. Now, just need a new version of the lib.

PR #1272, opened 2024-11-26
test: stabilize websocket tests

penovicp:
## Motivation and Resolution

This PR contains small tweaks to stabilize the tests:
- moving the `WebSocketChannel` initializations into setup hooks ensures they are bound to their respective `describe` scopes and properly followed by the teardown hooks, without the change they are treated as if they were in the same higher scope causing one of them to be orphaned in case of errors
- removing the underlying `isows` dependency with a mock allows the tests from `utils/stark.browser.test.ts` to run


tabaktoni (approved):
WP

PR #1289, opened 2025-01-01
fix: missing snip-12 enum type dependency

### LINKED ISSUE #1287: SNIP-12 enum type encoding fails to include referenced types except when used alone
**Describe the bug**

When encoding an enum type that uses another user-defined type as an array, the referenced type fails to be included in the encoded signature.

_Edit: After looking into the problematic code I realized this is the case not just for when using as array. It happens as long as `contains` is not **exactly** `(DEPENDENCY_TYPE)`. So the reproducer below would still trigger the issue even if `(Child*)` is changed to something like `(u128,Child)`._

**To Reproduce**

Run this code:

```ts
import { typedData, TypedDataRevision } from "starknet";

console.log(
  typedData.encodeType(
    {
      Parent: [{ name: "variant", type: "(Child*)" }],
      Child: [{ name: "test", type: "u128" }],
    },
    "Parent",
    TypedDataRevision.ACTIVE,
  ),
);
```

and it prints:

```log
"Parent"("variant":("Child*"))
```

**Expected behavior**

The correct encoding would instead be:

```log
"Parent"("variant":("Child*"))"Child"("test":"u128")
```

since `Child` is indeed used by `Parent`.

**Screenshots**

N/A

- Node version: `v20.17.0`
- Starknet.js version: `v6.20.3`

**Additional context**

N/A

xJonathanLEI:
## Motivation and Resolution

Resolves #1287.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__mocks__/typedData/example_enumNested.json
+++ b/__mocks__/typedData/example_enumNested.json
@@ -1,0 +1,33 @@
+{
+  "types": {
+    "StarknetDomain": [
+      { "name": "name", "type": "shortstring" },
+      { "name": "version", "type": "shortstring" },
+      { "name": "chainId", "type": "shortstring" },
+      { "name": "revision", "type": "shortstring" }
+    ],
+    "Example": [{ "name": "someEnum", "type": "enum", "contains": "EnumA" }],
+    "EnumA": [
+      { "name": "Variant 1", "type": "()" },
+      { "name": "Variant 2", "type": "(u128,StructA)" }
+    ],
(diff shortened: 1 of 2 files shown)


penovicp (commented):
Could you add a test message example similar to `__mocks__/typedData/example_enum.json` and corresponding tests for `encodeType`, `getTypeHash`, and `getMessageHash` so the issue that the PR fixes is covered in case of future code changes?

xJonathanLEI:
Sure thing. Sorry I should have added the tests to start with. I will generate the tests using my impl in `starknet-rs`.

PR #1292, opened 2025-01-07
Typed data enum fixes (SNIP-12)

penovicp:
## Motivation and Resolution

Aggregate for the typed data enum fixes. Extracted so potential impact can be coordinated and so that less impactful prereleases can proceed unhindered.

Aggregated PRs:
- #1281 
- #1288 
- #1289

### DIFF
--- a/__mocks__/typedData/example_enumNested.json
+++ b/__mocks__/typedData/example_enumNested.json
@@ -1,0 +1,33 @@
+{
+  "types": {
+    "StarknetDomain": [
+      { "name": "name", "type": "shortstring" },
+      { "name": "version", "type": "shortstring" },
+      { "name": "chainId", "type": "shortstring" },
+      { "name": "revision", "type": "shortstring" }
+    ],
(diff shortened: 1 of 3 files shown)


xJonathanLEI:
Hi @penovicp can we have confirmation on whether #1289 would be released? There seems to be consensus that this particular patch is the most harmless out of the three (in terms of breakage), but there doesn't seem to be any activity here.

penovicp:
> Hi @penovicp can we have confirmation on whether #1289 would be released? There seems to be consensus that this particular patch is the most harmless out of the three (in terms of breakage), but there doesn't seem to be any activity here.

Just released it as the [v6.24.1](https://github.com/starknet-io/starknet.js/releases/tag/v6.24.1) pre-release. it will be promoted as a main release before the v7 beta with RPC 0.8 support is promoted to a pre-release, so should be relatively soon.

tabaktoni:
@penovicp Should convert this to a draft while we agree on the new revision?

PR #1293, opened 2025-01-07
fix: repair fetch utility for browser environments

penovicp:
## Motivation and Resolution

`tough-cookie`, a sub-dependency of `fetch-cookie`, drops browser support in its latest major release which is required by the latest `fetch-cookie` minor release, this in turn prevents browser builds that include starknet.js as a dependency from successfully completing.


TeddyNotBear:
Hey, could you check your PR so that the patch works correctly in the prod version. We've been having a problem with Starknet.JS since this morning with tough-cookie It doesn't seem to work well on ^6.11.0.

image

PR #1301, opened 2025-01-27
fix: add starkid improvements

fricoben:
## Motivation and Resolution

Argent ran into a problem using `useDecoded` when decoding the domain `Grug.stark`, this should have been passed as `grug.stark` but they didn't have the `isStarkDomain` to do so easily. 
 
## Usage related changes

I'm addding two things to help wallet and devs in general:

1. Adding `isStarkDomain` to permit manual verification
2. Refactoring `getAddressFromStarkName` to make the function detect this kind of name errors in the future

### DIFF
--- a/__tests__/account.starknetId.test.ts
+++ b/__tests__/account.starknetId.test.ts
@@ -86,7 +86,7 @@ describe('deploy and test Wallet', () => {
   });
 
   test('Should throw error when invalid stark domain is provided', async () => {
-    await expect(account.getAddressFromStarkName('invalid-domain', namingAddress)).rejects.toThrow(
+    await expect(account.getAddressFromStarkName('invalid_domain', namingAddress)).rejects.toThrow(
       'Invalid domain, must be a valid .stark domain'
     );
   });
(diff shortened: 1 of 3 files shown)


__tests__/account.starknetId.test.ts
```diff
@@ -85,6 +85,12 @@ describe('deploy and test Wallet', () => {
     expect(hexToDecimalString(address)).toEqual(hexToDecimalString(account.address));
   });
 
+  test('Should throw error when invalid stark domain is provided', async () => {
+    await expect(account.getAddressFromStarkName('invalid-domain', namingAddress)).rejects.toThrow(
```
penovicp, line 89:
```suggestion
    await expect(account.getAddressFromStarkName('InvalidDomain', namingAddress)).rejects.toThrow(
```
Since the opening ternary from `getAddressFromStarkName` appends `.stark` to the name input `invalid-domain.stark` ends up matching the `isStarkDomain` regex and the test fails.

penovicp (changes requested):

__tests__/account.starknetId.test.ts
fricoben: Yep my bad, added a forbidden character `_` so it works

penovicp (approved):

PR #1302, opened 2025-01-28
logger, config, deprecate legacy tx

tabaktoni:
## Motivation and Resolution

Implement logger and config
deprecate warning for old transaction versions

Use logger for console logging
Use config for global configuration

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/channel/rpc_0_6.ts
+++ b/src/channel/rpc_0_6.ts
@@ -1,4 +1,4 @@
-import { NetworkName, StarknetChainId } from '../global/constants';
+import { NetworkName, StarknetChainId, SYSTEM_MESSAGES } from '../global/constants';
 import { LibraryError, RpcError } from '../utils/errors';
 import {
   AccountInvocationItem,
@@ -450,10 +450,10 @@ export class RpcChannel {
         },
       });
 
-      logger.warn(
-        'You are using deprecated transaction version (V0,V1,V2)!, \n Update to the latest V3 transactions!',
-        { version: RPC.ETransactionVersion.V1, type: RPC.ETransactionType.INVOKE }
-      );
+      logger.warn(SYSTEM_MESSAGES.legacyTxWarningMessage, {
+        version: RPC.ETransactionVersion.V1,
+        type: RPC.ETransactionType.INVOKE,
+      });
     } else {
       // V3
       promise = this.fetchEndpoint('starknet_addInvokeTransaction', {
@@ -500,10 +500,10 @@ export class RpcChannel {
         },
       });
 
-      logger.warn(
-        'You are using deprecated transaction version (V0,V1,V2)!, \n Update to the latest V3 transactions!',
-        { version: RPC.ETransactionVersion.V1, type: RPC.ETransactionType.DECLARE }
-      );
+      logger.warn(SYSTEM_MESSAGES.legacyTxWarningMessage, {
+        version: RPC.ETransactionVersion.V1,
+        type: RPC.ETransactionType.DECLARE,
+      });
     } else if (isSierra(contract) && !isV3Tx(details)) {
       // V2 Cairo1
       promise = this.fetchEndpoint('starknet_addDeclareTransaction', {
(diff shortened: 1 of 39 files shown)


src/channel/rpc_0_6.ts
```diff
@@ -448,6 +449,11 @@ export class RpcChannel {
           nonce: toHex(details.nonce),
         },
       });
+
+      logger.warn(
+        'You are using deprecated transaction version (V0,V1,V2)!, \n Update to the latest V3 transactions!',
```
penovicp, line 454:
```suggestion
        'You are using a deprecated transaction version (V0,V1,V2)!\nUpdate to the latest V3 transactions!',
```
Might also make sense to move the message into a `const enum` so that it gets both defined in one place and inlined.

src/global/config.ts
```diff
@@ -0,0 +1,63 @@
+import { DEFAULT_GLOBAL_CONFIG } from './constants';
+
+type ConfigData = {
```
penovicp, line 3: Might be worthwhile to interlace the keys with `DEFAULT_GLOBAL_CONFIG` for IDE autocompletion but we can revisit that in the future.

src/global/config.ts
```diff
@@ -0,0 +1,63 @@
...
+  public static getInstance(): Configuration {
+    if (!Configuration.instance) {
+      Configuration.instance = new Configuration();
+    }
+    return Configuration.instance;
+  }
+
+  public get<K extends keyof ConfigData>(
+    key: K,
+    defaultValue?: ConfigData[K]
```
penovicp, line 29: The `defaultValue` parameter looks like it doesn't provide much utility.

src/global/logger.ts
```diff
@@ -0,0 +1,171 @@
...
+interface LogMessage {
+  level: LogLevel;
+  message: string;
+  timestamp: string;
+  data?: any;
+}
+/**
+ * Logging class providing different levels of log
+ */
+class Logger {
```
penovicp, line 14: Probably makes sense to `export` so it is exposed to the documentation generator for the `API` section.

src/global/logger.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  private formatMessage(logMessage: LogMessage): string {
+    const { level, message, timestamp, data } = logMessage;
+    let formattedMessage = `[${timestamp}] ${level}: ${message}`;
+
+    if (data) {
+      try {
+        formattedMessage += `\n${JSON.stringify(data, null, 2)}`;
+      } catch (error) {
+        formattedMessage += `\n[JSON.stringify Error/Circular]`;
```
penovicp, line 47: Might make sense to propagate the error message since `stringify` could also fail for e.g. bigint values in data.

src/global/logger.ts
```diff
@@ -0,0 +1,171 @@
...
+   *
+   * @returns logs levels displayed on the configured LogLevel
+   */
+  public getEnabledLogLevels() {
+    const logLevelStringKeys = Object.keys(LogLevelIndex).filter(
+      (v) => Number.isNaN(Number(v)) && v !== 'OFF'
+    );
+    const logLevels = logLevelStringKeys.filter((s) => {
+      return this.shouldLog(LogLevelIndex[s as LogLevel]);
+    });
```
penovicp, line 153: The `Number.isNaN(Number(v))` check doesn't seem necessary so the two filters could be merged.

src/global/logger.type.ts
```diff
@@ -0,0 +1,14 @@
...
+  INFO: 4,
+  WARN: 3,
+  ERROR: 2,
+  FATAL: 1,
+  OFF: 0,
+};
+
+export type LogLevelIndex = ValuesType<typeof LogLevelIndex>;
+
+export type LogLevel = keyof typeof LogLevelIndex & string & NonNullable<unknown>;
```
penovicp, line 14: I think the ` & string & NonNullable<unknown>` segment can be removed.

penovicp (approved):
Mostly a bunch of nitpicks, main one is probably the one concerning doc generation visibility, in general lgtm.

I would also suggest adding a doc entry for the `Guides` section, but that can also be added afterwards.

src/global/logger.ts
tabaktoni: Leftover after refactoring from enum to const  pattern

src/global/logger.type.ts
tabaktoni: Yep, you are right, it looks like autocomplete with the keyof typeof work without shenanigans.

src/global/config.ts
tabaktoni: Nice catch, I managed to break this one too in the late-night refactor

src/global/config.ts
tabaktoni: It can be useful sometimes; I'll leave that one.

src/global/logger.ts
tabaktoni: I think I'll just add manual docs because it could confuse some users in this case.

penovicp (approved):

PR #1310, opened 2025-02-04
feat: handling of new cairo type: fixed array

PhilippeR26:
## Motivation and Resolution
Cairo has a new type of array : the fixed array, described as [type, quantity].
For example, for an array of 8 * u32 : `[core::integer::u32; 8]`.
The number of items in the array is frozen, so a prefix for the length of the array SHALL not be added.
Ex: encoding of `[core::integer::u8; 4]` for values `[1,2,3,4]` is `[1,2,3,4]` (no prefix).

> [!WARNING]
> For encoding functions that do not know the ABI, it's impossible to know if a JS Array is fixed or not. `CallData.compile()` has to be informed that you want to use a fixed array.
> `CairoFixedArray.compile()` has been introduced for use with this function.

## Usage related changes
For functions that uses the ABI, users are using JS arrays for both normal and fixed arrays.

To use a fixed array in `Calldata.compile`, you have to use `cairo.fixedArray()`:
```ts
const myArray = [1, 2, 3];
const myCallData = CallData.compile([CairoFixedArray.compile(myArray)]);
```
`CairoFixedArray.compile()` can be used in functions using the ABI, but it's useless.

A new CairoFixedArray, with both static & dynamic  :
- `isTypeFixedArray()`
- `getFixedArraySize()`
- `getFixedArrayType()`
- `compile()`

All functions related to encoding/decoding have been modified to handle this new type.

A dedicated  Cairo test contract has been added.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


PhilippeR26:
Implemented.
Not sure (I never worked in it), but could be impacts in Abiwan library.

penovicp (approved):

PR #1319, opened 2025-02-11
docs: add X feed to website

PhilippeR26:
## Motivation and Resolution
Improvements of the website.

## Usage related changes
- favicon replaced by starknet.js logo
- title replaced by `Starknet.js docs`
- added on the right of the landing page : X feed

- Helmet lib added in the package.json (to handle properly external scripts)
- The script has been generated by : https://publish.twitter.com/

## Checklist:


www/src/pages/index.tsx
```diff
@@ -20,6 +21,10 @@ function HomepageHeader() {
           </Link>
         </div>
       </div>
+      <a className="twitter-timeline" data-width="250" data-height="500" data-dnt="true" data-theme="dark" href="https://twitter.com/starknetjs?ref_src=twsrc%5Etfw">Tweets by Starknetjs</a>
```
tabaktoni, line 24: this shift the entire content left. I would suggest fixing it rigt or placing it in the separate container below the header

www/src/pages/index.tsx
```diff
@@ -28,7 +33,7 @@ export default function Home(): JSX.Element {
   const { siteConfig } = useDocusaurusContext();
   return (
     <Layout
-      title={`Hello from ${siteConfig.title}`}
+      title={``}
```
tabaktoni, line 36: looks like you removed the title completely

tabaktoni (changes requested):

www/src/pages/index.tsx
tabaktoni: The problem with fixing it right is that on mobile I will cover the entire screen, so I think best solution is to place it in container like this 3 row container bellow header

www/src/pages/index.tsx
PhilippeR26: No, it was in fact in double.
image
I just kept one:
image

www/src/pages/index.tsx
PhilippeR26: The difficulty is that it sticks on the right border. I will see what can be done.

PhilippeR26: It was a bit painful, but it's relocated below the header, and is displayed correctly on narrow screens.

www/docusaurus.config.js
```diff
@@ -15,7 +15,7 @@ const migrationGuideLink = `${generateBaseUrl(process.env.DOCS_BASE_URL)}docs/gu
 
 /** @type {import('@docusaurus/types').Config} */
 const config = {
-  title: 'Starknet.js',
+  title: 'Starknet.js docs',
```
penovicp, line 18: I would keep this as `Starknet.js`.

www/src/pages/index.tsx
```diff
@@ -22,17 +23,56 @@ function HomepageHeader() {
...
+  return (
+    <div
+      className={clsx('hero hero--primary', styles.heroBanner)}
+      style={{
+        height: height,
+        padding: 0,
+      }}
+    >
+      <div
+        style={{
```
penovicp, line 42: Looks like this could use a `maxWidth`, images in the posts get noticeably stretched without it.

www/src/pages/index.tsx
penovicp: To add a bit more context, setting this to an empty string isn't an issue since this is just the page title of the index page. All the pages get a title generated as `page title | site title` and it falls back to just the site title if the page title is empty.

www/docusaurus.config.js
PhilippeR26: @penovicp  proposed `Starknet.js`.
@tabaktoni proposed `Starknet.js docs`
Personaly, I don't care. Just tell me your final choice.

www/src/pages/index.tsx
PhilippeR26: ????
I don't understand.
Have you a screenshot showing these stretched images?

PhilippeR26:
I added a commit, due to problems with `npm run build` of Docusaurus SSR rendering.
As the `XFeed` component is using frontend `window` object, it's necessary to use the Docusaurus `BrowserOnly` component to wrap the frontend code (so the code using `window` object).
See https://docusaurus.io/docs/docusaurus-core/#browseronly

www/docusaurus.config.js
penovicp: I don't think the suggestions are in conflict. To reference my other comment https://github.com/starknet-io/starknet.js/pull/1319#discussion_r1954178678 for the `www/src/pages/index.tsx` change: 
>To add a bit more context, setting this to an empty string isn't an issue since this is just the page title of the index page. All the pages get a title generated as page title | site title and it falls back to just the site title if the page title is empty.

I believe Toni wanted the page title segment to be changed to reduce redundancy (which is done with the `www/src/pages/index.tsx` change) and this changes the site title segment so keeping it as `Starknet.js` makes sense to me.

www/src/pages/index.tsx
PhilippeR26: My problem is that on my PC, `npm run start` or `npm run serve`  are not displaying the X script correctly (as show in the Discord channel).
Have you a way to display correctly the Xfeed?
About the image stretching, I suppose that it's the behavior decided by the X dev of the script.
I tried to change the code about width, but as explained, I have no idea of the result.

www/docusaurus.config.js
PhilippeR26: Modified.

www/src/pages/index.tsx
penovicp: Sorry I missed this question when you posted it. It should be enough to be logged in with the current browser session for the feed to display. 

I merged the PR into beta after I made some tweaks to the appearance to account for the image stretch and the widget not loading.

PR #1323, opened 2025-02-13
fix spelling error

Vitaliyr888:
fixed:

nor - or

Sign an JSON - Sign a JSON(2x)


tabaktoni:
Old verson of the documentation, irelevant as it will be removed in next release anyway

PR #1328, opened 2025-02-18
Prepare beta v7

tabaktoni:
## Motivation and Resolution
Starknet 0.13.4

 RPC 0.8

## Usage related changes

- [ ] Changes on provider types (do tests and checks on Provider)

Some test captured BreakingChanged:
- [x] new feeMarginPercentage type
- [x]  removed enum FeeMarginPercentage from constants
- [x] added new FeeMarginPercentage to DEFAULT_GLOBAL_CONFIG
- [x] Transaction Receipt Typing refactor, also removed non-existing isRejected status 

When specVersion is not defined in constructor options, prefer usage of the `provider = await RpcProvider.create()` for safe spec version/ channel resolution based on the node. Else it fail back to config default rpc version that could be false based on node rpc.

- [x] Replace internal usage of default exported api types (in this case 0.8)
- [x] Consider disabling default export api (in this case 0.8)
- [x] Use merged types in all cases except channels

## Testing

- [x] backward compatibility RPC 0.7 V2
- [x] backward compatibility RPC 0.7 V3_old
- [x] compatibility RPC 0.8 V3_new

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


__tests__/utils/errors.test.ts
```diff
@@ -2,7 +2,8 @@ import { RPC, RpcError } from '../../src';
 
 describe('Error utility tests', () => {
   test('RpcError', () => {
-    const baseError: RPC.Errors.UNEXPECTED_ERROR = {
+    // TODO: @penovicp types 0.8 exporta u top level sve 0.7 exporta po komponentama, neznan jeli treba minjat test i mod da ocukuje oba tipa
```
penovicp, line 5: The type export change shouldn't be an issue. The RPC error utility relies on the mappings set up in `src/types/errors.ts`, so when the mapping is updated the utility should work regardless of how the underlying types are imported. It should work even without an update since the 0.7 errors are a subset of the 0.8 errors, the difference is that without it the new 0.8 errors wouldn't be recognised for type narrowing.

penovicp (approved):

PR #1334, opened 2025-02-28
Replace `#[external(v0)]` with `#[abi(embed_v0)]` to fix deprecation warning

Himess:
## Motivation and Resolution
The `#[external(v0)]` attribute has been deprecated in favor of `#[abi(embed_v0)]` or `#[abi(per_item)]`, as indicated by the following warning during compilation:

Pasted Graphic

## Usage related changes

Users no longer see the deprecation warning related to `#[external(v0)]` during compilation.
Contracts now use `#[abi(embed_v0)]` as per the latest Cairo/Starknet standards.

Removed deprecated `#[external(v0)]` and replaced it with `#[abi(embed_v0)]`.
Ensured that all modified files still compile successfully with `scarb build`.
Verified that no additional errors or warnings appear.
Confirmed that this change aligns with similar updates in the repository.

## Test Result

hinesseSenih-MacBook-Air test virgul   scarb build

### DIFF
--- a/__mocks__/cairo/cairo210/cairo210.cairo
+++ b/__mocks__/cairo/cairo210/cairo210.cairo
@@ -13,7 +13,7 @@ mod MyTestFelt {
         counter: u128,
     }
 
-    #[external(v0)]
+    #[abi(embed_v0)]
     impl TestFelt of super::ITestC210<ContractState> {
         fn test_felt(self: @ContractState, p1: felt252, p2: u128, p3: u8) -> u128 {
             p2 + 1


tabaktoni:
@PhilippeR26 is this legit ?

PhilippeR26:
That's right that it has evolved, but only for recent Cairo versions ; exemple in a recent test code : 
https://github.com/starknet-io/starknet.js/blob/d30e1302c927844b6cd5326f1ad6721978644b69/__mocks__/cairo/cairo292/fixed_array.cairo#L16
In the test suite, we are using Cairo code of varied versions, including Cairo 0 code.
In this example, as written in the comment in the first line, it's coded for Cairo v2.1.0. 
If I compile this code with Cairo 2.1.0, there is no warning or errors ; this code is valid for this version.
image
image
To ensure the compatibility of Starknet.js with all versions of Cairo (Cairo 0, Cairo 1, Cairo 2.0.0 to v2.11.0), we are testing codes compiled with all these versions. 
To be able to compile all the Cairo test codes with the last compiler is not an objective, and is even not desirable to be able to check the compatibility with all the history of Cairo implemented in the Starknet blockchain.
So, based on this explanation of the objectives of these Cairo test contracts, I think we should consider to not implement this proposal.

Himess:
@PhilippeR26 Yes, I see that I overlooked that detail. I understand now that this is not actually an error. Feel free to close the PR if needed.

PR #1360, opened 2025-03-24
feat: prune deprecated functionalities

penovicp:
## Motivation and Resolution

Removes long-standing deprecated functionalities and aliases.

Notes:
- the `ProviderInterface. invokeFunction()` deprecation seems to have been based on traits stemming from before the RPC support was introduced, it seems to have been made obsolete by the introduction of the `details` parameter - it is internally used so just the deprecation mark is removed
- the `WalletAccount` change could be made stricter by making the constructor private and introducing an optional `address` parameter to the `connect` methods to potentially skip the address retrieval

### DIFF
--- a/__tests__/account.starknetId.test.ts
+++ b/__tests__/account.starknetId.test.ts
@@ -46,39 +46,36 @@ describe('deploy and test Wallet', () => {
     });
     multicallAddress = multicallResponse.deploy.contract_address;
 
-    const { transaction_hash } = await account.execute(
-      [
-        {
-          contractAddress: devnetERC20Address,
-          entrypoint: 'approve',
-          calldata: [namingAddress, 0, 1], // Price of domain
-        },
-        {
-          contractAddress: identityAddress,
-          entrypoint: 'mint',
-          calldata: ['1'], // TokenId
-        },
-        {
-          contractAddress: namingAddress,
-          entrypoint: 'buy',
-          calldata: [
-            '1', // Starknet id linked
-            '1499554868251', // Domain encoded "fricoben"
-            '62', // days
-            '0', // resolver
-            0, // sponsor
-            0,
-            0,
-          ],
-        },
-        {
-          contractAddress: identityAddress,
-          entrypoint: 'set_main_id',
-          calldata: ['1'],
-        },
-      ],
-      undefined
-    );
+    const { transaction_hash } = await account.execute([
+      {
+        contractAddress: devnetERC20Address,
+        entrypoint: 'approve',
+        calldata: [namingAddress, 0, 1], // Price of domain
+      },
+      {
+        contractAddress: identityAddress,
+        entrypoint: 'mint',
+        calldata: ['1'], // TokenId
+      },
+      {
+        contractAddress: namingAddress,
+        entrypoint: 'buy',
+        calldata: [
+          '1', // Starknet id linked
+          '1499554868251', // Domain encoded "fricoben"
+          '62', // days
+          '0', // resolver
+          0, // sponsor
+          0,
+          0,
+        ],
+      },
+      {
+        contractAddress: identityAddress,
+        entrypoint: 'set_main_id',
+        calldata: ['1'],
+      },
+    ]);
 
     await provider.waitForTransaction(transaction_hash);
   });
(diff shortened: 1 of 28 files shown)


PhilippeR26 (commented):
My understanding:
1. `number` namespace removed. Use existing `num` namespace
2. Usage of `new WalletAccount(...)` is removed. Use existing `WalletAccount.connect(...)` or `WalletAccount.silentConnect(...)`
3. `UINT_128_MAX, UINT_256_MAX` are no more available in `uint256` namespace. Use existing direct import `import { UINT_256_MAX } from "starknet";`
4. `TypedDataRevision, StarknetEnumType, StarknetMerkleType, StarknetType, StarknetDomain, TypedData` are no more available in `typeData` namespace. Use existing direct import `import { StarknetDomain } from "starknet";`
5. In `transaction` namespace, functions `fromCallsToExecuteCalldataWithNonce(), transformCallsToMulticallArrays_cairo1()` are removed.
6. **`BigNumberish` is no more available in `num` namespace. Use existing direct import `import { BigNumberish } from "starknet";`**
7. `stringifyAlwaysAsBig()` is no more available in `json` namespace. Use existing `json.stringify()`
8. `stringToArrayBuffer()` is no more available in `encode` namespace. Use existing `encode.utf8ToArray()`
9. In `stark` namespace, function `makeAddress()` is removed.
10. `CallStruct` type is removed.
11. **`getEstimatefee()` method of `ProviderInterface` and of `RpcProvider` is removed. Use `myAccount.estimateInvokeFee()` or `myAccount.estimateDeclareFee()`**
12. `getPendingTransactions()` method of `RpcProvider` is removed. Use `myProvider.getBlockWithTxHashes(BlockTag.PENDING)`
13. **In `Account.execute()`, the `abi` parameter is removed in `execute(transactions: AllowArray<Call>, abis?: Abi[], transactionsDetail?: InvocationsDetails)`. Use only `myAccount.execute(transactions: AllowArray<Call>, transactionsDetail?: InvocationsDetails)`**
14. In `Account` class, function `verifyMessageHash` is removed. Use existing `myProvider.verifyMessageInStarknet()`
15. In `Account` class, function `verifyMessage` is removed. Use existing `myProvider.verifyMessageInStarknet()`

Items 6 could be problematic.
Changes of `Account` or `RpcProvider` interfaces are always risky (event if removal is less problematic that adds).
I think I will communicate with the community to see if I have complains about these changes.

PS: The deprecated tag has been removed from `Provider.invokeFunction()`. Why this tag is no more valid?

penovicp:
@PhilippeR26 When I set out to remove the deprecations my plan was to only remove long-standing ones, it's just a coincidence that it ended up being all of them. All are at least half a year old, with most being over a year (e.g. the `number` to `num` rename is part of the [v5 migration doc](https://starknetjs.com/docs/5.24.3/guides/migrate/#number-utility-replaced-by-num)), the only exception is the `WalletAccount` `address` as an optional parameter but that one explicitly mentions removal in the next major version.

I'll do a pass and check which deprecation notifications might not have been properly propagated, I'm guessing this might be the case for the type re-exports, but most deprecations should have already been clearly communicated through the deprecation annotations. I'll update the migration doc accordingly.

### For some of the specific points:

- The `WalletAccount` change still allows `new WalletAccount(...)` to be used it just makes the `address` parameter mandatory, so `WalletAccount.connect(...)` is for regular usage when the address is retrieved with a request and `new WalletAccount(...)` if for some reason the address is already known and the request wants to be skipped. I mentioned in the description of this PR that the change could be made stricter by making the constructor private and adding an optional `address` parameter to the `connect` methods but that adjustment isn't included in the PR (at least not currently).

- I mentioned the `ProviderInterface.invokeFunction()` difference in the PR description. To go into a bit more detail, [this is when it was introduced](https://github.com/starknet-io/starknet.js/pull/124/files#diff-a15cf25fca376d25e161dc1502509193687979d881961237e2d4b3fc4b996d8eR128) with an emphasis on fees. I believe the deprecation annotation was added in anticipation that the corresponding underlying method would be replaced with something completely different when transitioning from the Sequencer to the RPC API, but it ended up being replaced with a similar endpoint expanded with the details parameter that has fee bounding properties. With that assumption and considering that `invokeFunction` is still used in another part of the code I reckoned that the deprecation was no longer warranted.

tabaktoni (approved):
Just add to changes list of this

PR #1361, opened 2025-03-25
Remove fetch dependencies

penovicp:
## Motivation and Resolution

Removes `isomorphic-fetch` and `fetch-cookie` as production dependencies to reduce the number of dependencies and simplify integration.

## Usage related changes

For `isomorphic-fetch`, most users should not be affected since a fetch fallback should no longer be necessary in the majority of environments. Users with non-browser environments who need `fetch-cookie` will need to import it into their projects separately. 

A [`baseFetch` ](https://github.com/starknet-io/starknet.js/blob/v6.23.1/src/types/provider/configuration.ts#L16)override has already been enabled for the `RpcProvider` and `RpcChannel` classes through which both libraries can be used in conjunction with starknet.js.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


PhilippeR26 (approved):
Hopping it will work in all contexts.
But we had enough problems with these libs in the past. Happy to see them go away.

tabaktoni (approved):

PR #1369, opened 2025-04-01
feat: new external signer to handle external signer provider such as Dfns

fondation451:
## Motivation and Resolution

Before, there was only a default signer with direct access to the private key. I have been using wallet provider such as Dfns which provides signing functions as a service. I derived from the implementation of the default signer, a signer that rely on an external service. I think that it may be useful to have it in the library.

## Usage related changes

- Introduction of a new signer to use external signing service such as Dfns

- It is based on the default signer. All the method that needs the private key are overwritten to use the external service

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/signer/externalSigner.ts
+++ b/src/signer/externalSigner.ts
@@ -1,0 +1,23 @@
+import { Signature } from '../types';
+import { Signer } from './default';
+import { SignerInterface } from './interface';
+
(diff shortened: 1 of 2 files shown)


tabaktoni:
bump

adrienlacombe:
@fondation451 could you please add tests and doc 🙏

PR #1370, opened 2025-04-03
fix: make the provider create method generic to enable it when mixins are applied

penovicp:
## Motivation and Resolution

Ensures that `RpcProvider.create(...)` generates the chosen class with the appropriate TypeScript type for both the base class and extended classes after mixins are added. Also disables it for the `Account` class inheritance chain.

## Checklist:

- [x] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/__tests__/rpcProvider.test.ts
+++ b/__tests__/rpcProvider.test.ts
@@ -1,24 +1,5 @@
 import { getStarkKey, Signature, utils } from '@scure/starknet';
-import typedDataExample from '../__mocks__/typedData/baseExample.json';
-import {
-  Account,
-  Block,
-  CallData,
-  Contract,
-  FeeEstimate,
-  ProviderInterface,
-  RPC,
-  RPCResponseParser,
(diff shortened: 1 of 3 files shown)


penovicp:
> Nice: image
> 
> We should take the opportunity when solving this type problem, to also solve this known problem: image


`SupportedRpcVersion` from the second image is an object literal emulation of a TypeScript enum. The error prompt shown is a side-effect of reproducing the enum behaviour.

enum

PhilippeR26:
So, how a user can use `SupportedRpcVersion.xxx` as a constant **value**?
(I suppose that if it's located in `constants` namespace, it's to be used as constant by users...)

penovicp:
@PhilippeR26 After your last message I remembered that what you were actually referencing was that `SupportedRpcVersion` has an issue with the underlying object literal not being visible to TS in the published build. I believe I found an approach that resolves the issue, it's in the #1375 draft PR.

PR #1374, opened 2025-04-07
feat: remove WebSocket dependencies

penovicp:
## Motivation and Resolution

The most widely used browsers and the most recent Node.js LTS version provide native WebSocket support. This means that for most users the fallback modules should be unnecessary.

## Usage related changes

Most users should be unaffected. Users working with environments without native WebSocket support can use the `websocket` parameter of the `WebSocketChannel` constructor to provide a valid WebSocket connection themselves.

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


src/utils/connect/fetch.ts
```diff
@@ -0,0 +1,10 @@
+import { LibraryError } from '../errors';
+
+export default (typeof fetch !== 'undefined' && fetch) ||
+  (typeof window !== 'undefined' && window.fetch) ||
+  (typeof global !== 'undefined' && global.fetch) ||
```
aryzing, line 5:
```suggestion
  (typeof globalThis.fetch !== 'undefined' && globalThis.fetch) ||
```

Otherwise a discrete "global object" check per environment needs to be added. `globalThis` is the more modern cross-environment option.

src/utils/connect/fetch.ts
ondrej-secretkeylabs, line 1: Has this been tested in browsers?

I get `Failed to execute 'fetch' on 'Window': Illegal invocation`

aryzing: Seems it's due to the missing `.bind(window)` as in:

https://github.com/starknet-io/starknet.js/blob/960981c1d10822777d1442419f9510ae4f07b38b/src/utils/fetch.ts#L5

For @penovicp : do you plan on supporting non-fetch-supporting environments? Can we just get rid of these checks that add complexity? All modern environments support native fetch.

tabaktoni: I don't see why this is mandatory but will add window.fetch.bind(window)

tabaktoni (approved):

src/utils/connect/fetch.ts
aryzing: https://stackoverflow.com/a/69876888

src/utils/connect/fetch.ts
```diff
@@ -0,0 +1,11 @@
+import { LibraryError } from '../errors';
+
+export default (typeof fetch !== 'undefined' && fetch) ||
+  (typeof globalThis !== 'undefined' && globalThis.fetch) ||
```
aryzing, line 4:
```suggestion
  (typeof globalThis !== 'undefined' && globalThis.fetch.bind(globalThis) ||
```

Similar to `window` case on the next line. E.g., browser service workers support `fetch` and `globalThis`, but don't have `window`.

aryzing, line 4:
```suggestion
export default (typeof globalThis !== 'undefined' && globalThis.fetch.bind(fetch)) ||
```

The first check is equivalent to the second, but may be confusing to do it without checking for globalThis and then binding to it.

What works and what doesn't:

Works:

```ts
const myFetch = fetch;
await myFetch("https://example.com"); // ok
```

Does not work: `Uncaught (in promise) TypeError: Failed to execute 'fetch' on 'Window': Illegal invocation`

```ts
const myObj = {
  myFetch,
  fetch,
};

myObj.myFetch("https://example.com"); // error
myObj.fetch("https://example.com"); // error
```

Why? the `this` value in this second case is `myObj`.

Why does this fail with `starknet`? The key areas of the library's compiled code look like this:

```ts
// src/utils/connect/fetch.ts
var fetch_default =
  (typeof fetch !== "undefined" && fetch) ||
  (typeof globalThis !== "undefined" && globalThis.fetch) ||
  (typeof window !== "undefined" && window.fetch.bind(window)) ||
  (typeof global !== "undefined" && global.fetch) ||
  (() => {
    throw new LibraryError(
      "'fetch()' not detected, use the 'baseFetch' constructor parameter to set it"
    );
  });
```

and

```ts
this.baseFetch = baseFetch || config.get("fetch") || fetch_default;
```

with a random invocation looking like,

```ts
this.baseFetch(this.nodeUrl, {
  method: "POST",
  body: stringify2(rpcRequestBody),
  headers: this.headers,
});
```

Above, `this` is `RpcChannel`, not `window` or `globalThis`, so the call fails.

src/utils/connect/fetch.ts
victorkirov: We're having build issues due to the `global` reference not being defined. Trying out this branch, the build is still failing. We also tried removing the `global` reference but the library stopped functioning due to the bind being skipped. Here are our findings on the version that works:

```typescript
export default (
  // Fetch is defined on a global scale, so window.fetch === globalThis.fetch === fetch
  // This means that this first if statement always resolves to `fetch` but without binding
  // to globalThis, so the moment this library tries to make a fetch call, it throws an error
  typeof fetch !== 'undefined' && fetch) ||

  // This line is equivalent to the above line since fetch === globalThis.fetch
  // but is never hit since the above line would always resolve to 'fetch' 
  (typeof globalThis !== 'undefined' && globalThis.fetch) ||

  // Removing the top 2 lines makes the Starknet library work in web but fail in node
  // since window is defined in web and is being bound to the window object correctly here
  (typeof window !== 'undefined' && window.fetch.bind(window)) ||

  // This line completely breaks the build as there is no `global` variable defined
  // in web or NodeJS. `global` would effectively be `globalThis.global` which doesn't exist,
  // so it would be best to remove this line completely to avoid getting 
  // `Uncaught ReferenceError: global is not defined` in build and run time when this code block is hit
  (typeof global !== 'undefined' && global.fetch) ||

  ((() => {
    throw new LibraryError(
      "'fetch()' not detected, use the 'baseFetch' constructor parameter to set it"
    );
  }) as WindowOrWorkerGlobalScope['fetch']);
```

With the above in mind, since `fetch === window.fetch === globalThis.fetch` and in the web context `window === globalThis` but in NodeJS `window` doesn't exist but `globalThis` does, the best solution for extracting and binding fetch is just to do the below. This code snippet was tested in a web and NodeJS context and works as expected. 
```typescript
export default (globalThis?.fetch && globalThis.fetch.bind(globalThis)) ||
  ((() => {
    throw new LibraryError(
      "'fetch()' not detected, use the 'baseFetch' constructor parameter to set it"
    );
  }) as WindowOrWorkerGlobalScope['fetch']);
```

victorkirov: @tabaktoni it's also needed on globalThis and global for the same reason 🙏 
```typescript
  (typeof globalThis !== 'undefined' && globalThis.fetch.bind(globalThis)) ||
  (typeof window !== 'undefined' && window.fetch.bind(window)) ||
  (typeof global !== 'undefined' && global.fetch.bind(global)) ||
```

and this first line should not exist as it will override all the others as `fetch === globalThis.fetch === window.fetch === global.fetch`, so this line would always win over the others and the binding won't happen:
```typescript
(typeof fetch !== 'undefined' && fetch) ||
```

src/utils/connect/fetch.ts
tabaktoni: Ok, but why are you suggesting removing support for node.js fetch?

victorkirov: @tabaktoni plz see [this](https://github.com/starknet-io/starknet.js/pull/1374#discussion_r2081163721) comment on why the global `fetch` without a context should be removed. As per the comment, the bind also needs to be applied to `global` and `globalThis` or they will break in the same way that `window` broke.

aryzing: @tabaktoni, the changes suggested in the comment above don't remove support for Node.js fetch. If applied, as also suggested by @victorkirov  above, the final result would look like 

```ts
  (typeof globalThis !== 'undefined' && globalThis.fetch.bind(globalThis)) ||
  (typeof window !== 'undefined' && window.fetch.bind(window)) ||
  (typeof global !== 'undefined' && global.fetch.bind(global)) ||
```

- globalThis --> all modern clients
- window --> legacy browsers
- global --> legacy node.js

PR #1375, opened 2025-04-08
fix: ensure enum replacement objects are reachable for type resolution

penovicp:
## Motivation and Resolution

Completes the transition from TS enums into their companion pattern replacements. Also repairs the reachability of the `SupportedRpcVersion` object that was being hidden by the type in the distribution build.

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/global/constants.ts
+++ b/src/global/constants.ts
@@ -90,6 +90,8 @@ export { _TransactionHashPrefix as TransactionHashPrefix };
 const _SupportedRpcVersion = {
   0.7: '0.7',
   0.8: '0.8',
+  v07: '0.7',
+  v08: '0.8',
 } as const;
 type _SupportedRpcVersion = ValuesType<typeof _SupportedRpcVersion>;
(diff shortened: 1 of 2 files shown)


PhilippeR26:
The `SupportedRpcVersion` properties are difficult to reach :
`SupportedRpcVersion.0.7` is not valid, due to the point in the property name.
Maybe to replace by `0_7` and `0_8`...

penovicp:
> The `SupportedRpcVersion` properties are difficult to reach : `SupportedRpcVersion.0.7` is not valid, due to the point in the property name. Maybe to replace by `0_7` and `0_8`...

For dot notation to work the key would need to start with a letter, underscore, or dollar sign, so I went with `v07`.

tabaktoni (approved):

PhilippeR26 (approved):
We have both in vscode, but it's OK.
image

PR #1381, opened 2025-04-17
docs: remove consecutive duplicate words

zeroprooff:


PhilippeR26:
Hello,
The first word after {...} is the name of the function.
Then is written the description.
https://jsdoc.app/tags-param
So, in
```ts
* @param {UniversalDetails} details details of the transaction
```
the first `detail` is the name of the parameter, and the 2nd `detail` is the beginning of the description.
Conclusion: the current code is correct.
So, this PR should not be merged.

PR #1384, opened 2025-04-25
feat: full speck version implementation

tabaktoni:
## Motivation and Resolution

...

...

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/channel/rpc_0_7_1.ts
+++ b/src/channel/rpc_0_7_1.ts
@@ -32,7 +32,14 @@ import fetch from '../utils/fetch';
 import { getSelector, getSelectorFromName } from '../utils/hash';
 import { stringify } from '../utils/json';
 import { getHexStringArray, toHex, toStorageKey } from '../utils/num';
-import { Block, getDefaultNodeUrl, isV3Tx, wait } from '../utils/provider';
+import {
+  Block,
+  getDefaultNodeUrl,
+  isSupportedSpecVersion,
+  isV3Tx,
+  isVersion,
+  wait,
+} from '../utils/provider';
 import { decompressProgram, signatureToHexArray } from '../utils/stark';
 import { getVersionsByType } from '../utils/transaction';
 import { logger } from '../global/logger';
@@ -44,7 +51,12 @@ const defaultOptions = {
 };
 
 export class RpcChannel {
-  readonly id = 'RPC07';
+  readonly id = 'RPC071';
+
+  /**
+   * This Channel provide communication using spec version
+   */
+  readonly channelSpecVersion: SupportedRpcVersion = '0.7.1';
 
   public nodeUrl: string;
 
@@ -60,6 +72,9 @@ export class RpcChannel {
 
   private chainId?: StarknetChainId;
 
+  /**
+   * Connected RPC node specification version
+   */
   private specVersion?: SupportedRpcVersion;
 
   private transactionRetryIntervalFallback?: number;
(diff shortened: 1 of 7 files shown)


src/provider/rpc.ts
```diff
@@ -100,14 +105,27 @@ export class RpcProvider implements ProviderInterface {
...
-      return new this({ ...optionsOrProvider, specVersion: '0.8' }) as T;
+      return new this({
+        ...optionsOrProvider,
+        specVersion: SupportedRpcVersion.v0_8_1,
+      }) as T;
     }
 
-    throw new LibraryError('Unable to detect specification version');
+    throw new LibraryError(
+      `Provided RPC node specification versions ${spec} is not compatible with the SDK. SDK supported RPC versions ${Object.keys(SupportedRpcVersion).toString()}`
```
penovicp, line 127:
```suggestion
      `Provided RPC node specification version ${spec} is not compatible with the SDK. SDK supported RPC versions: ${Object.keys(SupportedRpcVersion).toString()}`
```

src/channel/rpc_0_7_1.ts
```diff
@@ -44,7 +45,12 @@ const defaultOptions = {
 };
 
 export class RpcChannel {
-  readonly id = 'RPC07';
+  readonly id = 'RPC071';
+
+  /**
+   * This Channel provide communication using spec version
```
penovicp, line 51:
```suggestion
   * RPC specification version this Channel class implements
```
This is a suggestion to sync the typedoc note with `this.specVersion`.

src/channel/rpc_0_7_1.ts
```diff
@@ -60,6 +66,9 @@ export class RpcChannel {
 
   private chainId?: StarknetChainId;
 
+  /**
+   * Connected RPC node specification version
```
penovicp, line 70:
```suggestion
   * RPC specification version of the connected node
```
This is a suggestion to sync the typedoc note with `this. channelSpecVersion `.

src/channel/rpc_0_7_1.ts
```diff
@@ -180,25 +196,37 @@ export class RpcChannel {
...
   }
 
   /**
-   * fetch spec version in extended format "M.m.p-?"
+   * fetch if undefined test and set specVersion, else just return this.specVersion
+   * @example this.specVersion = "0.7.1"
    */
-  public getSpecificationVersion() {
-    return this.fetchEndpoint('starknet_specVersion');
+  public async setupSpecVersion() {
```
penovicp, line 210: Nitpick, `setup` is usually used as a noun and `set up` as the corresponding verb form so the method would be `setUpSpecVersion`.

penovicp (approved):
A couple of rewording suggestions, otherwise lgtm

PhilippeR26 (commented):
So, these codes valid for Starknet.js v7.1.0 are no more valid:
`const myProvider = new RpcProvider({ nodeUrl: "https://free-rpc.nethermind.io/sepolia-juno/v0_7", specVersion: "0.7" }); ` 
`console.log( "rpc =", await myProvider.getSpecVersion()`
What's the motivation of this breaking change? Has someone requested it?
Isn't a patch number increase allowed only for non-breaking changes?
If the patch number become discriminant, is Starknet.js v7 compatible with rpc 0.7.0 & 0.7.1? with 0.8.0 & 0.8.1? 
Or only with 0.7.1 & 0.8.1? Will other patch versions be rejected when a`RpcProvider` will be instantiated?
We asked to Lava network to discriminate major.minor nodes versions. What will append with this PR?

tabaktoni:
As described isVersion has been refactored to work with a wildcard, so  during dev expectation is:
`isVersion('0.7', spec)` it's the same as sVersion('0.7.*', spec) or any 0.7.x version is true.
The only breaking change is the actual spec 'enums', but if they are used as consts it should not break as it will be updated in client code to the new value.

> What's the motivation of this breaking change? Has someone requested it?
Yes, SW to detect patch Spec incompatibility. As spec versioning is described as major.major.major whitch means all subversions are breaking, some more, some less. So the only logical solution is to have full version dininition in the code.

> Isn't a patch number increase allowed only for non-breaking changes?
Yes, but for starknetjs we are in pre-release and haven't done v7 main release yet for exactly that reason. Once we do, we can't introduce any more breaking changes before v8. For spec, it is as described above, but it is expected even do I would argue spec should only have major and minor.

> If the patch number becomes discriminant, is Starknet.js v7 compatible with rpc 0.7.0 & 0.7.1? with 0.8.0 & 0.8.1?
There are only warnings for patch mismatches and errors on major mismatches.

PR #1387, opened 2025-04-28
Fix crash on ReferenceError when run in service worker

### LINKED ISSUE #1385: Wrong check for global being undefined crashes starknet.js whenever run in Service Worker
**Describe the bug**
The code currently contains this:

```
(!isUndefined(global) && global.fetch) || // use built-in fetch in node, react-native and service worker if available
```

This code is run if window.fetch is not defined. If global is also _not_ defined, ReferenceError is thrown because an undeclared variable cannot be passed to a function (in this case isUndefined).

**To Reproduce**
Add the code from src/utils/fetch.ts to any service worker.

**Additional context**
I would advise against the usage of isUndefined in the rest of the codebase for this reason.

**Proposed fix**

https://github.com/ondrej-secretkeylabs/starknet.js/tree/fix/undefined-check-crash

As far as I know, whenever fetch is defined in global or window, it should also be defined in the truly-global scope. We can check for this easily with `typeof fetch === 'function'`.

ondrej-secretkeylabs:
## Motivation and Resolution

Resolves #1385 .

## Usage related changes

None

Removes extraneous check for IS_BROWSER as fetch being defined in the window (or there even being a window object) implies its global availability.

## Checklist:

- [ ] Updated the tests
- [x] All tests are passing

### DIFF
--- a/src/utils/fetch.ts
+++ b/src/utils/fetch.ts
@@ -1,9 +1,6 @@
-import { IS_BROWSER } from './encode';
 import { LibraryError } from './errors';
-import { isUndefined } from './typed';
 
-export default (IS_BROWSER && window.fetch.bind(window)) || // use built-in fetch in browser if available
-  (!isUndefined(global) && global.fetch) || // use built-in fetch in node, react-native and service worker if available
+export default (typeof fetch === 'function' && fetch) ||
   // throw with instructions when no fetch is detected
   ((() => {
     throw new LibraryError(


tabaktoni:
I agree with the issue for isUndefined. 
Also, we need to expand this module to cover existing fetches, not just completely drop all.
I think the best solution will be to rewrite the module with multiple try-catches and at the end throw or default export depending on the result. @PhilippeR26 @penovicp

PhilippeR26:
> I agree with the issue for isUndefined. Also, we need to expand this module to cover existing fetches, not just completely drop all. I think the best solution will be to rewrite the module with multiple try-catches and at the end throw or default export depending on the result. @PhilippeR26 @penovicp

My understanding is: for Starknet.js v7 we removed the fetch libraries, as they generated various problems, and because modern versions of node.js & browsers are all implementing their own `fetch` functionality.
And for specific cases that are not handling their own `fetch`, a solution is explained in the [doc](https://starknetjs.com/docs/next/guides/migrate#fetch-dependencies).
Would you like to cover some specific cases (as this service worker) without using the 'trick' of the doc?

penovicp:
@tabaktoni The WebSocket dependency PR #1374  includes a `fetch` refactor that removes `isUndefined` and preserves the existing global checks. I recommend merging that PR in favour of this one.

@PhilippeR26 You are correct that using a `fetch` override would avoid the reported problem. The reason the `fetch` handling should be changed to avoid encapsulating the `typeof` check with `isUndefined` is to ensure the conditional falls through to the final "fetch not detected" error, and not the current behaviour where it terminates early with a reference error.

penovicp (changes requested):
I believe the fetch refactor in #1374 makes this PR not necessary.

ondrejsojka:
Great, that's better. Will there be a release with the WebSocket updates?

ondrej-secretkeylabs:
Closing in favor of #1374

PR #1399, opened 2025-05-16
fix: parseEvents is not working twice

PhilippeR26:
## Motivation and Resolution
`myContract.parseEvents()` is working once, but not twice.
For example this code is not working:
```typescript
const txR: GetTransactionReceiptResponse = await myProvider.getTransactionReceipt(txH);
const events2 = myContract.parseEvents(txR);
const events3 = myContract.parseEvents(txR);
```
`events2` is OK, but `events3` is `[]`

Origin of the problem :
https://github.com/starknet-io/starknet.js/blob/050616221a58545d79eece7040b4abc205c220e0/src/utils/events/index.ts#L205-L210
`keys.shift()` is not local, and is executed in fact in the initial `txR`.

## Usage related changes
Bug Solved.

A deep copy is executed before use of `.shift()`:
https://github.com/PhilippeR26/starknet.js/blob/2e6343c90380a99f6c96c2030c0dd5aa8bdf8fab/src/utils/events/index.ts#L205

## Checklist:

- [ ] All tests are passing

### DIFF
--- a/src/utils/events/index.ts
+++ b/src/utils/events/index.ts
@@ -202,12 +202,13 @@ export function parseEvents(
   const ret = providerReceivedEvents
     .flat()
     .reduce((acc, recEvent: RPC.EmittedEvent | RPC.Event) => {
-      let abiEvent: AbiEvent | AbiEvents = abiEvents[recEvent.keys.shift() ?? 0];
+      const currentEvent: RPC.EmittedEvent | RPC.Event = JSON.parse(JSON.stringify(recEvent));
+      let abiEvent: AbiEvent | AbiEvents = abiEvents[currentEvent.keys.shift() ?? 0];
       if (!abiEvent) {
         return acc;
       }
       while (!abiEvent.name) {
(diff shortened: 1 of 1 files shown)


src/utils/events/index.ts
```diff
@@ -202,21 +202,22 @@ export function parseEvents(
   const ret = providerReceivedEvents
     .flat()
     .reduce((acc, recEvent: RPC.EmittedEvent | RPC.Event) => {
-      let abiEvent: AbiEvent | AbiEvents = abiEvents[recEvent.keys.shift() ?? 0];
+      const currentEvent: RPC.EmittedEvent | RPC.Event = JSON.parse(JSON.stringify(recEvent));
```
huynhducduy, line 205: Should we use `lossless-json` from `src/utils/json.ts` to prevent number corruption?

PhilippeR26: Normally not. All very large numbers should be STRING type. Block number & fee price are easily handled by JS numbers.

PR #1442, opened 2025-07-12
fix: improve hex string validation in isHex function

reallesee:
## Motivation and Resolution
Replace the regex in isHex function from /^0x[0-9a-f]*$/i to /^0x[0-9a-fA-F]*$/ to make the code more readable and slightly more efficient by explicitly including both uppercase and lowercase hex characters instead of using the case-insensitive flag.

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


tabaktoni:
@reallesee, There is a consequential issue with this PR, removing "i", you exclude 0X as valid hex. Check the tests.

PR #1449, opened 2025-07-23
test: modify estimation of tip for Devnet

PhilippeR26:
## Motivation and Resolution
When executing the test suite in Devnet, we have a flood of tip error messages.
Devnet is starting from block 0 (no history of transactions) and creates a new block at each transaction. So, the default `maxBlocks` and `minTxsNecessary` are not adapted for Devnet.
This PR create a small initiial history of Devnet transactions, and is changing the default values (only for Devnet)

## Usage related changes
N/A

- if Devnet is just launched (zero blocks), 3 random transactions are created, with random tips, to create an initial tip history.
- account.getEstimateTip() is overwritten when using Devnet, to use new default values:
```ts
minTxsNecessary: options.minTxsNecessary ?? 3,
maxBlocks: options.maxBlocks ?? 10,
```
- To create these 3 new transactions before launching the test suite, some parts of `tests/config/fixtures.ts` are necessary, but Jest do not accept at this stage code including the use of `subscribe`. So the file has been splitted in 2 files: `fixtures.ts` & `fixturesInit.ts`.
- When tip is estimated out of an account, the logger has been temporary set to `critcal`.

Tested with success in:
- New Devnet
- Devnet with history
- Testnet

> [!NOTE]
> - wording of `devnet-rs` replaced by `starknet-devnet`.
> - prettier has automatically modified some documentation files.

## Checklist:
- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/__tests__/WebSocketChannel.test.ts
+++ b/__tests__/WebSocketChannel.test.ts
@@ -2,7 +2,8 @@
 import { Provider, Subscription, WebSocketChannel } from '../src';
 import { logger } from '../src/global/logger';
 import { StarknetChainId } from '../src/global/constants';
-import { getTestAccount, getTestProvider, STRKtokenAddress, TEST_WS_URL } from './config/fixtures';
+import { getTestProvider, TEST_WS_URL } from './config/fixtures';
+import { getTestAccount, STRKtokenAddress } from './config/fixturesInit';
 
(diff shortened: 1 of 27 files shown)


tabaktoni (approved):
IGWT

PR #1462, opened 2025-07-29
docs: add Contract.providerOrAccount explanation in migration guide

PhilippeR26:
## Motivation and Resolution
Add in the migration guide an explanation to connect an account to a contract after contract instantiation.
v7:
```ts
myContract.connect(myAccount);
```

v8:
```ts
myContract.providerOrAccount = myAccount;
```

## Usage related changes
N/A

N/A

## Checklist:

### DIFF
--- a/www/docs/guides/migrate.md
+++ b/www/docs/guides/migrate.md
@@ -246,7 +246,7 @@ import { Contract } from 'starknet';
 const contract = await Contract.factory({
   contract: sierraContract, // Compiled Sierra contract
   casm: casmContract, // Compiled CASM contract
-  account: account,
+  account: account, // optional
   constructorCalldata: {
     name: 'Token',
     symbol: 'ERC20',
@@ -259,6 +259,9 @@ const contract = await Contract.factory({
   unique: true, // optional
   deployer: account.deployer, // optional
 });
+
+// if `account` property has not been defined above, it can be added later with:
+contract.providerOrAccount = myAccount;
 ```
 
 ### Removed Helper Functions


www/docs/guides/migrate.md
```diff
@@ -246,7 +246,7 @@ import { Contract } from 'starknet';
 const contract = await Contract.factory({
   contract: sierraContract, // Compiled Sierra contract
   casm: casmContract, // Compiled CASM contract
-  account: account,
+  account: account, // optional
```
penovicp, line 249: The `account` isn't optional here.

PhilippeR26: Should it be same concept in both?

tabaktoni: No because you cant do factory aka. Nor declare nor deploy without account.

To be honest optional provider or account on main class has no sense either, because without it you cant do anything with contract class

tabaktoni: I think this should be non-optional in both

PhilippeR26: Not sure.
- Create a Contract without account/provider : I have seen this case to use myContract.functions() to get a contract list of functions of a contract.
- Create a Contract with just a provider : Used to read Starknet (not write). Useful because you can read Starknet without asking to the user to connect an account.
- Create a Contract with an account: Used to write  Starknet.

tabaktoni: > * Create a Contract with just a provider : Used to read Starknet (not write). Useful because you can read Starknet without asking to the user to connect an account.
> * Create a Contract with an account: Used to write  Starknet.
Both these cases require a provider, so in dose it is not optional.


> * Create a Contract without account/provider : I have seen this case to use myContract.functions() to get a list of functions of a contract.
There are better ways to extract function but let say this is simplest one, in that case one can provide defaultProvider, or dummy provider. But this is an extreme edge case, one should use Calldata parser to do this.

PR #1475, opened 2025-08-18
[WIP] - replace request/response functional parsing to OO CairoTypes

tabaktoni:
## Motivation and Resolution
Goal:
OO "self-contained composable" - CairoType
AR "accurate representation" - Byte representation
DD "Data-driven configurability"  - Strategy Map

- [ ] Convert all implemented parsing types to CairoTypes
- [ ] Eventually remove request and response parsers
- [ ] Programmable request/response using Strategy map
- [ ] Remove formatter (replaced by Strategy map)
- [ ] refactor higher-order types to contain appropriate CairoTypes
- [ ] substitute lower-order types to contain CairoFelt instead of JS types like BigInt  (Felt should be the only one having value represented as non CairoType "ByteArray").

## Usage related changes
- At the end, for the Contract Class users, none ( parsing should be abstracted), and an addition should be an option to define a response type based on the cairo type using a strategy map.
- For the tooling devs, they should now have access to all Cairo representation types.  "self-contained composable units".

Instead of functional parsing and utils for api, have self-contained CairoTypes similar to the native JS types (Number, String, BigInt....)

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing


tabaktoni:
Issue with Diff

Issue #1477, opened 2025-08-21
CairoByteArray toBuffer() and toHexString() corruption


xxu-circle:
**Describe the bug**
Follow up from #1455 
it works partially, but the parsing the response still get corrupted bytearray. I have below as minimal repro
I believe the problem are inside toHexString, toBuffer which was doing some bigint conversion - and lost some zeroes along the way

**To Reproduce**
```typescript
import { CairoByteArray } from "starknet";

describe("bytearray", () => {
  describe("bytearray", () => {
    const content = "0x000000010000001900000002";
    const contentBuffer = Buffer.from(content.slice(2), "hex");
    const cairoByteArray = new CairoByteArray(contentBuffer);
    const cairoBuffer: Buffer = cairoByteArray.toBuffer();
    const cairoHexString = cairoByteArray.toHexString();

    it("toBuffer() should match", () => {
      expect(cairoBuffer.toString("hex")).toEqual(contentBuffer.toString("hex"));
    });

    it("toHexString() should match", () => {
      expect(cairoHexString).toEqual(content);
    });
  });
});
```

```
 FAIL  e2e/bytearray.test.ts
  ● bytearray › bytearray › toBuffer() should match

    expect(received).toEqual(expected) // deep equality

    Expected: "000000010000001900000002"
    Received: "010000001900000002"

  ● bytearray › bytearray › toHexString() should match

    expect(received).toEqual(expected) // deep equality

    Expected: "0x000000010000001900000002"
    Received: "0x10000001900000002000000"
```

**Expected behavior**
`toBuffer()` and `toHexString()` should resolve to the same byte array. 

- Starknet.js version 8.5.0

xxu-circle:
@tabaktoni

penovicp:
This issue from the preceding comment has been addressed in version 8.5.2.

xxu-circle:
Sounds good. I'll use the provided code snippet for now before you decide on the approach!

xxu-circle:
Provided code works with some modification - I do not pad last two felts because padding `pending_word` will cause [paddingDifference](https://github.com/starknet-io/starknet.js/blob/74b8d7257dd53b784aba602cfc3879c2293eb745/src/utils/cairoDataTypes/byteArray.ts#L253) to become negative.

tabaktoni:
This should be resolved with [v8.5.3](https://github.com/starknet-io/starknet.js/releases/tag/v8.5.3).
[...]

PR #1483, opened 2025-09-02
fix: enforce fixed size for cairo bytes

penovicp:
## Motivation and Resolution

Resolves the latest issue described in #1477.

Modifies the initialization of `CairoBytes31`, `CairoFelt252`, and consequently `CairoByteArray` to be more uniform, and adjusts their output methods to be better aligned with how RPC nodes communicate.

###  Initialization changes
The main change is that for the `CairoBytes31` class the `data` `Uint8Array` will always be initialized to the exact size of 31 regardless of input size. As an example, `new CairoBytes31('0x1')` and `new CairoBytes31('0x0001')` will both have the value of `data` set to `[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1]`.

A similar change is made to `CairoFelt252`, the underlying `data` isn't prepended with zeros to match a fixed size like with `CairoBytes31`, and is instead truncated to exclude leading zeros. `new CairoFelt252('0x1')` and `new CairoFelt252('0x0001')` will both result in `data` having the value `[1]`.

###  Method changes

The existing behaviour of the `CairoBytes31` `toHexString()` method is modified to exclude leading zeros, to include them `toHexString('padded')` can be used.

The items in the `CairoByteArray` `toApiRequest()` output will not have leading zeros to match how RPC nodes format `ByteArray` data in their responses.

The `CairoByteArray` `toElements()` method is introduced to enable a simple overview of the unified underlying data, it outputs and array of `Uint8Array` entries that represent all the `bytes31` data chunks and the pending word.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/utils/cairoDataTypes/CairoByteArray.test.ts
+++ b/__tests__/utils/cairoDataTypes/CairoByteArray.test.ts
@@ -707,13 +707,43 @@ describe('CairoByteArray Unit Tests', () => {
         '0x' +
         '000000019900000000000002222222374206275726e206d65737aaaa000001' +
         '000000029900000000000002222222374206275726e206d65737aaaa000002' +
-        '000000039900000000000002222222374206275726e206d65737aaaa000003';
+        '000000039900000000000002222222374206275726e206d65737aaaa000003' +
+        '00d0f0';
       const buffer = Buffer.from(content.slice(2), 'hex');
       const byteArray = new CairoByteArray(buffer);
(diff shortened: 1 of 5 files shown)


tabaktoni (approved):
lgtm

PR #1492, opened 2025-09-15
fix: fix waitfortransaction for rotating nodes services with lifeCycl…

tabaktoni:
## Motivation and Resolution
If a user connects to a node service or infra that rotates rpc nodes, the node can be unsynced, yet when conducting tx status, and returns TXN_HASH_NOT_FOUND after the tx has status. This indicates the tx lifecycle, but in this case, it could be an unsynced node.

The solution is to add lifecycle retries, default to 3, before concluding that it is evicted or another lifecycle event happened.
Also added retries directly to the method instead of just provider config.

## Usage related changes
waitForTransaction options
```ts
  /**
   * Define the number of retries before throwing an error for the transaction life cycle when the transaction is not found after it had a valid status.
   * This is useful for nodes that are not fully synced yet when connecting to service that rotate nodes.
   */
  lifeCycleRetries?: number;
  /**
   * Define the number of retries before throwing an error
   */
  retries?: number;
  /**
   * Define the time interval between retries in milliseconds
   */
```

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/channel/rpc_0_8_1.ts
+++ b/src/channel/rpc_0_8_1.ts
@@ -396,21 +396,22 @@ export class RpcChannel {
 
   public async waitForTransaction(txHash: BigNumberish, options?: waitForTransactionOptions) {
     const transactionHash = toHex(txHash);
-    let { retries } = this;
+    let retries = options?.retries ?? this.retries;
+    let lifeCycleRetries = options?.lifeCycleRetries ?? 3;
     let onchain = false;
     let isErrorState = false;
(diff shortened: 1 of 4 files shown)


penovicp (approved):

PhilippeR26 (approved):
Tested with success in Alchemy rpc 0.8 & 0.9.
Seems that default `retryInterval` can now be reduced a little.

PR #1514, opened 2025-11-07
fix(utils): fetch bindings

alexandre-abrioux:
## Motivation and Resolution

When running tests with Jest using `testEnvironment: "jsdom"` to simulate a browser environment, `globalThis` is defined but `globalThis.fetch` is removed by Jest.

When using Starknet's Javascript library, we end up with the following error in tests:
> TypeError: Cannot read properties of undefined (reading 'bind')

This is a fix to remediate that issue.

## Usage related changes

N/A

N/A

## Checklist:

- [ ] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/src/utils/connect/fetch.ts
+++ b/src/utils/connect/fetch.ts
@@ -1,8 +1,8 @@
 import { LibraryError } from '../errors';
 
-export default (typeof globalThis?.fetch !== 'undefined' && globalThis.fetch.bind(globalThis)) ||
-  (typeof window?.fetch !== 'undefined' && window.fetch.bind(window)) ||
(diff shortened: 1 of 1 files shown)


alexandre-abrioux:
@tabaktoni Thanks for reviewing this PR! And I apologize for my mistake; I should have tested this further.

I agree that `jsdom` is not the best solution, and ideally, we should use a headless browser. Unfortunately, I'm currently unable to do that, but I've already applied a quick fix, similar to the one you suggested.

The goal of this PR was to contribute to a fix that prevents it from happening to someone else. I made another change that should work for everyone. Let me know what you think!

PR #1516, opened 2025-11-12
fix(docs): correct TokenData field names in paymaster guide

avorylli:
Replace incorrect field names (address, tokenAddress) with token_address in React example and output sample. 

Add null check for gasToken before use.

Fix matches the actual TokenData interface structure.

### DIFF
--- a/www/docs/guides/account/paymaster.md
+++ b/www/docs/guides/account/paymaster.md
@@ -290,7 +290,7 @@ const App: FC = () => {
       {!gasToken && <p>Select a gas token</p>}
       <div>
         {account && (
-          <button disabled={loading} onClick={onClickExecute}>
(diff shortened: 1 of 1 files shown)


www/docs/guides/account/paymaster.md
```diff
@@ -242,6 +242,7 @@ const App: FC = () => {
   }
 
   const onClickExecute = () => {
+    if (!gasToken) return;
```
PhilippeR26: Not necessary. See L294

PhilippeR26:
Why have you removed ` || !gasToken` in ` disabled={loading || !gasToken}` ?

avorylli:
> Why have you removed ` || !gasToken` in ` disabled={loading || !gasToken}` ?

oops, my mistake, thought the 245 should be and 294 deleted, nevermind, reverted commit

PR #1520, opened 2025-11-19
feat: ressourceBounds are taking into account L2_MIN_PRICE

PhilippeR26:
## Motivation and Resolution
`estimateFee` is providing a `resourceBound` object that is not taking into account L2 MIN PRICE.
This value is 3 Gfri, and is defined in  https://governance.starknet.io/voting-proposals/9

## Usage related changes
Today, if `l2.max_price_per_unit` is lower than L2_MIN_PRICE, the node will answer with an error 53.
Now, automatic definition of resourceBounds is taking care of L2_MIN_PRICE and is not generating errors.
User can anyway defines manually resourceBounds with any (possibly invalid) value.

Just added 2 functions to get min and max of a collection of bigints.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/src/utils/stark/index.ts
+++ b/src/utils/stark/index.ts
@@ -29,11 +29,13 @@ import {
 import { parse, stringify } from '../json';
 import {
   addPercent,
+  bigIntMax,
   bigNumberishArrayToDecimalStringArray,
   bigNumberishArrayToHexadecimalStringArray,
   toHex,
 } from '../num';
 import { isBigInt, isObject, isString } from '../typed';
+import { L2_MIN_PRICE } from '../../global/constants';
 
 type V3Details = Required<
   Pick<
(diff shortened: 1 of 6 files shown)


src/utils/stark/index.ts
```diff
@@ -200,9 +203,12 @@ export function toOverheadResourceBounds(
         estimate.l2_gas_consumed,
         overhead !== false ? overhead.l2_gas.max_amount : 0
       ),
-      max_price_per_unit: addPercent(
-        estimate.l2_gas_price,
-        overhead !== false ? overhead.l2_gas.max_price_per_unit : 0
+      max_price_per_unit: bigIntMax(
+        L2_MIN_PRICE,
+        addPercent(
```
tabaktoni: We can have a warning or info here.

tabaktoni:
Should be resolved on Node/feeder level

PR #1521, opened 2025-11-20
Change tip alert

PhilippeR26:
## Motivation and Resolution
Devs are complaining to have a flood of error messages when using snjs Devnet, integration or Testnet networks.
Estimation of tip is optimized for Mainnet, but is generating a lot of error messages when using other networks, due to a low quantity of transactions per block.
To be less disruptive for devs, I propose to replace the error message by a warning message.

## Usage related changes
In node.js, jest tests, browser console, very aggressive error messages are now replaced by warnings. 

## Checklist:

- [x] All tests are passing

### DIFF
--- a/src/provider/modules/tip.ts
+++ b/src/provider/modules/tip.ts
@@ -316,7 +316,7 @@ async function getTipStatsParallel(
 
     // Handle insufficient transaction data
     if (allTips.length < minTxsNecessary) {
-      logger.error(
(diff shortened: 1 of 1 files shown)


tabaktoni (approved):
Sure, it's fine.
It was important when we migrated to the fee mechanism for devs to be aware, but now it is time to switch to warn.

PR #1558, opened 2025-12-24
fix: add fail-fast guard for pre_confirmed block tag

Ocheretovich:
Problem
Number("0.9.0") returns NaN, making the version guard always falsy. pre_confirmed could be used silently with RPC versions < 0.9.
Fix
Replaced with compareVersions() already used in the codebase for semver comparisons.

### DIFF
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -15,7 +15,7 @@ import { isSierra } from './contract';
 import { formatSpaces } from './hash';
 import { parse, stringify } from './json';
 import { isHex, toHex } from './num';
-import { toApiVersion } from './resolve';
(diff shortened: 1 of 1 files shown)


Ocheretovich:
Hey @tabaktoni, please respond if anything needs to be corrected.

tabaktoni:
@Starknetio For future dev, this should be checked @penovicp metioned fix on this already imeplemnted but need to be confirmed

Ocheretovich:
Number("0.9.0") returns NaN, so the guard was never triggered. Replaced with compareVersions() which is already used for semver comparisons in the codebase. Also cached config.get('rpcVersion') to avoid double call.

PR #1559, opened 2026-01-02
Enable tree shaking

### LINKED ISSUE #1324: Enable tree shaking
It would be great if we could enable tree shaking of the starknet.js library for the next version.

ondrej-secretkeylabs:
This tells bundlers:
- index.* has side effects (because it re-exports from provider/paymaster/deployer)
- provider/index.* has side effects (defaultProvider = new RpcProvider())
- paymaster/index.* has side effects (defaultPaymaster = new PaymasterRpc())
- deployer/index.* has side effects (defaultDeployer, legacyDeployer)

## Motivation and Resolution

Resolves https://github.com/starknet-io/starknet.js/issues/1324

n/a

## Usage related changes

None

opt-in

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,12 @@
     "CHANGELOG.md",
     "dist"
   ],
+  "sideEffects": [
+    "./dist/index.*",
+    "./dist/provider/index.*",
+    "./dist/paymaster/index.*",
+    "./dist/deployer/index.*"
+  ],
   "scripts": {
     "prepare": "npm run build && husky",
     "build": "tsup && npm run build:esm && npm run build:iife && npm run build:dts",


tabaktoni (changes requested):
But the build (tsup config) has a single entry point src/index.ts and outputs a flat bundle — everything goes into dist/index.js, dist/index.mjs, etc. There are no dist/provider/index.* or dist/paymaster/index.* files. Those subdirectories don't exist in dist/.

So ./dist/provider/index.*, ./dist/paymaster/index.*, and ./dist/deployer/index.* would never match anything. Only ./dist/index.* would match.

This means bundlers consuming the published package would treat provider/index, paymaster/index, and deployer/index as side-effect-free (since the globs don't match any real files), which is technically harmless for the flat bundle case but misleading.

1. For this to actually enable tree shaking, you'd need:
Multiple entry points in tsup (one per module)
Subpath exports in package.json ("./provider", "./account", etc.)
Then the sideEffects globs would match real files

The PR is well-intentioned and the side effect analysis is correct, but it has no practical effect given the current single-entry-point build. The three subdirectory globs don't match any files in dist/. It's harmless but doesn't actually enable tree shaking. Would need build config changes (multiple entry points + subpath exports) to be meaningful.

PR #1582, opened 2026-02-20
feat!: refactor Account to use composition pattern, plugin system, docs - V10

tabaktoni:
- [x] BREAKING CHANGE: Account no longer extends Provider. Provider methods must now be accessed via account.provider property.

Changes:
- Account uses composition instead of inheritance
- Add plugin system to replace ts-mixer
- Update all tests to use account.provider.xyz()
- Create migration guide and plugin documentation

- [x] [remove pako package and replace with native and faster CompressionStream, update TS 5.9](https://github.com/starknet-io/starknet.js/pull/1582/changes/75b8226ff70fa243d06f7530fb9053d26a9740ad)

## Checklist:
- [ ] Updated the tests
- [ ] All tests are passing


PhilippeR26 (approved):
- `.use()` is not very descriptive. Something like `.usePluging()` could be more easy to understand.
- Pako removal: Are we sure to not have the same problem than fetch, buffer, .... (do not work properly outside of node.js, in exotic environment)?
- Personnaly, I find migrate.md and plugins.md too verbose. It's a bit boring to read, with not important data, and a lot of words to say few things (why explain how to create a npm package, utility of best practices, same subject explained several times,...)

src/provider/index.ts
```diff
@@ -1,9 +1,9 @@
...
 
-export { RpcProvider as Provider } from './extensions/default'; // backward-compatibility
+export { RpcProvider as Provider } from './rpc'; // backward-compatibility
 export { LibraryError, RpcError } from '../utils/errors';
 export * from './interface';
-export * from './extensions/default';
+export * from './rpc';
 export * from './modules';
 
 export const defaultProvider = new RpcProvider({ default: true });
```
penovicp: I think we could remove the `defaultProvider` and let users manage it themselves if they want a memoized instance.

penovicp (approved):

PR #1603, opened 2026-04-08
SNIP-36 - helper for creation of signed transaction

PhilippeR26:
## Motivation and Resolution
In the frame of SNIP-36, proof are generated by an off-line sequencer backend. A `RPC.INVOKE_TXN_V3` signed transaction has to be sent to this backend.
Today, `Account.execute` is generating this `RPC.INVOKE_TXN_V3` internally and send it to the node, but we don't have a method to provide these data to users.
This PR is adding a such method.

## Usage related changes
- Added a new `Account.getSignedTransaction()` method. Parameters are 100% the same than `Account.execute()`. Example:
```ts
const myCall = myTestContract.populate("Say_HelloPhil126",  { messages: "0x123" });
const tx3: INVOKE_TXN_V3 = await account0.getSignedTransaction(myCall);
  ```
  Then `tx3` can be sent to the backend, to generate the proof.
-  Added a new `RpcProvider.invokeSignedTx()`. This method is not necessary in the frame of SNIP-36, but it's consistent to have in Starknet.js both a method to create a signed Tx and a method to consume it. Example:
```ts
const res = await account0.provider.channel.invokeSignedTx(tx3);
const txR3 = await account0.provider.waitForTransaction(res.transaction_hash);
```

As 90% of code of `Account.execute()` is common with `Account.getSignedTransaction()`, this common part has been written is a new `prepareInvoke()` function.

## Checklist:

- [x] Updated the tests
- [x] All tests are passing


tabaktoni:
@copilot resolve the merge conflicts in this pull request

tabaktoni (commented):
Ideally, this would be extracted as a plugin in the new plugin system, as this is helper and not a core protocol method.
Also, the channel is dedicated to pure RPC methods non apstractioned, so this should be on a level up.

adrienlacombe:
> Ideally, this would be extracted as a plugin in the new plugin system, as this is helper and not a core protocol method. Also, the channel is dedicated to pure RPC methods non apstractioned, so this should be on a level up.

+1

PhilippeR26:
Two points to address:

- **On extracting getSignedTransaction() as a plugin:**
`getSignedTransaction() `relies on `prepareInvoke()`, an internal method introduced in this branch to share the signing pipeline between **execute()** and **getSignedTransaction()**. That method in turn uses `resolveDetailsWithTip()` and `resolveTransactionVersion()`, both private, which access instance properties (`transactionVersion`, `defaultTipType`) that are not part of` AccountInterface`.

Moving this to a plugin would require either 
-  promoting` prepareInvoke()` into `AccountInterface` — _a breaking change for all downstream implementors_ 
- or duplicating ~40 lines of internal signing logic inside the plugin. Neither is acceptable. Unlike `fastExecute` or `starknetId` plugins, which operate purely on the public `AccountInterface` API, this method is structurally wired into the signing internals.

- **On invokeSignedTx() placement:**
`fetchEndpoint` is protected on the channel — no higher layer can call `starknet_addInvokeTransaction` directly without going through a channel method. Moving the logic to the provider would mean either making `fetchEndpoint` public or duplicating the RPC call there, both of which are worse. It is also worth noting that the existing `channel.invoke()` already builds a transaction internally before sending it — so the channel is not strictly limited to "pure RPC methods" in the current design. `invokeSignedTx()` is actually less abstracted than` invoke()`: it takes an already-built transaction and sends it as-is.

`provider.invokeSignedTx()` is the user-facing API; the channel method exists solely to support it, which follows the same pattern as every other operation in this layer stack.

In conclusion, while I understand the intent behind these review comments, a closer look at the details reveals that the current implementation is the right approach.

adrienlacombe (approved):

PR #1614, opened 2026-05-27
docs: correct current docs content

adrienlacombe:
## Summary
- Correct the current API intro text for `RpcProvider` and `Account` provider delegation.
- Add the current Outside Execution SNIP-9 compatibility check.
- Clarify current `WalletAccount` provider options without touching frozen docs versions.

## Split from
Part 1 of the split from #1613: pure content correction.

## Validation
- `npm run typecheck` in `www`
- `npm run ts:check`
- `DOCS_BASE_URL=/starknet.js/ npm run build`
- `npm audit --omit=dev` in `www`
- `git diff --check`

PR #1615, opened 2026-05-27
fix: accept provider options in WalletAccountV5.connect

adrienlacombe:
## Summary
- Broaden `WalletAccountV5.connect` to accept `ProviderOptions | ProviderInterface`.
- Align the static helper signature with `WalletAccountV5Options.provider` and the documented `{ nodeUrl }` usage.

## Stack
Part 2 of the split from #1613: code correction.

Base PR: #1614

## Validation
- `npm run typecheck` in `www`
- `npm run ts:check`
- `DOCS_BASE_URL=/starknet.js/ npm run build`
- `npm audit --omit=dev` in `www`
- `git diff --check`

### DIFF
--- a/src/wallet/account.ts
+++ b/src/wallet/account.ts
@@ -174,7 +174,7 @@ export class WalletAccount extends Account {
   }
 
   static async connectSilent(
-    provider: ProviderInterface,
+    provider: ProviderOptions | ProviderInterface,
     walletProvider: StarknetWalletProvider,
     cairoVersion?: CairoVersion,
     paymaster?: PaymasterOptions | PaymasterInterface
--- a/src/wallet/accountV5.ts
+++ b/src/wallet/accountV5.ts
@@ -19,6 +19,7 @@ import {
   CompiledSierra,
   DeclareContractPayload,
   MultiDeployContractResponse,
+  ProviderOptions,
   TypedData,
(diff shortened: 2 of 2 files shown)


PhilippeR26 (changes requested):
`connectSilent` has the same restriction in both classes and should receive the same fix:

- `src/wallet/account.ts` — `WalletAccount.connectSilent(provider: ProviderInterface, ...)`
- `src/wallet/accountV5.ts` — `WalletAccountV5.connectSilent(provider: ProviderInterface, ...)`

Both delegate directly to their respective `connect(provider, ..., true)`, so the runtime already supports `ProviderOptions` — only the type signature is too narrow. `ProviderOptions` is already imported in both files after this PR.

Suggested change (same pattern, both files):

```ts
static async connectSilent(
  provider: ProviderOptions | ProviderInterface,
  ...

adrienlacombe:
@PhilippeR26 good catch — fixed in 754d9e09. Both `WalletAccount.connectSilent` and `WalletAccountV5.connectSilent` now accept `ProviderOptions | ProviderInterface`, matching `connect()`. `npm run ts:check` passes. Re-requesting your review.

PhilippeR26 (approved):
Looks good to me

PR #1616, opened 2026-05-27
docs: upgrade docs website to Docusaurus 3

adrienlacombe:
## Summary
- Upgrade the docs site to Docusaurus 3 and matching React/TypeDoc/Docusaurus package versions.
- Keep the Docusaurus config in TypeScript as `www/docusaurus.config.ts`; remove the JavaScript config.
- Rework generated API docs output and sidebars for the new typedoc-plugin-markdown layout.
- Add build-time compatibility handling for frozen versioned-doc links without editing `www/versioned_docs/**`.

## Stack
Part 3 of the split from #1613: Docusaurus upgrade.

Base PRs: #1614, then #1615

## Validation
- `npm run typecheck` in `www`
- `npm run ts:check`
- `DOCS_BASE_URL=/starknet.js/ npm run build`
- `npm audit --omit=dev` in `www`
- `git diff --check`


www/docusaurus.config.ts
```diff
@@ -0,0 +1,367 @@
...
+
+  plugins: [
+    [
+      'docusaurus-plugin-typedoc',
+      {
+        entryPoints: ['../src/index.ts'],
+        tsconfig: '../tsconfig.json',
+        out: 'docs/API',
+        name: 'Starknet.js API',
+        includeVersion: true,
```
PhilippeR26, line 343: Here it's necessary to add:
```ts
cleanOutputDir: false,
```
Otherwise, `www/docs/API/.gitkeep` is killed at build

PhilippeR26: See `.gitignore` about this file.

PhilippeR26 (changes requested):
A small problem identified above.

When the final url of deployment will be known, I think that it has to be added in GitHub settings / variables, and Docusaurus should use it. It's not recommended to have this url hard-coded in the code.
Let's wait the final url to approve this PR.

adrienlacombe:
Thanks @PhilippeR26 — addressed both points in c4985474:

1. **`.gitkeep` killed at build** — added `cleanOutputDir: false` to the `docusaurus-plugin-typedoc` options, so `www/docs/API/.gitkeep` (per `www/.gitignore`) survives the build.

2. **Hard-coded deployment URL** — the site `url` is no longer hard-coded. It now reads from a mandatory `DOCS_URL` env var (build throws if it's unset/empty — no default), wired through the deploy workflow from the `DOCS_URL` repository variable. Once the final URL is known, just set the `DOCS_URL` variable in repo settings.

Heads-up: since `DOCS_URL` is now required, the GitHub variable must be populated before the next Pages deploy will succeed.

PhilippeR26:
Let's wait the final url to finalize this PR.

www/docusaurus.config.ts
adrienlacombe: Addressed in `bc83dba9`: added `cleanOutputDir: false` to the TypeDoc plugin config so `www/docs/API/.gitkeep` is preserved during docs builds.

www/docusaurus.config.ts
```diff
@@ -0,0 +1,384 @@
...
+const DEFAULT_DOCS_URL = 'https://starknet-io.github.io';
+const DEFAULT_DOCS_BASE_URL = '/starknet.js/';
+const docsBaseUrl = generateBaseUrl(process.env.DOCS_BASE_URL || DEFAULT_DOCS_BASE_URL);
+const migrationGuideLink = `${docsBaseUrl}docs/guides/migrate`;
+// const migrationGuideLink = `${docsBaseUrl}docs/next/guides/migrate`;
+
+const config: Config = {
+  title: 'Starknet.js',
+  tagline: 'JavaScript library for Starknet',
+  url: requireEnv('DOCS_URL', DEFAULT_DOCS_URL),
```
PhilippeR26, line 180: For a local build, there is no DOCS_URL defined, so only the DEFAULT_DOCS_URL fallback is used. 
In local build, the deployed url must not be use (all has to remain local). The code from the previous commit is the right one : 
`url: requireEnv('DOCS_URL', 'http://localhost:3000')`.
For deployment, the CI will handle this value properly.

www/docusaurus.config.ts
```diff
@@ -0,0 +1,384 @@
...
+  baseUrl: docsBaseUrl,
```
PhilippeR26, line 181: Same than L180, there is no DOCS_BASE_URL defined in a local build, so the fallback DEFAULT_DOCS_BASE_URL is used. For local build, it has to be "/", not "/starknet.js/", otherwise the test in a local browser will not be "http://localhost:3000/"
For deployment, the CI will handle this value properly.

.github/workflows/manual-docs-deploy-pages.yml
```diff
@@ -33,6 +34,7 @@ jobs:
           rm -rf www/docs/API/
           npm run docs:build
         env:
+          DOCS_URL: ${{ vars.DOCS_URL || 'https://starknet-io.github.io' }}
```
PhilippeR26, line 37: vars.DOCS_URL is not existing in github/settings .
Today, only the CI fallback is working.
To clean the situation:
- add  the var in Variables

Otherwise, future maintainers will be lost.

.github/workflows/manual-docs-deploy-pages.yml
```diff
@@ -33,6 +34,7 @@ jobs:
...
           DOCS_BASE_URL: ${{ vars.DOCS_BASE_URL || '/starknet.js/' }}
```
PhilippeR26, line 38: vars.DOCS_BASE_URL is not existing in github/settings/variables. It's available in github/settings/secrets.
Today, only the CI fallback is working.
To clean the situation:
- add  the var in Variables
- remove it from secrets.
Otherwise, future maintainers will be lost.

PhilippeR26:
Globally, 2 main problems:

- Local build uses the fallback values from docusaurus.config, but these are not the right values for a local build. This was fixed in the previous commit but reverted in the last one.
- For CI deployment, some variables used in the workflow do not exist in the GitHub settings, so only the fallback values are used. This is clearly not a clean approach (and is not a new problem).

www/docusaurus.config.ts
adrienlacombe: Addressed in `be3d4c5a`: restored the local `DOCS_URL` fallback to `http://localhost:3000`.

www/docusaurus.config.ts
adrienlacombe: Addressed in `be3d4c5a`: restored the local `DOCS_BASE_URL` fallback to `/`, while CI continues to provide `/starknet.js/` for deployment.

.github/workflows/manual-docs-deploy-pages.yml
adrienlacombe: Addressed: added `DOCS_URL=https://starknet-io.github.io` to the `github-pages` environment Variables, and updated the workflow comment in `be3d4c5a` to document that setting.

.github/workflows/manual-docs-deploy-pages.yml
adrienlacombe: Addressed: `DOCS_BASE_URL=/starknet.js/` is set in the `github-pages` environment Variables, and the duplicate `DOCS_BASE_URL` secret has been removed from both environment and repository secrets.

adrienlacombe:
> Globally, 2 main problems:
> 
> * Local build uses the fallback values from docusaurus.config, but these are not the right values for a local build. This was fixed in the previous commit but reverted in the last one.
> * For CI deployment, some variables used in the workflow do not exist in the GitHub settings, so only the fallback values are used. This is clearly not a clean approach (and is not a new problem).

sorry for the regression, hopefully it is better now, ty

PhilippeR26 (approved):
Looks good to me.

PR #1617, opened 2026-05-30
fix: add validation to cairofelt to ensure numbers are valid felt252 …

### LINKED ISSUE #1584: CallData.compile serializes bigint > felt252 as single felt, causing RPC "expected hex string... undefined" when value is > felt252
**Describe the bug**
When using `CallData.compile()` without an ABI, passing a bigint larger than max felt252 (e.g. `UINT_256_MAX` for a felt252) for a field causes that value to be serialized via `felt(value)` as a single felt. The value does not fit in a felt252, so the resulting payload is invalid. The RPC then fails with a generic error: `-32602: Invalid invoke transaction v3: expected hex string to be prefixed by '0x': undefined`
 (or similar deserialization error), instead of a clear client-side error that the value is out of range or must be encoded as a single felt.

**To Reproduce**
1. Build calldata with `CallData.compile()` without ABI.
2. Include a field that is a u256 but pass it as a single bigint, e.g. `UINT_256_MAX` (2^256 - 1).
3. Use that calldata in an invoke.
4. Call `estimateInvokeFee` or `addInvokeTransaction`.
5. Observe: RPC returns `-32602: Invalid invoke transaction v3: expected hex string to be prefixed by '0x': undefined` (or similar deserialization error).

- Node version [e.g. 16.0.1]: e.g. 24.x
- Starknet.js version: 9.2.1
- Network [devnet, testnet]: devnet (fork) / testnet as applicable

qphe95:
## Motivation and Resolution

When using `CallData.compile()` without an ABI, passing a bigint larger than the felt252 prime (e.g. `UINT_256_MAX`) caused that value to be silently serialized via `felt(value)` as a single felt. Since the value doesn't fit in a felt252, the resulting payload was invalid and the RPC failed with a generic deserialization error like:

```
-32602: Invalid invoke transaction v3: expected hex string to be prefixed by '0x': undefined
```

This change adds range validation to the `CairoFelt()` function so that any numeric value (bigint, number, hex string, or whole number string) outside the felt252 range `[0, PRIME)` throws a clear client-side error immediately:

```
Value <val> is out of felt252 range [0, <PRIME>)
```

This aligns the deprecated `CairoFelt()` function with its successor `CairoFelt252`, which already performed this validation.

N/A

## Usage related changes

- `CallData.compile()` without ABI now throws a clear error when a leaf value exceeds the felt252 range, instead of producing invalid calldata that causes cryptic RPC errors.
- Direct calls to `cairo.felt()` or `CairoFelt()` with out-of-range values now throw immediately.

- Updated `CairoFelt` tests to expect errors for values outside felt252 range.
- Added explicit test coverage for the case where a value is within `uint256` range but outside `felt252` range.

fixes https://github.com/starknet-io/starknet.js/issues/1584

## Checklist:

- [x] Rebased to the last commit of the target branch (or merged it into my branch)\- [x] Linked the issues which this PR resolves
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/utils/cairoDataTypes/CairoFelt.test.ts
+++ b/__tests__/utils/cairoDataTypes/CairoFelt.test.ts
@@ -41,19 +41,19 @@ describe('CairoFelt function', () => {
 
   test('should properly handle large BigInt values', () => {
     // Examples of large BigInt values found in blockchain environments.
-    expect(
+    expect(() =>
       CairoFelt(
         BigInt('57896044618658097711785492504343953926634992332820282019728792003956564819967')
       )
-    ).toBe('57896044618658097711785492504343953926634992332820282019728792003956564819967');
+    ).toThrow(/out of felt252 range/);
     expect(CairoFelt(BigInt('1524157875019052100'))).toBe('1524157875019052100');
   });
 
   test('should properly handle large hex number strings', () => {
     // Examples of large hex number strings found in blockchain environments.
-    expect(CairoFelt('0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141')).toBe(
-      '115792089237316195423570985008687907852837564279074904382605163141518161494337'
-    );
+    expect(() =>
+      CairoFelt('0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141')
+    ).toThrow(/out of felt252 range/);
     expect(CairoFelt('0x10A')).toBe('266');
   });
 
@@ -81,12 +81,12 @@ describe('CairoFelt function', () => {
   });
 
   test('should correctly handle zero-prefixed and hexadecimal string numbers', () => {
-    expect(CairoFelt('00123')).toBe('00123');
+    expect(CairoFelt('00123')).toBe('123');
     expect(CairoFelt('0xFF')).toBe(BigInt('0xFF').toString());
   });
 
-  test('should properly handle smallest integer', () => {
-    expect(CairoFelt(Number.MIN_SAFE_INTEGER)).toBe('-9007199254740991');
+  test('should reject negative integers', () => {
+    expect(() => CairoFelt(Number.MIN_SAFE_INTEGER)).toThrow(/out of felt252 range/);
   });
 
   test('should properly handle largest integer', () => {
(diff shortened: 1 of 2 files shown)


__tests__/utils/cairoDataTypes/CairoFelt.test.ts
```diff
@@ -207,11 +207,10 @@ describe('CairoFelt function', () => {
   });
 
   test('should properly handle edge numeric values and formats', () => {
-    expect(CairoFelt(Number.MIN_SAFE_INTEGER)).toBe('-9007199254740991');
+    expect(() => CairoFelt(Number.MIN_SAFE_INTEGER)).toThrow(/out of felt252 range/);
```
PhilippeR26: Isn't it the same as L88?

src/utils/cairoDataTypes/felt.ts
```diff
@@ -44,6 +48,8 @@ export function CairoFelt(it: BigNumberish): string {
     }
     // Whole numeric strings
     if (isStringWholeNumber(it)) {
+      const val = BigInt(it);
+      assert(val >= 0n && val < PRIME, `Value ${val} is out of felt252 range [0, ${PRIME})`);
       return it;
```
PhilippeR26: Why not use here also  `return val.toString()` ?
Today, "0007" is returning "007".
Corrected, it will be "7".

src/utils/cairoDataTypes/felt.ts
```diff
@@ -23,14 +23,18 @@ import { addCompiledFlag } from '../helpers';
 export function CairoFelt(it: BigNumberish): string {
   // BN or number
   if (isBigInt(it) || Number.isInteger(it)) {
-    return it.toString();
+    const val = BigInt(it);
+    assert(val >= 0n && val < PRIME, `Value ${val} is out of felt252 range [0, ${PRIME})`);
```
PhilippeR26: This line is used 3 times. It has to be refactored in an util function. For example as a static function of CairoFelt252:
```ts
static assertRange(val: bigint) {
  assert(val >= 0n && val < PRIME, `Value ${val} is out of felt252 range [0, ${PRIME})`);
}
```

PhilippeR26 (changes requested):
3 small topics to handle.

PhilippeR26 (approved):
Looks good to me.

One note for next time: it's generally a bad habit to submit a PR directly from your fork's `develop` branch. Please create a dedicated feature branch (e.g. `fix/cairofelt-validation`), implement your changes there, and open the PR from that branch. It keeps things cleaner for everyone reviewing.

PR #1619, opened 2026-06-04
fix: fix snip-12 u256 hash mismatch

### LINKED ISSUE #1464: SNIP-12 u256 hash mismatch
**Describe the bug**
Currently the struct hash of typed data containing u256 doesn't match the hash struct of u256 calculated with in cairo, causing a mismatch of message signature 

**To Reproduce**
[link to gist](https://gist.github.com/JoE11-y/19b8a02c29e2f6e029f3d3fb5945add3)

**Expected behavior**
Should match

**Screenshots**

- Browser & version [e.g. chrome, safari, webworker]
- Node version [e.g. 16.0.1]
- Starknet.js version: 7.6.4
- Network [devnet, testnet]: devnet

**Additional context**

qphe95:
fixes #1464

## Motivation and Resolution

In SNIP-12 revision 1 (ACTIVE), `starknet.js` was computing the struct hash of `u256`-containing typed data by treating `u256` as a **nested struct** with its own type-hash prefix:

```
// OLD (incorrect)
structHash(u256) = poseidon([typeHash(u256), low, high])
```

However, in Cairo `u256` is a core language primitive whose built-in `Hash` trait simply flattens `low` and `high` directly into the parent hash state. When OpenZeppelin's SNIP-12 guide uses `update_with(*self)` on a struct containing `u256`, Cairo adds `low` and `high` as individual felts — **not** as a nested sub-hash. This caused off-chain signatures produced by starknet.js to fail on-chain verification.

We fixed this by introducing an internal `encodeValues` helper that allows a single field to expand into multiple encoded values. For `u256` in ACTIVE revision, we now return `['u128', 'u128']` / `[low, high]` directly, matching Cairo's behavior. The exported `encodeValue` function remains backward-compatible as a thin wrapper.

I also ran additional validation with https://gist.github.com/JoE11-y/19b8a02c29e2f6e029f3d3fb5945add3 and verified the reported issue was fixed.

Here are the results:

```
Cairo output
Chain ID: 0
Version: 0

Expected Type hash:  0x36991bcf00d0c2ac546204827d8a5e9924e5495b1184866fb61407bc92f76d3
Actual Type hash:    0x36991bcf00d0c2ac546204827d8a5e9924e5495b1184866fb61407bc92f76d3 ✅

Expected Struct Hash: 0x3bd5c51013c6a5d6870fcb918d4aa263da7c989ae9b87a27e70e9b62c8def63
Actual Struct Hash:   0x163fd1c083b57a2c46420dad63029329de470d95ddd6738637dcc11bc3a5d7d ✅ (matches fixed starknet.js)

Expected Message Hash: 0x506cb0d4fa139e1add1e9d756fa2bba5f930aea55534685ca9077ae52ce5afc
Actual Message Hash:   0x5200dc6d2e4c406ad9a4a3975c60c524a473cce24bba0c98a1505f15fc7e57
```

I verified that both hashes now match exactly between Cairo and the fixed starknet.js when using the same domain values:

```
Hash
Cairo
Fixed starknet.js
Struct Hash
0x163fd1c083b57a2c46420dad63029329de470d95ddd6738637dcc11bc3a5d7d
0x163fd1c083b57a2c46420dad63029329de470d95ddd6738637dcc11bc3a5d7d ✅
Message Hash
0x5200dc6d2e4c406ad9a4a3975c60c524a473cce24bba0c98a1505f15fc7e57
0x5200dc6d2e4c406ad9a4a3975c60c524a473cce24bba0c98a1505f15fc7e57 ✅
```

(The message hash alignment required using the same domain values as the Cairo test: chainId: 0, version: 1, revision: 1. The original JS gist used 'SN_SEPOLIA' / '1' / '1' which is why the message hash differed even after the struct fix.)

N/A — this is a pure client-side hashing fix.

---

## Usage related changes

- `u256` fields in SNIP-12 ACTIVE revision typed data now produce struct hashes that match Cairo's on-chain `Hash` implementation.
- `getMessageHash`, `getStructHash`, and `encodeData` all reflect this change automatically.
- The exported `encodeValue` function throws for `u256` (since it returns multiple values), but this was already an internal-only edge case; the main entry points work correctly.

---

- Added an internal `encodeValues` helper returning `[string[], string[]]` to support multi-value field encodings (currently only `u256`).
- Updated `encodeData`, array (`*`), enum, and merkle-tree handling to use `encodeValues` and properly spread multi-value results.
- Added dedicated tests for `u256` struct hash correctness and `u256` out-of-bounds validation.
- Updated inline snapshot for `examplePresetTypes` message hash to reflect the new flattened encoding.

---

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/utils/typedData.test.ts
+++ b/__tests__/utils/typedData.test.ts
@@ -339,7 +339,7 @@ describe('typedData', () => {
 
     messageHash = getMessageHash(examplePresetTypes, exampleAddress);
     expect(messageHash).toMatchInlineSnapshot(
-      `"0x185b339d5c566a883561a88fb36da301051e2c0225deb325c91bb7aa2f3473a"`
+      `"0x16e540c07db37a545d639a19511fb28d0886786332a93c1e381d00aa30094f2"`
     );
 
     messageHash = getMessageHash(exampleEnum, exampleAddress);
@@ -356,6 +356,48 @@ describe('typedData', () => {
     expect(spyPoseidon).toHaveBeenCalled();
     spyPedersen.mockRestore();
     spyPoseidon.mockRestore();
+  });
+
+  test('should flatten u256 to match Cairo hash', () => {
+    const u256TypedData = {
+      types: {
+        StarknetDomain: [
+          { name: 'name', type: 'shortstring' },
+          { name: 'version', type: 'shortstring' },
+          { name: 'chainId', type: 'shortstring' },
+          { name: 'revision', type: 'shortstring' },
+        ],
+        SimpleStruct: [{ name: 'some_u256', type: 'u256' }],
+      },
+      primaryType: 'SimpleStruct',
+      domain: {
+        name: 'DappName',
+        version: '1',
+        chainId: 'SN_SEPOLIA',
+        revision: '1',
+      },
+      message: {
+        some_u256: {
+          low: 1000,
+          high: 0,
+        },
+      },
+    };
+
+    const typeHash = getTypeHash(u256TypedData.types, 'SimpleStruct', TypedDataRevision.ACTIVE);
+    expect(typeHash).toMatchInlineSnapshot(
+      `"0x36991bcf00d0c2ac546204827d8a5e9924e5495b1184866fb61407bc92f76d3"`
+    );
+
+    const structHash = getStructHash(
+      u256TypedData.types,
+      'SimpleStruct',
+      u256TypedData.message,
+      TypedDataRevision.ACTIVE
+    );
+    expect(structHash).toMatchInlineSnapshot(
+      `"0x163fd1c083b57a2c46420dad63029329de470d95ddd6738637dcc11bc3a5d7d"`
+    );
   });
 
   describe('should fail validation', () => {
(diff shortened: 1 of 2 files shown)


PhilippeR26 (commented):
Hi @qphe95, and thank you for this PR — the diagnosis is genuinely well done and the on-chain verification work you provided is appreciated. You've correctly identified a real and long-standing source of confusion around `u256` in SNIP-12.

After reviewing it carefully against the spec, however, we won't be able to merge this, and I'd like to explain why so the path forward is clear.

**The current `starknet.js` behavior is conformant with SNIP-12 (revision 1, ACTIVE).** In the spec, `u256` is defined as a *preset type*, i.e. a struct: `"u256": [{ "name": "low", "type": "u128" }, { "name": "high", "type": "u128" }]`.

The encoding rule for any object/struct is a single, explicit one: `Enc[x] = hash_array( type_hash(x), Enc[param1], ..., Enc[paramN] )`.

There is no exception carved out for `u256`. Since it is a preset *struct*, it receives the standard object encoding — **with the `type_hash` prefix**. That is exactly what `starknet.js` does today. This PR removes that prefix and flattens `low`/`high` directly, which makes the output match Cairo's native `Hash` trait, but in doing so it diverges from the literal text of SNIP-12.

That distinction is the key point. `starknet.js` is intended to be a strict reflection of SNIP-12. When there is a discrepancy between the spec and what Cairo implementations actually do on-chain — and you're right that there is one here — the correct place to resolve it is **the SNIP-12 specification first**, not the client library. Changing `starknet.js` ahead of the spec would put us in the position of silently implementing behavior the spec doesn't describe, and would break conformance for everyone else relying on it as the reference.

So the right next step is to take this to the [SNIPs repository](https://github.com/starknet-io/SNIPs): open an issue/PR proposing that `u256` (and by extension the preset types that contain it, like `TokenAmount` and `NftId`) be flattened without a `type_hash` prefix, ideally as a clarification or a new revision. Once the spec is amended and accepted, we can implement the matching change in `starknet.js` with confidence — and your analysis here would be excellent material to support that proposal.

I'm going to close this for now, but please don't take it as a dismissal of the work — the underlying issue (#1464) stays open, and the spec discussion is the productive way to move it forward. Thanks again.

PR #1627, opened 2026-06-15
docs: add AGENTS.md and infra-free test:unit script for agents

adrienlacombe:
## What & why

Makes the repo friendlier to AI coding agents (Claude Code, Cursor, Copilot, …) — and to new human contributors — by documenting conventions in one place and giving a verification loop that doesn't need live infrastructure.

Today an agent has to rediscover the build/test/commit conventions every session, and the default `npm test` **hard-fails without a running devnet** (`globalSetup` throws), so an agent often can't verify its own changes.

## Changes

- **`AGENTS.md`** (+ a one-line `CLAUDE.md` pointing to it): setup & core commands, the test-verification loop, a `src/` architecture map, conventions (Conventional Commits, PR target branches, semantic-release owns version/CHANGELOG), and gotchas (`posttest` reformats files, `pretest` gates on lint/types, commitlint rejects bad messages, `.claude/` is git-ignored).
- **`jest.unit.config.ts` + `npm run test:unit`**: runs the pure unit suites under `__tests__/utils/` **without** the devnet-connecting `globalSetup`. Excludes the two network-dependent suites (`ethSigner`, `batch`). Infra-free, ~20s.
- **`tsconfig.eslint.json`**: includes the new config so it passes `lint` / `ts:check` (mirrors how `jest.config.ts` is already listed).

No source or public-API changes; `docs:`/tooling only, so it does not affect the changelog.

## Verification

Run on this branch (no devnet):

- `npm run test:unit` → **60 suites, 1471 tests pass** (~24s)
- `npm run ts:check` → clean
- `npx eslint jest.unit.config.ts` → clean
- `npx prettier --check` on all changed files → clean

The full integration suite (`npm test`) still requires a devnet/RPC node and was not run here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,2 @@
+See [AGENTS.md](./AGENTS.md) for agent guidance (setup, the infra-free `npm run test:unit`
+verification loop, the `src/` architecture map, conventions, and gotchas).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,2 +1,0 @@
-See [AGENTS.md](./AGENTS.md) for agent guidance (setup, the infra-free `npm run test:unit`
-verification loop, the `src/` architecture map, conventions, and gotchas).
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -99,4 +99,6 @@ pinned `starknet_specs` dev-dependencies define them. Be version-aware when touc
 - The default test suite hard-fails without a reachable devnet (see above). A bare
   `jest <path>` will also hit this — go through `npm run test:unit` for infra-free runs.
 - `.claude/` is git-ignored, so per-user agent config there is local-only. This
-  `AGENTS.md` (and `CLAUDE.md`) are the shared, committed source of truth.
+  `AGENTS.md` is the shared, committed source of truth. Tools that only read a
+  `CLAUDE.md` (e.g. Claude Code) can use a local, git-ignored `CLAUDE.md` that
+  points here.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,102 @@
+# AGENTS.md
+
+Guidance for AI coding agents (Claude Code, Cursor, Copilot, etc.) working in this
+repository. Humans: see [CONTRIBUTING.md](./CONTRIBUTING.md) — this file is a
+condensed, agent-oriented companion to it, not a replacement.
+
+`starknet.js` is the JavaScript/TypeScript SDK for Starknet. It is a published npm
+library (`starknet`), so the public API surface is a contract — treat breaking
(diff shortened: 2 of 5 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,2 @@
+See [AGENTS.md](./AGENTS.md) for agent guidance (setup, the infra-free `npm run test:unit`
```
PhilippeR26: ### On the committed `CLAUDE.md` — I'd suggest dropping it
My one objection is the **committed `CLAUDE.md`**. I'd like it removed, for two reasons:

**1. Vendor neutrality.** `starknet.js` is a vendor-neutral SDK. `CLAUDE.md` is specific to one AI vendor (Anthropic). Committing it privileges one tool and sets a precedent for `CURSOR.md`, `COPILOT.md`, `.windsurfrules`, etc. piling up at the repo root. `AGENTS.md` is exactly the cross-tool, vendor-neutral convention meant to avoid that — let's keep it as the single shared source of truth.

**2. It collides with per-developer local configs.** CLAUDE users already keep a personal `CLAUDE.md` locally. Committing one to the repo breaks that: on the next pull, git refuses to merge (`untracked working tree file 'CLAUDE.md' would be overwritten`), and their richer personal file gets replaced by a 2-line pointer.

### Suggested change

- **Remove the committed `CLAUDE.md`.** Keep `AGENTS.md` as the shared, committed guidance.
- Optionally slim `AGENTS.md` so it points to `CONTRIBUTING.md` for the parts it currently duplicates.
- **Keep `test:unit` as-is** — that's the part worth merging.

### For devs who use Claude Code

Claude Code reads `CLAUDE.md` automatically, not `AGENTS.md`. So if you want that auto-pickup, create/modify a **local, git-ignored** `CLAUDE.md` with a single line:

```md
See [AGENTS.md](./AGENTS.md) for agent guidance.
```

PhilippeR26 (changes requested):
Looks good, ... except the presence of CLAUDE.md in the canonical repo.

CLAUDE.md
adrienlacombe: ✅

PhilippeR26 (approved):
Thanks for this 👍 The `test:unit` / `jest.unit.config.ts` part is the real win — I verified it locally: **1471 tests pass with no devnet**. Removing the committed `CLAUDE.md` is the right call for vendor neutrality.

I pushed a small commit to fix one leftover: the *Gotchas* section still listed `CLAUDE.md` as a "shared, committed source of truth", which is no longer true now that it's removed. Updated it to point devs to a local, git-ignored `CLAUDE.md` instead.

LGTM after that.

Issue #1238, opened 2024-10-03
Add Error when rpc url fails.


ikemHood:
**Is your feature request related to a problem? Please describe.**

I've been using alchemy rpc for a while and it works fine wothout appending an apikeys sometimes.

maybe a change from them or something but it suddenly didn't work anymore with 404 error..

making contract calls now fails with 
```bash
2415 |    * @param response string[] - response from the method
2416 |    * @return Result - parsed response corresponding to the abi
2417 |    */
2418 |   parse(method, response) {
2419 |     const { outputs } = this.abi.find((abi) => abi.name === method);
2420 |     const responseIterator = response.flat()[Symbol.iterator]();
                                    ^
TypeError: undefined is not an object (evaluating 'response.flat')
      at parse (/home/ikem/dev/horuslab/karst-backend/node_modules/starknet/dist/index.mjs:2420:30)
```

Throwing a better error when the apicall fails  make it easier to debug and find the actual cause of a misbehaior

A little digging and `await rawResult.json();` here returns error as undefind when this error happens

```js
  async fetchEndpoint(method, params) {
    try {
      const rawResult = await this.fetch(method, params, this.requestId += 1);
      const { error, result } = await rawResult.json();
      console.log({error, result})
      this.errorHandler(method, params, error);
      return result;
    } catch (error) {
      this.errorHandler(method, params, error?.response?.data, error);
      throw error;
    }
  }
```
`{
  error: undefined,
  result: undefined,
}
`

**Describe the solution you'd like**
Throw a detail rpc error when the api call fails

CanutoIX:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PhilippeR26:
I looked into this one. The original trigger — Alchemy returning a `404` without an API key — was a provider-side change and is no longer reproducible: the latest starknet.js works fine against an Alchemy node.

However, the underlying problem the report points at is still valid. When a node replies with **neither `result` nor `error`** (a malformed or non-JSON-RPC response), `fetchEndpoint` silently returns `undefined`, which then crashes downstream with the cryptic `TypeError: response.flat is not a function` instead of a meaningful error.

[...]

PR #1629, opened 2026-06-16
fix: surface clear error instead of crashing on empty node reply

PhilippeR26:
## Motivation and Resolution
Addresses the actionable part of #1238. The original trigger (Alchemy returning a `404` without an API key) was a provider-side change and is no longer reproducible. The valid underlying problem remains: when a node replies with **neither `result` nor `error`** (malformed or non-JSON-RPC response), `fetchEndpoint` silently returned `undefined`, crashing downstream with the cryptic `TypeError: response.flat is not a function`.

`fetchEndpoint` now throws an explicit `LibraryError` (naming the RPC method) when the response is empty/malformed. Detection is structural — only when both `result` and `error` are absent — so falsy-but-valid results (e.g. genesis block number `0`) are preserved.

Applies to both the RPC `0.9` and `0.10.x` channels.

## Usage related changes

- A malformed/empty node reply now raises a clear `LibraryError` naming the failed RPC method, instead of an obscure `response.flat is not a function` crash further down the stack.

- Added unit tests for the empty/malformed response case (throws `LibraryError`) and for the falsy-but-valid `result: 0` case (returned as-is).

## Checklist:

- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/__tests__/rpcChannel010.test.ts
+++ b/__tests__/rpcChannel010.test.ts
@@ -306,4 +306,31 @@ describe('UNIT TEST: RPC 0.10.1 Channel - New API features', () => {
       expect(result).not.toHaveProperty('proof_facts');
     });
   });
+
+  describe('malformed RPC response', () => {
+    test('throws a clear error when response has neither result nor error', async () => {
+      // Reproduces issue #1238: a node (e.g. Alchemy returning 404) replies with a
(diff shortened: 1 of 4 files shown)


PhilippeR26:
tested in Devnet rpc0.ç & 0.10, and Tesnet rpc 0.10

Issue #1237, opened 2024-10-03
Proposal for `waitForTransaction` improvement


Cussone:
Now that block times are much faster it would make sense to lower the `retryInterval` param from `5000ms` (waited on twice, so it's `10s`) to something more realistic.

Also, I find it weird that `retryInterval` is an optional param. In my opinion, it would make more sense to require it explicitly, with a suggested value in the docs. I see it as a better DX.

This would be a breaking change. So, it is something to consider for the next major release.

Cussone:
> hello @Cussone can i hop on this issue?

Not sure what you are referring to, but feel free to write anything you think about this.

PhilippeR26:
Has been discussed in Slack: will not be modified, as user can customize the value.

Cussone:
> Has been discussed in Slack: will not be modified, as user can customize the value.

Yup, that was the conclusion for the current version, but I opened this proposal for the next version, as we also discussed on Slack.

CanutoIX:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

Sekhmet:
Another improvements would be:
- do not call `sleep` after receipt is successfully resolved [here](https://github.com/starknet-io/starknet.js/blob/3bdee6797a72053a699e363388a8d24275e1f389/src/channel/rpc_0_7.ts#L360) - right now even if transaction is ready we are still waiting for 5s.
- do not call `sleep` on entry [here](https://github.com/starknet-io/starknet.js/blob/3bdee6797a72053a699e363388a8d24275e1f389/src/channel/rpc_0_7.ts#L304)

I've had issue with my tests where everything was super slow (even when using starknet-devnet) and actually the issue is that each call to `waitForTransaction` will take at least 10s unless you modify `transactionRetryIntervalFallback`.

PhilippeR26:
Closed, as:
- If you want to go fast (with a single and fast access node, no events usage), you  can use fastExecute() function.
- For other cases (slow nodes, parallelized nodes, slow infra), the 5 s delay is necessary, especially is you declare/deploy/use events. If you are confident that your specific case/infra can go faster, you can customize the retry interval.

Issue #827, opened 2023-11-13
[V6] Lower timer - speed up


tabaktoni:
**Is your feature request related to a problem? Please describe.**
After refactoring waitForTransaction on [v5.23.0](https://github.com/starknet-io/starknet.js/releases/tag/v5.23.0), it is possible now to lower timeout to 1s and gain ~x2.5 performance.
However, some contract tests still failed at this speed.
This feature needs more testing / communicating with nodes regarding the "Readiness" status for TX.
At the moment even though the transaction status is SUCCEEDED and ACCEPTED_ON_L2 pending block nodes are not ready to work with the contract/transaction.
Investigate for contract do we need to check that the latest block, instead of pending?
For transactions why txHash is unknown if the status is SUCCEEDED and ACCEPTED_ON_L2 on pending Block

**Describe the solution you'd like**
Lower RPC retryInterval to 1000 ms.

**Describe alternatives you've considered**

**Additional context**

vibenedict:
Hi, i will like to take up this issue.

Issue #688, opened 2023-07-15
waitForTransaction is waiting too much time


PhilippeR26:
**Is your feature request related to a problem? Please describe.**
Until now, Starknet was a little ... slow, the `provider.waitForTransaction ()` is first waiting 8 seconds before testing the transaction status. If necessary, it's retested each 8s.

https://github.com/0xs34n/starknet.js/blob/7bc96fae09c4cfb0427b622794be884d3dc4e554/src/provider/sequencer.ts#L495

https://github.com/0xs34n/starknet.js/blob/7bc96fae09c4cfb0427b622794be884d3dc4e554/src/provider/rpc.ts#L442

But now, Cairo_lang 0.12.0 is installed in all Starknet networks, and all is much much faster.
These 8s of wait seems to be no more adapted to the "new" Starknet, and should be reduced

**Describe the solution you'd like**
I propose to reduce this value to something between 4 and 5s.

**Describe alternatives you've considered**
Be patient... even when Starknet has already ended the job...

**Additional context**

PhilippeR26:
https://twitter.com/Starknet/status/1679762624756715520?t=ww4tn0u59aAXb_E8RZLICg&s=19

tabaktoni:
For now i will leave it as is.
Check mainet transaction time https://voyager.online/analytics?page=network
it is between 30-50 sec

PhilippeR26:
Shall we consider time for block approval, or time to transaction Approved in L1?

tabaktoni:
Reducer to 5sec in 0.12.1 RP

PR #1631, opened 2026-06-16
fix: avoid extra waitForTransaction receipt delay

### LINKED ISSUE #1632: waitForTransaction adds an extra polling delay after receipt is available
## Description

`waitForTransaction` can return later than necessary after a transaction receipt is already available.

The current flow first polls `getTransactionStatus` until the transaction reaches a configured success state. It then enters a second loop that calls `getTransactionReceipt`. That receipt loop sleeps for `retryInterval` at the end of every iteration, including the iteration where `getTransactionReceipt` succeeds.

As a result, successful calls pay one extra polling interval before returning the receipt.

## Impact

For callers using a short polling period, such as `retryInterval: 500`, this adds a guaranteed extra 500ms after the receipt is available. With the default fallback interval, the extra delay can be much larger.

This also makes `waitForTransaction` appear noticeably slower than a direct websocket subscription, especially when the websocket client observes status or receipt availability immediately while `waitForTransaction` is still sleeping after its successful receipt fetch.

## Expected behavior

Once `getTransactionReceipt` succeeds, `waitForTransaction` should return the receipt immediately.

The retry sleep should still happen after failed receipt attempts, preserving the existing pacing for nodes that report transaction success before the receipt endpoint is ready.

## Affected code

- `src/channel/rpc_0_9_0.ts`
- `src/channel/rpc_0_10_2.ts`

## Proposed fix

Move the receipt-loop sleep into the failed-receipt path only, so retry pacing is preserved without delaying successful returns.

A draft fix is open in #1631.

loothero:
Fixes #1632

## Summary
- Remove the unconditional post-receipt polling delay from `waitForTransaction`
- Preserve retry pacing when `getTransactionReceipt` is not ready yet
- Apply the fix to both RPC 0.9 and RPC 0.10 channel implementations
- Add regression tests proving successful receipt availability resolves without another polling interval

## Validation
- `npx jest __tests__/rpcChannel010.test.ts __tests__/rpcChannel09.test.ts --runInBand --testNamePattern "returns immediately" --config '{"transform":{"^.+\\\\.(t|j)sx?$":"@swc/jest"},"testEnvironment":"node","setupFilesAfterEnv":["./__tests__/config/jest.setup.ts"]}'`
- `npx eslint src/channel/rpc_0_10_2.ts src/channel/rpc_0_9_0.ts __tests__/rpcChannel010.test.ts __tests__/rpcChannel09.test.ts`
- `npm run ts:check`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/__tests__/rpcChannel010.test.ts
+++ b/__tests__/rpcChannel010.test.ts
@@ -74,6 +74,41 @@ describe('UNIT TEST: RPC 0.10.1 Channel - New API features', () => {
     fetchSpy?.mockRestore();
   });
 
+  describe('waitForTransaction', () => {
+    test('returns immediately after the receipt is available', async () => {
+      jest.useFakeTimers();
+      const receipt = { transaction_hash: '0x123' };
+      const transactionStatusSpy = jest
+        .spyOn(channel, 'getTransactionStatus')
+        .mockResolvedValueOnce({
+          finality_status: 'ACCEPTED_ON_L2',
+          execution_status: 'SUCCEEDED',
+        });
+      const transactionReceiptSpy = jest
+        .spyOn(channel, 'getTransactionReceipt')
+        .mockResolvedValueOnce(receipt as any);
+
+      const promise = channel.waitForTransaction('0x123', { retryInterval: 1_000 });
+      let settled = false;
+      const settledPromise = promise.then((result) => {
+        settled = true;
+        return result;
+      });
+
+      try {
+        await jest.advanceTimersByTimeAsync(1_000);
+        await Promise.resolve();
+
+        expect(settled).toBe(true);
+        await expect(settledPromise).resolves.toBe(receipt);
+      } finally {
+        transactionStatusSpy.mockRestore();
+        transactionReceiptSpy.mockRestore();
+        jest.useRealTimers();
+      }
+    });
+  });
+
   describe('response_flags (includeProofFacts)', () => {
     test('getBlockWithTxs with includeProofFacts sends response_flags', async () => {
       fetchSpy = jest.spyOn(channel, 'fetch');
--- a/__tests__/rpcChannel010.test.ts
+++ b/__tests__/rpcChannel010.test.ts
@@ -107,6 +107,49 @@ describe('UNIT TEST: RPC 0.10.1 Channel - New API features', () => {
         jest.useRealTimers();
       }
     });
+
+    test('waits one retry interval when the receipt is not available yet', async () => {
+      jest.useFakeTimers();
+      const receipt = { transaction_hash: '0x123' };
+      const transactionStatusSpy = jest
+        .spyOn(channel, 'getTransactionStatus')
+        .mockResolvedValueOnce({
+          finality_status: 'ACCEPTED_ON_L2',
+          execution_status: 'SUCCEEDED',
+        });
+      const transactionReceiptSpy = jest
+        .spyOn(channel, 'getTransactionReceipt')
+        .mockRejectedValueOnce(new Error('Transaction hash not found'))
+        .mockResolvedValueOnce(receipt as any);
+
+      const promise = channel.waitForTransaction('0x123', { retryInterval: 1_000 });
+      let settled = false;
+      const settledPromise = promise.then((result) => {
+        settled = true;
+        return result;
+      });
+
+      try {
+        // status polling interval elapsed, first receipt attempt rejected
+        await jest.advanceTimersByTimeAsync(1_000);
+        expect(transactionReceiptSpy).toHaveBeenCalledTimes(1);
+        expect(settled).toBe(false);
+
+        // pacing between two receipt attempts is preserved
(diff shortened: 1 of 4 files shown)


src/channel/rpc_0_10_2.ts
```diff
@@ -569,10 +569,10 @@ export class RpcChannel {
         if (retries <= 0) {
           throw new Error(`waitForTransaction timed-out with retries ${this.retries}`);
         }
+        // eslint-disable-next-line no-await-in-loop
+        await wait(retryInterval);
       }
       retries -= 1;
```
PhilippeR26: In the same logical, we can also move this line in the catch block.
Valid for both rpc version.

__tests__/rpcChannel010.test.ts
```diff
@@ -74,6 +74,41 @@ describe('UNIT TEST: RPC 0.10.1 Channel - New API features', () => {
...
+        expect(settled).toBe(true);
+        await expect(settledPromise).resolves.toBe(receipt);
+      } finally {
+        transactionStatusSpy.mockRestore();
+        transactionReceiptSpy.mockRestore();
+        jest.useRealTimers();
+      }
+    });
+  });
+
```
PhilippeR26, line 154: We ha a missing test, that validate that getTransactionReceipt() can reject once, then resolve. 
Valid both both rpc version.

PhilippeR26 (changes requested):
Thanks for this PR, and for the clear write-up in the linked issue.

First, some context from reviewers, because this is a **sensitive area**. The timing of `waitForTransaction` has been discussed many times over the years (e.g. #688, #827, #1237). Requests to **lower the retry interval** or make it mandatory were repeatedly declined, because slow nodes and load-balanced / parallelized RPC services (slightly out of sync with each other) genuinely need that pacing between retries to behave correctly — especially for declare/deploy and event-reading flows. For users who run a single fast node and want speed, `fastExecute()` already exists, and the interval is customizable.

What makes this PR **different from those rejected requests** is important: it does **not** touch the retry pacing at all. It only removes the `wait(retryInterval)` that was executed *after* `getTransactionReceipt` has already succeeded — i.e. an extra sleep happening once the final receipt is already in hand, just before returning it. The failure path is strictly preserved:

- the `if (retries <= 0)` check stays *before* the decrement, so the timeout threshold is unchanged;
- the inter-retry `wait` is kept on every failed receipt attempt, so the pacing protecting slow / desynced nodes is intact.

I verified this by walking each execution path (nominal success, failure-then-success, and timeout): only the post-success terminal wait disappears, everything else is byte-for-byte equivalent in behavior.

### Testing
Given the history, I ran extensive tests before approving:

- **Setup**: a counter contract, looped `execute(setCounter(i)) → waitForTransaction → read-back` sequences [20 iterations], nonce left to automatic resolution, verifying after each wait: receipt (`isSuccess`), state read (`getCounter`) compared against the expected value, and emitted event.
- **retryInterval values**: 500, 5000, 6000 ms.
- **Providers**: Alchemy Sepolia (load-balanced), Zan Sepolia (2 parallel nodes, slower),  local node  Pathfinder / remote private Pathfinder node, Devnet, ...
  Pathfinder v0.22.7, Starknet v0.14.3, rpc 0.10.3
- **Result**: all sequences passed. The state read-backs and nonce progression were always consistent right after the wait returned, with no regression observed across providers and intervals.

### Review comments
I left two inline comments:
1. The added tests only cover the "success on first attempt" case. A **failure-then-success** test (receipt rejects once, then resolves) would lock in the preserved inter-retry pacing — that's the regression guard that really matters here. Could you add it?
2. Minor: `retries -= 1` now runs on the success path too (harmless, the loop exits). See inline note.

Overall this looks correct and low-risk to me, pending the additional test. Thanks again.

PhilippeR26:
Since there was no follow-up, I pushed a commit addressing the two review points myself:

1. **`retries -= 1` moved into the `catch` block** of the receipt loop, right after the timeout check and before the `wait` (both RPC 0.9 and 0.10). The decrement now only counts *failed* receipt attempts; the timeout threshold is unchanged since `if (retries <= 0)` still runs before the decrement.

2. **Added the failure-then-success regression test** (`waits one retry interval when the receipt is not available yet`) to both `__tests__/rpcChannel010.test.ts` and `__tests__/rpcChannel09.test.ts`. `getTransactionReceipt` rejects once and then resolves; the test asserts, to the millisecond, that no second attempt happens before `retryInterval` has elapsed — locking in the inter-retry pacing that protects slow / desynced nodes.

I verified the new test is a real guard: removing the `await wait(retryInterval)` from the `catch` makes it fail (`Expected 1 call, received 2`).

Validation: `npx jest __tests__/rpcChannel010.test.ts __tests__/rpcChannel09.test.ts --runInBand` (devnet), `npx eslint` on the four changed files, `npm run ts:check` — all green.

Thanks again for the fix.

PhilippeR26 (approved):
OK for me

PR #1667, opened 2026-08-11
fix(shortString): zero-pad each byte in encodeShortString

latent-9:
## Summary

`encodeShortString` builds the felt hex with `charCodeAt(0).toString(16)` and no zero-padding:

```ts
str.replace(/./g, (char) => char.charCodeAt(0).toString(16))
```

For any ASCII character below `0x10` (tab `0x09`, and other control chars that `isASCII` accepts), this emits a single hex nibble, which misaligns every following byte and yields the wrong felt. Newlines (`0x0A`/`0x0D`) are not matched by `.` at all, so they pass through verbatim and later break `BigInt`.

```
encodeShortString('a\tb')  ->  0x61962   (wrong, felt 399714)
correct                    ->  0x610962  (felt 6359394)
decodeShortString(encodeShortString('a\tb'))  ->  'a2'   (not 'a\tb')
```

This reaches the typed-data (SNIP-12) signing path: `typedData.ts getHex` routes felt/shortstring string values through `encodeShortString`, so a control character in a signed field produces a wrong message hash.

## Fix

Pad each byte to two hex digits and match all characters including newlines:

```ts
str.replace(/[\s\S]/g, (char) => char.charCodeAt(0).toString(16).padStart(2, '0'))
```

Regression-safe: for printable characters (code >= 0x20) the output already had two hex digits, so `padStart` is a no-op and all existing outputs are unchanged.

## Test

Added a control-char round-trip test (`'a\tb'` encodes to `0x610962` and decodes back, plus a newline case). The existing tests only cover printable strings, so this path was uncovered.

### DIFF
--- a/__tests__/utils/shortString.test.ts
+++ b/__tests__/utils/shortString.test.ts
@@ -12,6 +12,14 @@ describe('shortString', () => {
     expect(encodeShortString('hello')).toMatchInlineSnapshot(`"0x68656c6c6f"`);
   });
 
+  test('should zero-pad control chars and round-trip', () => {
+    // 0x09 (tab) is a valid ASCII char < 0x10; without padding it would emit a
+    // single nibble and misalign every following byte.
+    expect(encodeShortString('a\tb')).toBe('0x610962');
+    expect(decodeShortString(encodeShortString('a\tb'))).toBe('a\tb');
+    expect(decodeShortString(encodeShortString('a\nb'))).toBe('a\nb');
+  });
+
   test('should throw if string to encode is too long', () => {
     expect(() =>
       encodeShortString('hello world hello world hello world hello world hello world hello world')
--- a/src/utils/shortString.ts
+++ b/src/utils/shortString.ts
@@ -119,7 +119,9 @@ export function splitLongString(longStr: string): string[] {
 export function encodeShortString(str: string): string {
   if (!isASCII(str)) throw new Error(`${str} is not an ASCII string`);
   if (!isShortString(str)) throw new Error(`${str} is too long`);
-  return addHexPrefix(str.replace(/./g, (char) => char.charCodeAt(0).toString(16)));
+  return addHexPrefix(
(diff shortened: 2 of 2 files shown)


PhilippeR26:
Thanks for the report — the bug is real, and it is a bit wider than described here.

Confirmed on `develop`:

```
encodeShortString('a\tb')     ->  0x61962      (should be 0x610962)
encodeShortString('a\nb')     ->  "0x61\n62"   (not a hex string; throws on the next BigInt)
encodeShortString('\x01\x02') === encodeShortString('\x12')   // both 0x12
```

The typed-data consequence you describe is accurate: with `{ note: 'a\tb' }` in a `shortstring`
field, `getMessageHash` returns a hash silently derived from the wrong felt, and the `\n` case
throws instead.

What the patch does not cover is the same defect in the other direction. `decodeShortString`
splits on `/.{2}/g`, which cannot handle an odd-length hex string — and that is exactly what a
node returns for a short string whose first byte is below `0x10`, since `NUM_AS_HEX` carries no
leading zero:

```
decodeShortString('0x961')  ->  "1"    // the felt for '\ta'
decodeShortString('2401')   ->  "1"    // same value, decimal form
```

The added test does not catch this, because it only decodes the zero-padded output of the fixed
encoder, never a felt in the shape a node actually returns. Merging the encode side alone would
leave the pair asymmetric.

We are not going to take this on the v10 line, for a reason unrelated to the quality of the patch.
`encodeShortString` and `decodeShortString` have been deprecated since v8.2.0 (August 2025) and are
scheduled for removal in v11. Their replacement already returns the correct value for your example,
in both directions:

```ts
new CairoBytes31('a\tb').toHexString(); // 0x610962
new CairoBytes31('0x961').decodeUtf8(); // '\ta'
```

So instead of patching a function that disappears in the next major, we are removing it and
documenting the migration in the v11 guide. Closing this PR on that basis.

The report was useful: it is what confirmed the removal is the right call rather than another
round of patching, and it surfaced the decode-side gap, which will be covered by the migration
notes. Anyone still on v10 who hits this before upgrading can get the same value from
`CairoBytes31`, which is already exported today.

PR #1612, opened 2026-05-26
[codex] Address high CodeQL findings

adrienlacombe:
## Summary
- replace CodeQL-flagged regex checks with deterministic string parsing for tuple, fixed-array, and Starknet ID domain validation
- replace single-occurrence `*` removal with `replaceAll` in calldata array type handling
- add focused regression coverage for the flagged patterns

## Validation
- `npm run lint`
- `npm run ts:check`
- `npx jest -i --config "{\"transform\":{\"^.+\\\\.(t|j)sx?$\":\"@swc/jest\"},\"testMatch\":[\"**/__tests__/**/(*.)+(spec|test).[jt]s?(x)\"]}" __tests__/utils/calldata/cairo.test.ts __tests__/utils/cairoDataTypes/CairoFixedArray.test.ts __tests__/utils/starknetId.test.ts`

Note: the repo default Jest config requires a local Starknet devnet, so the focused unit test command bypasses global setup for these pure utility tests.

PR #1669, opened 2026-08-11
fix(starknetId): accept bigAlphabet characters in isStarkDomain

### LINKED ISSUE #1668: isStarkDomain rejects bigAlphabet characters, so getAddressFromStarkName cannot resolve names the same version encodes correctly
**Describe the bug**

`getAddressFromStarkName` refuses to resolve names that the same library version encodes correctly.

StarknetID's encoding uses two alphabets, both defined in `src/utils/starknetId.ts` and both used by `useEncoded` / `useDecoded`:

```js
const basicAlphabet = 'abcdefghijklmnopqrstuvwxyz0123456789-';
const bigAlphabet = '这来';
```

The `isStarkDomain` guard that `getAddressFromStarkName` calls only accepts `[a-z0-9-]`, so a registered name containing a `bigAlphabet` character is rejected before the call ever reaches the naming contract. The library therefore contradicts itself: `useEncoded` produces the correct felt for such a name and `useDecoded` round-trips it, while `isStarkDomain` reports the same name as invalid.

**Version boundary**

The guard was added in #1301 and released in 6.21.2. `isStarkDomain` does not exist in 6.21.0 or 6.21.1, so these names resolved before that release and have failed since. The original implementation:

```js
/^(?:[a-z0-9-]{1,48}(?:[a-z0-9-]{1,48}[a-z0-9-])?\.)*[a-z0-9-]{1,48}\.stark$/
```

#1612 (released in 10.3.1) replaced the regex with per-character parsing to address a CodeQL finding, but kept the same `a-z` / `0-9` / `-` character set, so the behaviour is unchanged. Reproduced on 6.21.2, 6.24.1 and 10.6.8.

**To Reproduce**

`来baba这.stark` is registered and live on mainnet.

```js
const { RpcProvider } = require('starknet');

const provider = new RpcProvider({ nodeUrl: 'https://rpc.starknet.lava.build' });
await provider.getAddressFromStarkName('来baba这.stark');
```

On 6.21.1:

```
0x180f9e6c6cc47e2db5d888df7fb9f904fd59cb1a559824ad4f6e597c7c86ae7
```

On 6.21.2 and later:

```
Error: Invalid domain, must be a valid .stark domain
```

The encoding half of the same install disagrees with the guard:

```js
starknetId.isStarkDomain('来baba这.stark');    // false
starknetId.useEncoded('来baba这').toString();  // '11885385095'
starknetId.useDecoded([11885385095n]);        // '来baba这.stark'
```

`11885385095` is the felt the naming contract expects, and it is byte for byte identical on 6.21.1 and on 6.24.1 / 10.6.8. Encoding never regressed. Only the guard in front of it changed.

**Expected behavior**

`isStarkDomain` should accept every character the library's own encoder accepts, that is `basicAlphabet` plus `bigAlphabet`, so that any name registered through StarknetID resolves.

The current implementation iterates with `[...label].every(char => ...)`, so extending the existing character test with a `bigAlphabet.includes(char)` branch is sufficient. Both `bigAlphabet` characters are single UTF-16 code units, so the existing 48 character label limit is unaffected.

**Environment**

- Node 22
- starknet.js: fails on 6.21.2 through 10.6.8, works on 6.21.1
- Network: mainnet

**Additional context**

The same regex is present in starknetid.js at `packages/core/src/utils.ts`, so the fix likely applies there too.

SnowingFox:
## Problem

`getAddressFromStarkName` refuses to resolve names that the same library version encodes correctly. A registered Starknet.id name containing a `bigAlphabet` character (e.g. `来baba这.stark`, live on mainnet) is rejected by the `isStarkDomain` guard before the call ever reaches the naming contract, so it throws `Invalid domain, must be a valid .stark domain`. This is a self-inconsistency: `useEncoded`/`useDecoded` round-trip the same name, and the felt they produce matches what the naming contract expects.

## Root cause

The guard added in #1301 (and kept unchanged by the per-character rewrite in #1612) only accepts `[a-z0-9-]`, but the encoder in `src/utils/starknetId.ts` accepts `basicAlphabet` plus `bigAlphabet` (`'这来'`). `getAddressFromStarkName` calls `isStarkDomain` first (`src/plugins/starknet-id/index.ts`), so any name using a `bigAlphabet` character fails validation before it is ever encoded.

## Fix

Extend the per-character test in `isStarkDomain` with a `bigAlphabet.includes(char)` branch, so the guard accepts exactly the characters the encoder accepts. Both `bigAlphabet` characters are single UTF-16 code units, so the existing 48-character label limit is unchanged. The same character set also lives in starknetid.js (`packages/core/src/utils.ts`) and can be aligned separately there.

## Test

Added a regression test in `__tests__/utils/starknetId.test.ts` asserting that `isStarkDomain` accepts `bigAlphabet` names (including the mainnet example `来baba这.stark`), stays consistent with the encoder for the same name, and still enforces the 48-character label limit. Runs with the infra-free unit config:

```
npx jest -c jest.unit.config.ts -i __tests__/utils/starknetId.test.ts -t "bigAlphabet"
```

Fixes #1668

### DIFF
--- a/__tests__/utils/starknetId.test.ts
+++ b/__tests__/utils/starknetId.test.ts
@@ -65,10 +65,12 @@ describe('Should test StarknetId utils', () => {
   });
 
   test('Should validate StarknetId domain label boundaries', () => {
-    // exactly 48 chars — valid
-    expect(isStarkDomain(`${'a'.repeat(48)}.stark`)).toBe(true);
-    // subdomain label exactly 48 chars — valid
-    expect(isStarkDomain(`${'a'.repeat(48)}.example.stark`)).toBe(true);
+    // 47 chars — the longest ASCII label whose encoding still fits in a felt
+    expect(isStarkDomain(`${'a'.repeat(47)}.stark`)).toBe(true);
+    // subdomain label of 47 chars — valid
+    expect(isStarkDomain(`${'a'.repeat(47)}.example.stark`)).toBe(true);
+    // 48 chars — under the 48-char label limit, but the encoding overflows the felt
+    expect(isStarkDomain(`${'a'.repeat(48)}.stark`)).toBe(false);
     // subdomain label 49 chars — invalid (uniform 48-char limit on all labels)
     expect(isStarkDomain(`${'a'.repeat(49)}.example.stark`)).toBe(false);
     // empty label from double dot — invalid
@@ -90,8 +92,24 @@ describe('Should test StarknetId utils', () => {
     expect(useDecoded([useEncoded(name)])).toBe(`${name}.stark`);
     expect(isStarkDomain(`${name}.stark`)).toBe(true);
 
-    // both bigAlphabet chars are single UTF-16 units, so the 48-char label limit is unchanged
-    expect(isStarkDomain(`${'来'.repeat(48)}.stark`)).toBe(true);
-    expect(isStarkDomain(`${'来'.repeat(49)}.stark`)).toBe(false);
+    // a bigAlphabet character eats far more encoding capacity than an ASCII one, so what caps
+    // these labels is the felt bound, not the 48-char limit: 20 chars for '来', 40 for '这'
+    expect(isStarkDomain(`${'来'.repeat(20)}.stark`)).toBe(true);
+    expect(isStarkDomain(`${'来'.repeat(21)}.stark`)).toBe(false);
+    expect(isStarkDomain(`${'这'.repeat(40)}.stark`)).toBe(true);
+    expect(isStarkDomain(`${'这'.repeat(41)}.stark`)).toBe(false);
+  });
+
+  test('Should reject characters outside the Starknet.id alphabets', () => {
+    // An unknown character used to be skipped silently, so a mistyped name encoded as an
+    // existing one: 'Grug' and 'rug' both gave 9441n and resolved to the very same address.
+    expect(() => useEncoded('Grug')).toThrow('Invalid character "G"');
+    expect(() => useEncoded('cafè')).toThrow('Invalid character "è"');
+    // useEncoded takes a single label, so the separator is not part of its alphabets either
+    expect(() => useEncoded('starknet.js')).toThrow('Invalid character "."');
+
+    // the guard reports the same names as invalid, rather than resolving another account
+    expect(isStarkDomain('Grug.stark')).toBe(false);
+    expect(isStarkDomain('cafè.stark')).toBe(false);
   });
 });
--- a/__tests__/utils/starknetId.test.ts
+++ b/__tests__/utils/starknetId.test.ts
@@ -76,4 +76,22 @@ describe('Should test StarknetId utils', () => {
     // no name before .stark — invalid
     expect(isStarkDomain('.stark')).toBe(false);
   });
+
+  test('Should accept bigAlphabet characters that the encoder supports', () => {
+    // '这' and '来' are the two bigAlphabet characters used by useEncoded/useDecoded.
+    // The encoder round-trips names containing them, so the guard must accept them too,
+    // otherwise getAddressFromStarkName rejects names the same version encodes correctly.
+    expect(isStarkDomain('来baba这.stark')).toBe(true);
+    expect(isStarkDomain('starknet这.stark')).toBe(true);
+    expect(isStarkDomain('这.stark')).toBe(true);
+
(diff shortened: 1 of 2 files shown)


PhilippeR26:
Thanks for the patch. I reproduced the issue and confirm the diagnosis: `useEncoded` / `useDecoded` round-trip `bigAlphabet` names correctly and the naming contract resolves the felt they produce, while `isStarkDomain` rejects the same name. The guard is wrong.

Three things before this can be merged.

**1. Please retarget this PR to `beta`, not `develop`.**

The guard has behaved this way since 6.21.2 (January 2025) and no consumer is blocked on a `develop` release, so this ships with v11 rather than as a patch. `src/utils/starknetId.ts`, `src/plugins/starknet-id/index.ts` and `__tests__/utils/starknetId.test.ts` are byte-identical on both branches, so the patch applies as is — only the base branch needs to change.

**2. Please fix the root cause in the same PR, not just the guard.**

`useEncoded` silently skips any character outside both alphabets, and the truncated result resolves to a real third party:

```
Grug     -> 9441      -> 0x60b56b67e1b4dd1909376496b0e867f165f31c5eac7902d9ff48112f16ef9b
rug      -> 9441      -> 0x60b56b67e1b4dd1909376496b0e867f165f31c5eac7902d9ff48112f16ef9b
Vitalik  -> 809637090 -> resolves to italik.stark, not vitalik.stark
```

A single mistyped capital sends funds to someone else's address. This is also why the guard was introduced in #1301 in the first place. And `src/plugins/brother-id/index.ts` calls `useEncoded` with no guard at all, so `.brother` names are exposed to the same silent misresolution today.

Since this lands on a breaking branch, please address both:

- `useEncoded` throws on any character outside `basicAlphabet` + `bigAlphabet`.
- `isStarkDomain` derives validity from the encoder instead of restating the alphabet — that duplication is exactly what produced this bug:

```ts
import { PRIME, StarknetChainId, ZERO } from '../global/constants';

return name.split('.').every((label) => {
  if (label.length === 0 || label.length > 48) return false;
  try {
    return useEncoded(label) < PRIME;
  } catch {
    return false;
  }
});
```

The `PRIME` bound matters on its own: `useEncoded('z'.repeat(48))` overflows the field, so the current 48-character limit accepts labels the encoder cannot represent.

**3. Minor correction to the description.**

`来baba这.stark` expired on 2023-12-07. It still resolves because `domain_to_address` does not check expiry, not because the registration is live. The inconsistency you report is real either way — please just reword that claim so the rationale stays accurate.

**Commits and tests.**

Please split this into two commits so the changelog stays readable: a `fix(starknetId):` for the guard, and a `feat(starknetId)!:` with a `BREAKING CHANGE:` footer for the encoder, noting that `encodeBrotherDomain` now throws on invalid input. Test cases to add: a `bigAlphabet` name accepted by the guard, a 48-character label that overflows `PRIME`, and `useEncoded('Grug')` throwing instead of returning `9441n`.

PhilippeR26:
I pushed a commit onto this branch (thanks for leaving maintainer edits enabled). Your diagnosis was correct and your commit stays at the head of the branch — here is what changed on top of it, and why.

**The guard now derives from the encoder.** Your patch added `bigAlphabet` to the character test, which fixes the reported case. I went one step further and replaced the character test with `useEncoded(label) < PRIME` inside a `try/catch`: the guard restating the alphabets that `useEncoded` already defines is precisely what let the two drift apart in #1301, and a single source of truth keeps that from happening again.

**`useEncoded` now throws instead of skipping unknown characters.** Silently skipping them encoded a mistyped name as an existing one: `Grug` and `rug` both produced `9441` and resolved to the same mainnet address, so one stray capital sent funds to another account. `encodeBrotherDomain` calls the encoder with no guard at all, so it was exposed to the same thing.

**Two test cases had to be corrected, including one of ours.** No 48-character label can be encoded into a felt — the real ceiling is 47 ASCII characters, 40 for `这` and 20 for `来`, since a `bigAlphabet` character consumes far more encoding capacity than an ASCII one. That makes the pre-existing `'a'.repeat(48)` case from #1612 wrong, and your closing `'来'.repeat(48)` case too; both now assert the measured boundaries.

The base branch is `beta`, so this ships with v11 rather than as a patch, and the encoder change is marked as breaking.

Verified locally: 1468 unit tests, `ts:check` and `eslint` clean.

Happy to discuss any of it if you disagree with the direction.

PR #1693, opened 2026-09-07
chore(codeql): resolve warning findings

adrienlacombe:
## Summary
- assert successful calldata validation does not throw instead of reading its guaranteed void result
- remove the unused initial calldata buffer allocation in LedgerSigner231

## Testing
- npm run test:unit -- __tests__/utils/calldata/validate.test.ts
- npx prettier --check __tests__/utils/calldata/validate.test.ts src/signer/ledgerSigner231.ts
- CodeQL js/use-of-returnless-function query (zero findings in touched files)


PhilippeR26 (approved):
Looks good to me.

PR #1702, opened 2026-09-11
docs: add security policy with supported versions and reporting process

adrienlacombe:
## Motivation and Resolution

starknet.js has no `SECURITY.md`, so the repository has never stated which versions receive security fixes, how to report a vulnerability, or what reporters can expect. Recent reports reached maintainers by word of mouth, and there was no public basis for not patching 7.x, 8.x and 9.x.

This PR adds a security policy at the repository root. GitHub surfaces it under **Security > Policy** and links it from the private vulnerability reporting flow. It:

- states the supported version lines (10.x and the 11.x pre-release) and declares 7.x, 8.x and 9.x end-of-life as of September 2026,
- points reporters at private vulnerability reporting instead of public issues,
- sets response targets and describes the coordinated disclosure process,
- notes that starknet.js is outside the Starknet Immunefi bounty scope.

Two points for reviewers: the response targets (3 business days to acknowledge, 10 to assess) are a proposal, and the bug bounty sentence can be dropped if we would rather not state it publicly.

### RPC version (if applicable)

N/A

## Usage related changes

- Users can see which versions receive security fixes and are directed to upgrade off 7.x, 8.x and 9.x.
- Security researchers have a documented private reporting channel and know what to expect.

## Development related changes

- Documentation only. No code, test or CI changes.
- Suggested follow-ups outside this PR: mark the 6.x to 9.x documentation versions as unmaintained on starknetjs.com, and remove the `maintenance/*` branch pattern from the semantic-release config so an unsupported line cannot be released by accident.

## Checklist:

- [x] Performed a self-review of the code
- [x] Rebased to the last commit of the target branch (or merged it into my branch)
- [ ] Linked the issues which this PR resolves (none)
- [ ] Documented the changes in code (API docs will be generated automatically): N/A, documentation only
- [ ] Updated the tests: N/A, documentation only
- [x] All tests are passing (no code changes)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1703, opened 2026-09-11
docs: make 10.x the default docs version and flag 6.x-9.x snapshots as unmaintained

adrienlacombe:
## Motivation and Resolution

On starknet-js.com the default docs version is 9.2.1, shown with no banner as if it were the maintained line, while the 10.x docs are labelled **Next** and carry an "unreleased documentation" banner. That is the reverse of the release state: 10.x is the supported line and 7.x, 8.x and 9.x are end-of-life (see #1702).

The cause is that the docs preset sets no `lastVersion`, so Docusaurus falls back to the first entry in `versions.json` (9.2.1) and treats the working docs as unreleased.

This PR:

- sets `lastVersion: 'current'` so `www/docs` (the 10.x line) becomes the default version at `/docs/`, labelled **10.x** in the version dropdown, with no banner;
- marks the 9.2.1, 8.6.0, 7.6.4 and 6.24.1 snapshots explicitly as `unmaintained`, so each page shows "This is documentation for Starknet.js X, which is no longer actively maintained" with a link to the 10.x docs. 6.x to 8.x already showed this by default; 9.2.1 did not;
- replaces the hard-coded `/docs/next/API/classes/WebSocketChannel` link in the WebSocket guide (current docs and the 7.6.4, 8.6.0 and 9.2.1 snapshots) with a relative link to the same version's own API page. The `next` path no longer exists after this change, so the absolute links would fail the `onBrokenLinks: 'throw'` check; the relative form matches how the other guides link into the API.
- adds `@docusaurus/plugin-client-redirects` (pinned to 3.10.1 like the other Docusaurus packages) with a `createRedirects` rule that emits a `/docs/next/*` redirect page for every current-docs route, so existing links to the old `next` URLs keep working. Versioned snapshot routes are skipped since they never lived under `/docs/next/`.

Verified with a local `npm run docs:build` using the production `DOCS_URL` and `DOCS_BASE_URL`.

N/A

## Usage related changes

- `/docs/...` now serves the 10.x documentation. 9.2.1 moves to `/docs/9.2.1/...`. Existing links to `/docs/next/...` redirect to the matching `/docs/...` page; links to `/docs/...` now land on the supported line, which is what the README, the announcement bar and the security policy point at.
- Readers of any 6.x to 9.x page see an unmaintained banner pointing to the 10.x docs.
- The version dropdown reads 10.x, 9.2.1, 8.6.0, 7.6.4, 6.24.1.

- When the next major becomes the development line, update `versions.current.label` in `www/docusaurus.config.ts`. The `/docs/next/*` redirect rule should be dropped at the same time if `next` is reintroduced as a real path for the pre-release docs.
- When the "[Manual] Documentation Version PR" workflow adds a snapshot, add an explicit `banner` entry for it in the same `versions` block. Without one, Docusaurus infers `unmaintained` for anything older than `lastVersion`, so the default is safe, but explicit is clearer.
- Suggested follow-ups, not in this PR: the footer link labelled "Migrate to v9" is stale, and the 7.6.4 migration guide still carries absolute links to `starknetjs.com`, which now redirects to a third-party site.

## Checklist:

- [ ] Linked the issues which this PR resolves (none; related to #1702)
- [ ] Updated the tests: N/A, documentation site configuration only
- [x] All tests are passing (docs build passes locally; no library code changes)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/www/docs/guides/websocket_channel.md
+++ b/www/docs/guides/websocket_channel.md
@@ -365,5 +365,5 @@ main();
 ```
 
 All the types (`WebSocketOptions`, `Subscription`, the `Subscribe…Params` interfaces, …) are exported
-from `starknet`; see the [API documentation](/docs/next/API/classes/WebSocketChannel) for the full
+from `starknet`; see the [API documentation](../API/classes/WebSocketChannel) for the full
 list.
--- a/www/docusaurus.config.ts
+++ b/www/docusaurus.config.ts
@@ -356,6 +356,20 @@ const config: Config = {
 
   plugins: [
     [
+      '@docusaurus/plugin-client-redirects',
+      {
+        // Until 10.x became the default version, the current docs were served under
+        // /docs/next/. Keep those links alive by redirecting every /docs/next/* page to
+        // its /docs/* counterpart. Paths here are relative to `baseUrl`. Versioned
+        // snapshots (/docs/9.2.1/...) never lived under /docs/next/ and are skipped.
+        createRedirects(existingPath: string) {
+          const match = existingPath.match(/^\/docs\/(.+)$/);
+          if (!match || /^\d/.test(match[1])) return undefined;
+          return [`/docs/next/${match[1]}`];
+        },
+      },
+    ],
+    [
       'docusaurus-plugin-typedoc',
       {
         entryPoints: ['../src/index.ts'],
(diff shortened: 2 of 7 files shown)


PhilippeR26 (approved):
Looks good to me

PR #1705, opened 2026-09-11
feat(provider): support nodeName option in RpcProvider and RpcChannel

### LINKED ISSUE #1303: Using network names as network url
**Describe the bug**
Parameter name doesn't reflect the underlying data

**To Reproduce**
When creating a provider, such as with:

```ts
const myProvider = new RpcProvider({ nodeUrl: constants.NetworkName.SN_MAIN });
```

a network name, accessed under the `NetworkName` prop, is being assigned to a prop called `nodeUrl`, which should accept a URL.

**Expected behavior**
The prop name should match the data being assigned to it.

- If it's a network name, it should be called `nodeName`
- If it's a network URL, it should be called `nodeUrl`

lx3133584:
Fixes #1303

## Problem
`RpcProvider` and `RpcChannel` accepted network names (e.g. `constants.NetworkName.SN_MAIN`) passed via the `nodeUrl` option. While convenient, the prop name `nodeUrl` implies an HTTP(S) URL, creating a semantic mismatch when passing a network identifier instead of an actual URL endpoint.

## Solution
- Added `nodeName?: NetworkName` to `RpcProviderOptions`.
- Updated `RpcChannel` (both v0.9 and v0.10 implementations) to check `nodeName` first and resolve it to the default RPC URL via `getDefaultNodeUrl`.
- Retained full backwards compatibility for callers passing `NetworkName` via `nodeUrl`.

## Testing
Added unit test suite in `__tests__/utils/provider.test.ts` asserting:
1. `nodeName` correctly resolves to the expected network URL on both `RpcProvider` and `RpcChannel`.
2. Explicit `nodeName` takes precedence when provided.
3. Passing `NetworkName` as `nodeUrl` continues to resolve as before.
4. Custom string URLs in `nodeUrl` continue to work unchanged.

### DIFF
--- a/__tests__/utils/provider.test.ts
+++ b/__tests__/utils/provider.test.ts
@@ -1,0 +1,33 @@
+import { constants, RPC09, RPC0104, RpcProvider } from '../../src';
+
+describe('RpcProvider & RpcChannel nodeName / nodeUrl configuration', () => {
+  test('resolves nodeUrl from nodeName using NetworkName', () => {
+    const provider = new RpcProvider({ nodeName: constants.NetworkName.SN_MAIN });
+    expect(provider.channel.nodeUrl).toContain('mainnet');
+
+    const channel09 = new RPC09.RpcChannel({ nodeName: constants.NetworkName.SN_SEPOLIA });
+    expect(channel09.nodeUrl).toContain('sepolia');
+
+    const channel010 = new RPC0104.RpcChannel({ nodeName: constants.NetworkName.SN_MAIN });
+    expect(channel010.nodeUrl).toContain('mainnet');
+  });
+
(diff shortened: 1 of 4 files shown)


PhilippeR26:
Thanks for the PR and the tests.

We're closing this, together with #1303. `nodeName` would be a second name for what `nodeUrl: constants.NetworkName.*` already does: both paths end in `getDefaultNodeUrl()`. So it adds API surface without enabling anything new.

The default public nodes are meant for demos and first steps; the library already logs a hint to pass a real `nodeUrl`. If you want `nodeUrl` to only ever hold a URL, that already works: `nodeUrl: provider.getDefaultNodeUrl(constants.NetworkName.SN_MAIN)`.

PR #1709, opened 2026-09-14
build: expose tree-shakeable utility entry points

### LINKED ISSUE #1241: Better exports
Currently starknet.js export many of its utilities as object, which prevent best dead code elimination 

It would be better to expose the utils  via package.json `exports` field

so if you were to export num utils this way in your package.json:

```
{
  "name": "starknet",
   ...
  "exports": {
    ".": {
      "import": {
        "types": "./dist/index.d.ts",
        "default": "./dist/index.js"
      }
    },
    "./utils": {
      "import": {
        "types": "./dist/utils/index.d.ts",
        "default": "./dist/utils/index.js"
      }
    },
    "./utils/num": {
      "import": {
        "types": "./dist/utils/num.d.ts",
        "default": "./dist/utils/num.js"
      }
    },
  },
...
}

```

we would then be able to import it this way and benefit from dead code elimination if we were to only use part of the num exports:

```
import {toBigInt} from 'starknet/utils/num';
```

Also another bonus is that it will let us access types defined there too

CryptoAstuce:
## Motivation

Resolves #1241 by making the utility modules addressable through package subpath exports. Consumers currently have to import the root bundle even when they only need a small utility, which makes tree-shaking and type-oriented imports harder.

## Changes

- add the ./utils and ./utils/num package exports for both ESM and CommonJS consumers;
- generate the corresponding utility bundles and declaration files from the existing tsup configuration;
- keep the root export unchanged for backwards compatibility.

This keeps the public surface explicit and avoids exposing arbitrary internal files through package resolution.

## Verification

I reviewed the existing package metadata and build entry point against the current develop branch. The repository CI should validate the generated artifacts and package-resolution matrix.


PhilippeR26:
Thanks for picking this up — #1241 is a real gap, and it stays open. I'm closing this PR, though; here's the reasoning, for the record.

## `./utils` points at a file that is never built

`src/utils/index.ts` doesn't exist in this repository — the utils are re-exported from `src/index.ts` (lines 19-47), there is no barrel under `src/utils/`.

`tsup` doesn't fail on a missing entry, it silently drops it: the build logs `Building entry: src/index.ts, src/utils/num.ts` and exits successfully. So `dist/utils/index.*` is never emitted while `package.json` advertises the `./utils` subpath pointing at it. CI stays green, the package publishes, and `import … from 'starknet/utils'` fails at runtime for consumers.

## Tree shaking isn't actually enabled

`sideEffects` is still absent, so bundlers must assume every module has side effects and cannot drop unused re-exports. `import { toBigInt } from 'starknet'` still pulls the whole bundle — which is what #1241 is about. Only consumers who rewrite their imports to `starknet/utils/num` see any gain, and only for `num`.

This is the criterion published when #1324 was closed: (1) multiple entry points with preserved module structure, (2) subpath exports, (3) a `sideEffects` list whose globs match emitted files, (4) lazy default instances. This PR covers 1 and 2 partially and skips 3 and 4. #1559 was previously declined for doing *only* (3) — neither half produces tree shaking on its own.

## Measured cost

Supplying the missing barrel so the proposal could be built, full four-pass build with sourcemaps, against the same build without the extra entries:

| Format | current | with this PR | delta |
|---|---:|---:|---:|
| CJS | 2,443,783 | 4,029,035 | +64.9% |
| ESM | 2,432,387 | 2,449,214 | +0.7% |
| IIFE | 3,239,024 | 5,726,139 | +76.8% |
| **Total** | **8,886,311** | **12,997,482** | **+46.3%** |

ESM is the one format where this works: shared chunks, `utils/num.mjs` at 925 bytes. CJS and IIFE emit a standalone bundle per entry, and account for 99% of the growth. Note also that the `./utils` barrel builds to 497,530 bytes against 778,708 for the entire library — 64% of it — so that particular subpath buys almost no tree shaking.

## IIFE global collision

`entry` lives in the shared `baseConfig`, so the IIFE pass is affected too: it emits `dist/utils/index.global.js` and `dist/utils/num.global.js`, all declaring `var starknet = …`. These ship via `files: ["dist"]` and are served by jsdelivr and unpkg, so a page loading two of them gets `window.starknet` silently overwritten.

---

One last thing: please fill in `.github/pull_request_template.md` next time, the checklist in particular.

